Issue #28, opened 2026-07-23
Custom provider loads incorrect upstream models


mllt992:
<img width="1279" height="708" alt="Image" src="https://github.com/user-attachments/assets/c8a22f2a-eb58-4acc-8590-0cd843f77fd5" />

After clicking Reload, the model list is completely incorrect and does not contain the upstream models.

mllt992:
而且配置好后，列表页没有正确显示我配置的内容
点击编辑，回显空白

mllt992:
<img width="1220" height="707" alt="Image" src="https://github.com/user-attachments/assets/56efa23e-9dc4-4955-942b-0e4003cdd96c" />

astaxie:
帮你最新的代码验证一下

mllt992:
> 帮你最新的代码验证一下

<img width="997" height="880" alt="Image" src="https://github.com/user-attachments/assets/3bbac7c5-9030-44b7-bc6d-b622d8a7ef7b" />

baseurl如上图，然后我在模型映射里面点重新加载，并不是我上游配置的URL加载出来的模型

<img width="1024" height="917" alt="Image" src="https://github.com/user-attachments/assets/48ea5853-3a95-4362-aa2c-8e9f7917fe11" />

而且很奇怪的一点

<img width="1198" height="513" alt="Image" src="https://github.com/user-attachments/assets/a9f4f9fc-24fe-43e7-abf8-b1e63a6a09ea" />

我这是自定义模型，点击编辑，上边显示的却是302AI

<img width="1773" height="1222" alt="Image" src="https://github.com/user-attachments/assets/4f13d015-7698-4b15-bdbc-3710279c828c" />

需要我手动点击一次 自定义渠道商，他就会恢复成我设置的内容


<img width="1043" height="630" alt="Image" src="https://github.com/user-attachments/assets/31b9ffda-242f-4702-a493-62f0fe34053a" />

<img width="1093" height="862" alt="Image" src="https://github.com/user-attachments/assets/da286716-98ad-4290-bbd3-b90fccd10af2" />

这个时候再去加载模型，数量是对的，5个，但是他说不匹配，就很奇怪了，明明他就是OpenAI兼容

可以看下我在newapi的配置

<img width="1018" height="1094" alt="Image" src="https://github.com/user-attachments/assets/00968f5c-c40f-4cfd-9e70-55d6ecfb002a" />

好吧，那我能不能自定义输入模型名字呢

我点击路由配置

[...]

PR #43, opened 2026-07-24
fix: correct settings display across the admin console

astaxie:
## Summary

- Fix settings display and navigation behavior in the admin console
- Update provider catalog, seed data, persistence types, and server handling to keep settings data consistent
- Refresh localized labels and provider editor configuration across English and Japanese interfaces

## Testing

- Not run (not requested)


astaxie:
fix #28

PR #53, opened 2026-07-25
feat(migration): add tokenhub-migrate CLI with LiteLLM source adapter

coldbrewtea:
## Summary

Add a complete migration framework (`tokenhub-migrate` CLI) that enables repeatable, idempotent migration of competing AI gateways into TokenHub. The initial implementation ships a LiteLLM file-based source adapter and both store-backed and HTTP-backed TokenHub sinks.

## Related Issue

N/A

### Backend
- New `backend/cmd/tokenhub-migrate/` CLI entrypoint with cobra commands: `inspect`, `extract`, `plan`, `apply`, `verify`, `rollback`
- `backend/internal/migration/bundle/` — CanonicalMigrationBundle types, JSON Schema validation, SecretRef handling, versioned ID generation
- `backend/internal/migration/source/litellm/` — LiteLLM file-based adapter (supports ≥1.52.0, <1.70.0)
- `backend/internal/migration/sink/tokenhub/` — StoreSink (direct DB, for testing) and HTTPSink (remote Admin API) implementations
- `backend/internal/migration/cli/` — cobra command registration and orchestration
- Minor changes to `backend/internal/server/http.go` and `store.go` to expose admin list endpoints needed by the migration sink

### Deployment
- `deploy/docker-compose.migration-e2e.yml` — Docker Compose stack for E2E testing

### CI
- `.github/workflows/migration-ci.yml` — unit test job + optional E2E job (on `migration:e2e` label)

### SDK
- `sdk/migration-e2e/` — Node.js E2E test harness validating the full extract→plan→apply→verify→rollback cycle

### Documentation
- `docs/migration/` — English user-facing docs (README, architecture, bundle-schema, CLI, LiteLLM, E2E)
- `docs/zh-CN/migration/` — Simplified Chinese translations
- `docs/ja/migration/` — Japanese translations

## Type of Change

- [x] New feature
- [x] Documentation

## Verification

- [ ] Frontend: not applicable (no frontend changes)
- [x] SDK smoke tests against a compatible backend
- [ ] Docker Compose configuration rendered successfully (Docker not available locally; CI validates)

Verification details:

- All `go test ./internal/migration/...` pass (bundle, source, sink packages)
- `go vet ./...` clean
- `gofmt` clean
- `git diff --check` passes
- Real E2E tested locally: LiteLLM (v1.58.2) → `tokenhub-migrate` extract/plan/apply/verify → TokenHub → Bohrium upstream (3 models: qwen3.5-flash, DeepSeek-V3.2, glm-4.7) — all returned correct responses
- Idempotent re-apply verified (Created: 0 on second run)

- OpenAI-compatible `/v1` API impact: None. The migration framework operates entirely through the Admin API; no changes to the `/v1` proxy endpoints.
- Security or credential-handling impact: Secrets are never stored in bundles. All credentials use `SecretRef` (`{"$secretRef": "ENV_NAME"}`) and are resolved at apply time from environment variables. No plaintext secrets are committed.
- Database, environment, or deployment impact: New binary `tokenhub-migrate` is built separately from the main backend. No schema changes to the main TokenHub database. The `.gitignore` addition for `backend/tokenhub-migrate` prevents accidental commits of the built binary.
- Rollout and rollback considerations: The migration CLI is an additive tool; it does not affect existing TokenHub deployments. Checkpoint-based rollback is built into the apply flow for resources created during migration.

## Checklist


backend/internal/migration/source/litellm/build.go
```diff
@@ -0,0 +1,619 @@
...
+		})
+	}
+	return projects
+}
+
+func buildAPIKeys(config *Config, b *bundle.CanonicalMigrationBundle) []bundle.APIKeyRef {
+	keys := make([]bundle.APIKeyRef, 0, len(config.KeyManagement.VirtualKeys))
+	seen := map[string]struct{}{}
+	for _, key := range config.KeyManagement.VirtualKeys {
+		externalID := firstNonEmpty(key.KeyAlias, key.Token)
```
legendtkl: **[P1] Avoid embedding raw virtual-key tokens.** When `key_alias` is absent, `externalID` falls back to `key.Token` and is then serialized into `external_ref`, resource IDs, API key names, and sometimes project names. This places a plaintext LiteLLM token throughout the supposedly secret-free bundle. Please generate a non-secret synthetic identifier or hash instead; the raw token should only be represented by `SecretRef`.

backend/internal/migration/sink/tokenhub/http_client.go
```diff
@@ -0,0 +1,300 @@
...
+
+func (c *AdminAPIClient) ListProviders(ctx context.Context) ([]server.Provider, error) {
+	var resp listResponse[server.Provider]
+	if err := c.doJSON(ctx, http.MethodGet, c.endpoint("/api/admin/providers"), nil, &resp); err != nil {
+		return nil, err
+	}
+	return resp.Data, nil
+}
+
+func (c *AdminAPIClient) CreateProvider(ctx context.Context, req server.Provider) (server.Provider, error) {
```
legendtkl, line 187: **[P1] Serialize provider credentials explicitly.** `CreateProvider` and `UpdateProvider` marshal `server.Provider`, whose `APIKey` field is tagged `json:"-"`. The secret resolved by `HTTPSink` is therefore silently omitted from both requests. Please use a dedicated Admin API request DTO that includes `api_key`.

backend/internal/migration/cli/apply.go
```diff
@@ -0,0 +1,203 @@
...
+				report, err = newStoreSink().Plan(migrationBundle)
+			}
+			if err != nil {
+				return errExit(ExitSinkRejected, err.Error())
+			}
+			fmt.Printf("Dry-run plan:\n  Created: %d\n  Updated: %d\n  Skipped: %d\n", report.Created, report.Updated, report.Skipped)
+			return nil
+		}
+		if strings.TrimSpace(baseURL) != "" {
+			result, err := newHTTPSink(baseURL, token).Apply(context.Background(), migrationBundle)
```
legendtkl: **[P1] Persist one-time keys and the rollback checkpoint.** The CLI discards `result.NewKeys` and `result.Checkpoint` and only prints counters. Generated API keys are visible once, so operators cannot recover them, and `rollback` has no checkpoint file to consume. Please implement secure structured output, preferably with file mode `0600`, before returning success.

backend/internal/migration/cli/apply.go
```diff
@@ -0,0 +1,203 @@
...
+		baseURL = strings.TrimSpace(os.Getenv("TOKENHUB_API"))
+	}
+	if strings.TrimSpace(token) == "" {
+		token = strings.TrimSpace(os.Getenv("TOKENHUB_ADMIN_TOKEN"))
+	}
+	return baseURL, token
+}
+
+func newStoreSink() *migrationtokenhub.StoreSink {
+	return migrationtokenhub.NewStoreSink(server.NewMemoryStore(), secretsResolver())
```
legendtkl: **[P1] Do not report success from an ephemeral store.** Every CLI invocation creates a fresh in-memory store. Running `apply` without `--to` reports resources as created, but the state disappears when the process exits and the next `verify` sees an empty database. Please require a remote target or open an explicitly configured durable store.

backend/internal/migration/sink/tokenhub/http_sink.go
```diff
@@ -0,0 +1,726 @@
...
+	created, err := s.client.CreateModel(ctx, spec)
+	if err != nil {
+		return Change{}, server.Model{}, err
+	}
+	s.refIndex.models[item.ExternalRef.ID] = created.Name
+	return Change{Resource: "model", ID: created.Name, Action: ActionCreate}, created, nil
+}
+
+func (s *HTTPSink) applyUserHTTP(ctx context.Context, existing []server.AdminUser, item bundle.UserRef) (Change, error) {
+	if current, found := findAdminUserByIdentity(existing, item.Spec); found {
```
legendtkl: **[P1] Actually apply the desired user state.** Existing users are reported as updated without any API request, while new users are imported from `item.Spec` without resolving `item.TeamRef`. The LiteLLM adapter leaves `Spec.TeamID` empty, so remote migration loses team membership. Please resolve references, invoke the import/update path for drifted users, and refetch the resulting user ID.

backend/internal/server/http.go
```diff
@@ -3777,7 +3777,18 @@ func (s *Server) handleAdminProviderResourceNested(w http.ResponseWriter, r *htt
 	if !ok {
 		return
 	}
-	parts := strings.Split(strings.TrimPrefix(r.URL.Path, "/api/admin/provider-resources/"), "/")
+	remainder := strings.Trim(strings.TrimPrefix(r.URL.Path, "/api/admin/provider-resources/"), "/")
+	parts := []string{}
+	switch {
+	case remainder == "":
+		parts = []string{}
+	case strings.HasSuffix(remainder, "/health"):
```
legendtkl: **[P1] Preserve quota and token-refresh routes.** The new suffix parser recognizes only `/health` and `/test`. Requests ending in `/quota` or `/refresh-token` fall into the single-resource branch and return `method_not_allowed` before reaching the existing handlers below. Please include every supported nested action in the parsing logic.

backend/internal/migration/bundle/idgen.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+	if externalID == "" {
+		return "", fmt.Errorf("bundle: MintID: externalID is required")
+	}
+	switch strategy {
+	case IDStrategyStable:
+		sum := sha256.Sum256([]byte(system + "\x00" + externalID))
+		return system + "-" + hex.EncodeToString(sum[:6]), nil
+	case IDStrategyPrefixed:
+		return system + ":" + externalID, nil
```
legendtkl: **[P1] Keep generated IDs path-safe.** Prefixed IDs include the external ID verbatim. LiteLLM external IDs contain `api_base` URLs, producing IDs such as `litellm:provider:openai:http://host/v1`. Provider update and delete handlers still split decoded paths on `/`, which breaks updates and rollback. Please encode or hash the external component.

sdk/migration-e2e/litellm-e2e.mjs
```diff
@@ -0,0 +1,126 @@
...
+  record(plan.stdout.includes("Created"), "Plan prints created count", plan.stdout);
+
+  console.log("\n5. Apply the bundle (store-backed current implementation)");
+  const apply = run(MIGRATE_BIN, ["apply", "--bundle", bundlePath], { silent: true, env: { TOKENHUB_API, TOKENHUB_ADMIN_TOKEN: TOKENHUB_TOKEN } });
+  record(apply.success, "Apply succeeds", apply.stderr || apply.stdout);
+  record(apply.stdout.includes("Apply complete"), "Apply reports completion", apply.stdout);
+  writeFileSync(checkpointPath, JSON.stringify({ changes: [] }, null, 2));
+
+  console.log("\n6. Verify command behavior");
+  const verify = run(MIGRATE_BIN, ["verify", "--bundle", bundlePath], { silent: true });
```
legendtkl: **[P2] Exercise the actual remote migration cycle.** Only the first `apply` invocation receives `TOKENHUB_API`. `verify` and re-apply silently use fresh in-memory stores, while rollback consumes a fabricated empty checkpoint. The test even expects verification to fail. Please pass the remote target through the complete flow and assert real idempotence, verification, and resource removal.

legendtkl (changes requested):
## Review summary

Request changes. This PR introduces a useful migration framework, but the current implementation has several blocking correctness and security issues:

- A LiteLLM virtual-key token can be serialized into the bundle in plaintext when `key_alias` is absent.
- Provider credentials are silently omitted by the HTTP client.
- One-time generated API keys and rollback checkpoints are discarded by the CLI.
- The default local apply path reports success against a non-persistent in-memory store.
- The HTTP user path does not fully apply the desired user state.
- Path handling breaks existing provider-resource actions and generated IDs are not path-safe.
- The E2E harness does not exercise the advertised remote apply/verify/re-apply/rollback cycle.

Validation performed:

- `go test ./...`: passed
- `go test ./internal/migration/... -race`: passed
- `go vet ./...`: passed
- `gofmt` and `git diff --check`: passed
- `npm ci`: failed because `sdk/migration-e2e/package-lock.json` is missing
- `docker compose config`: syntax rendered, but the referenced `deploy/litellm-config.yaml` and `deploy/mock-upstream.conf` files are missing

I did not launch a real LiteLLM instance because the PR's Compose stack is incomplete as checked in. The PR also currently conflicts with the latest `main`.

backend/internal/migration/source/litellm/build.go
coldbrewtea: Fixed. When `key_alias` is absent, the external ID now falls back to a synthetic identifier derived from a SHA-256 hash of the token (`buildKeyExternalID`), so the raw virtual-key secret never appears in bundle IDs, refs, or logs.

backend/internal/migration/sink/tokenhub/http_client.go
coldbrewtea: Fixed in b278f49. Provider create/update now marshal a dedicated `providerWriteRequest` DTO that mirrors the server-side `ProviderCreateRequest` and carries `api_key` explicitly (`server.Provider` tags it `json:"-"`). It also pins `create_routes: false` so provider writes no longer auto-create catalog routes — routes are migrated as first-class bundle items. Regression test: `TestProviderRequestsCarryAPIKey`.

backend/internal/migration/cli/apply.go
coldbrewtea: Fixed in b278f49. `apply` now persists the rollback checkpoint to `--checkpoint-out` (default `<bundle>.checkpoint.json`) and, when new API keys were generated, their one-time plaintext to `--new-keys-out` (default `<bundle>.new-keys.json`). Both files are written with mode 0600 and the CLI prints a warning to distribute the keys securely and delete the file. Docs updated (en/zh-CN/ja).

backend/internal/migration/cli/apply.go
coldbrewtea: Fixed in b278f49. `plan`/`apply`/`verify`/`rollback` now require a remote TokenHub target (`--to` or `TOKENHUB_API`) and exit with code 5 (sink rejected) otherwise; the transient in-memory store path has been removed from the CLI entirely. Docs updated accordingly.

backend/internal/migration/sink/tokenhub/http_sink.go
coldbrewtea: Fixed in b278f49. `applyUserHTTP` now really applies the desired state: existing users are PATCHed through a new `UpdateAdminUser` client method, and new users are created via the CSV import endpoint then re-fetched so the change records the server-assigned ID into the ref index. `TeamRef` is resolved through the migrated team index consistently across apply/plan/verify. Regression test: `TestHTTPSinkApplyUserCreatesAndUpdates`.

backend/internal/server/http.go
coldbrewtea: Fixed in b278f49. The nested suffix parser now recognizes `/refresh-token` and `/quota` alongside `/health` and `/test`, restoring both routes. Regression test: `TestAdminProviderResourceActionSuffixRouting`.

backend/internal/migration/bundle/idgen.go
coldbrewtea: Fixed. Prefixed IDs no longer embed the external ID verbatim: `MintID` keeps a sanitized path-safe slug (letters, digits, `.`, `_`, `-`, capped at 48 chars) for readability and appends a short SHA-256 hash of the original value so distinct inputs stay distinct. Generated IDs are safe inside URL path segments.

sdk/migration-e2e/litellm-e2e.mjs
coldbrewtea: Fixed in b278f49. The E2E harness now runs the full remote cycle against the Admin API: every migrate command receives `TOKENHUB_API`/`TOKENHUB_ADMIN_TOKEN`; apply persists a real checkpoint (asserted on disk); verify must PASS; a re-apply asserts `Created: 0` (idempotency); rollback consumes the real checkpoint and must revert >0 changes; and a final verify is expected to FAIL, proving the rollback actually took effect.

coldbrewtea:
All 8 review findings (7 P1 + 1 P2) are addressed in b278f49, and the previously missing files (sdk/migration-e2e/package-lock.json, deploy/litellm-config.yaml, deploy/mock-upstream.conf) are now committed. Each review thread has an individual reply describing the fix and its regression test. gofmt / go vet / go test ./... all pass locally. @legendtkl PTAL — ready for re-review.

backend/internal/server/http.go
coldbrewtea: Follow-up in 9bca377: the per-suffix switch was the root cause of this regression class — every new nested action had to be added in two places (the parser and the 404 guard). Both handlers now share a table-driven `splitNestedAdminPath` helper with per-route action tables (`providerResourceActions`, `routingRuleActions`), so adding an action is a single table entry and the redundant guards are gone. Covered by the new `TestSplitNestedAdminPath` (including slashed IDs and the `res-refresh-token` boundary case) plus the existing suffix-routing regression test.

legendtkl (approved):
## Review summary

Approving. All eight blocking findings from the previous review are addressed, and the semantic conflicts that surfaced after merging the latest `main` are resolved in 975cb97, which I pushed to this branch.

### Previously requested changes — resolved by the author

- Raw virtual-key tokens no longer reach bundle IDs (`buildKeyExternalID` hashes the token).
- Provider credentials are serialized through a dedicated `providerWriteRequest` DTO carrying `api_key`.
- `apply` persists the rollback checkpoint and one-time keys with mode 0600.
- The transient in-memory store path is gone; a remote target is required.
- `applyUserHTTP` really applies desired user state and resolves `TeamRef`.
- `/quota` and `/refresh-token` routing restored, then generalized into a table-driven parser.
- Generated IDs are path-safe.
- The E2E harness runs the full remote apply/verify/re-apply/rollback cycle.

The missing `sdk/migration-e2e/package-lock.json`, `deploy/litellm-config.yaml` and `deploy/mock-upstream.conf` are all present now.

### Conflicts with current `main` — fixed in 975cb97

There was no textual merge conflict, but three tests and two repository gates failed once `main` was merged:

- **Teams were never created.** Both sinks treated bundle teams as pure ID mappings. `main` now validates team existence on every user and project mutation (`lockUserTeamsForMutation`), so apply failed with `team_not_found` and the HTTP path lost the imported user. Both sinks now resolve teams through one shared function — anchored on the bundle external ref, falling back to an exact ID match, never on the display name — and create the ones the target lacks. An ID collision under a different anchor, or an inactive match, is refused rather than reused, since `CreateResource` is an upsert and activating a team is an authorization change. Plan, Verify and Rollback share the same rules.
- **Route apply was no longer idempotent.** `sameRoute` deep-equalled the whole `ModelRoute`; the `ProjectScope`/`ProjectIDs` that `main` added are normalized to `all` on create, so every re-apply became an update. It now mirrors `UpdateRoute` merge semantics over the fields the bundle owns. An unspecified scope is left alone rather than defaulted to `all`, which would widen project access on a target route scoped to include/exclude.
- **Extra team memberships were silently dropped.** The bundle carries a single team ref, but the store rewrites a user's whole team list from every patch, so an unrelated field change wiped the others. Both sinks carry the existing list over and the HTTP DTO sends `team_ids`. Regression test included.
- Teams now go through the ID strategy instead of embedding the raw LiteLLM team ID, and the generic admin resource route parses the escaped path so an ID containing `/` survives — without it, rollback could create a team it could not delete.
- Gate fixes: this PR's additions moved into `http_nested_path.go`, `store_memory.go` and `team.go`; `TOKENHUB_API` documented in the env example and all three deployment guides.

### Known limitations — follow-up, not blocking

1. The `TeamIDs` read-then-write-back is still open to a lost update: an admin adding a team between the sink's list call and its patch would have that membership overwritten. Fixing it properly requires the server to distinguish "team_ids omitted, keep current" from "explicitly cleared", which changes an existing PATCH contract and touches the frontend. The current change is still a strict improvement — the memberships were unconditionally wiped before.
2. Team resolve-then-create is not atomic. A team created concurrently between the two steps would be overwritten by the upsert. A create-if-absent endpoint would be needed to close this.

### Validation

- `go test ./...`, `go test ./internal/migration/... -race`, `go vet ./...`, `gofmt`: all pass
- `node --test tools/*.test.mjs` (80 pass), `check-source-lines`, `check-env-contract`, `check-doc-translations --base/--head`: all pass
- `git diff --check`: clean
- Frontend `npm run typecheck` and `npm run build`: pass

Not run: the LiteLLM E2E suite, which needs a live LiteLLM instance.

PR #54, opened 2026-07-25
fix(routing): recover parked provider resources via half-open trials

legendtkl:
## Summary

A provider resource that tripped `TOKENHUB_RESOURCE_FAILURE_THRESHOLD` was parked with `healthy = false` **and** a cooldown deadline, but candidate selection only accepted `healthy = true` and no code path ever restored it automatically. The effective behaviour was "disabled until an administrator clicks test": `TOKENHUB_RESOURCE_COOLDOWN_SECONDS` never expired on its own, and the cooldown branch in `CheckProviderResourceCapacity` was unreachable during normal routing because the candidate query had already excluded the resource.

This also surfaced as an admin-visible bug: testing an `openai_codex` account from the console reported success while the account stayed unusable, because the probe path never restored health.

## Related Issue

N/A

## Changes

- **Half-open recovery on real traffic.** Once a cooldown lapses the resource is readmitted as a candidate, and the first request to reach admission claims the trial by atomically pushing `cooldown_until` forward. That single UPDATE is simultaneously the concurrency limit (one trial in flight), the failure backoff (a failed trial has already armed the next window) and the cross-replica guard (any replica losing the compare-and-set is rejected) — no new column, scheduler or distributed lock. A background prober was considered and rejected: only `openai_codex` implements `ProviderResourceProber`, so for every other provider a sweep would "recover" resources it never verified while burning real subscription quota on the one it can probe.
- **Graded attempt outcomes.** `Store.FinishProviderResourceAttempt` now takes an `AttemptOutcome` (`Succeeded` / `Neutral` / `Failed`) instead of a `success bool`. `providerAttemptCountsAsHealthy` reported success for client disconnects, policy refusals and unsupported models — outcomes that must not count against a resource but are no evidence the upstream works. Neutral outcomes now add no failure and clear none, so an alternating failure/disconnect pattern still trips the breaker.
- **Claimant-gated recovery.** Only the request holding the half-open permit can close the breaker, tracked through the call context. Without this, a request already in flight when the breaker tripped would close it on return, making the breaker largely ineffective under load.
- **Exponential backoff** on repeated recovery failures, capped by the new `TOKENHUB_RESOURCE_COOLDOWN_MAX_SECONDS` (default 3600). Doubling uses repeated addition with an early return past `max/2`, and the stored failure counter saturates, so neither the duration nor the counter can overflow.
- **Probe-driven recovery** for adapters that support probing, confined to the prober branch — the no-prober fallback never contacts the upstream, so its "success" proves nothing and must not recover.
- Docs and env synchronized across en / zh-CN / ja.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor or maintenance
- [x] Documentation

## Verification

- [ ] Frontend: `npm run typecheck` and `npm run build`
- [ ] SDK smoke tests against a compatible backend
- [x] Docker Compose configuration rendered successfully

Verification details:

- `gofmt -l internal/ cmd/` clean; `go vet ./...` clean; `go test ./...` passes (`ok tokenhub/backend/internal/server 16.1s`), including the pre-existing `TestProviderResourceCooldownAfterFailures`.
- `docker compose --env-file deploy/.env.example -f deploy/docker-compose.yml config` renders successfully.
- `git diff --check` passes.
- 20 new tests in `provider_resource_recovery_test.go`, including two end-to-end cases through `/v1/chat/completions` covering the full breaker lifecycle (trip → park → half-open trial → recover, and trial failure → re-park), plus the concurrency case where a straggler success must not close the breaker.
- Frontend and SDK checks were not run: this change is backend-only and touches no frontend code or SDK surface.

- OpenAI-compatible `/v1` API impact: none. No request or response contract changes. A parked resource that recovers simply becomes routable again.
- Security or credential-handling impact: none. No credential paths touched. Administrative status remains the highest-priority signal — a resource an admin disabled is never readmitted or recovered, and `RecoverProviderResource` re-checks status inside the UPDATE so a concurrent disable on another replica cannot be overwritten.
- Database, environment, or deployment impact: no schema change. One new optional variable, `TOKENHUB_RESOURCE_COOLDOWN_MAX_SECONDS` (default 3600), synchronized across `deploy/.env.example`, all three Compose files, and the deployment docs.
- Rollout and rollback considerations: behaviour change is confined to resources that have already tripped the breaker. Setting `TOKENHUB_RESOURCE_COOLDOWN_MAX_SECONDS` equal to `TOKENHUB_RESOURCE_COOLDOWN_SECONDS` disables the backoff widening. Rollback is a plain binary rollback; no data migration is involved, and the only persisted state used is the existing `cooldown_until` / `failure_count` columns. Note that after rollback a parked resource returns to requiring a manual test.

## Checklist

- [x] Environment variable changes are synchronized across examples, Compose, `start.sh`, and deployment documentation where applicable. (`start.sh` does not enumerate the resource breaker variables, so it needed no change.)


legendtkl:
Rebased onto `main` after #57 and #58 landed.

All conflicts were additive — both sides appending a config key, an env row or a
docs section at the same anchor. No Go logic conflicted: `store.go`, `http.go` and
`provider_error.go` merged cleanly, since this change touches
`FinishProviderResourceAttempt` / `CheckProviderResourceCapacity` while #57 touched
`FinishCall` / `RecordRejectedRequest`.

Two resolutions worth noting:

- `docs/*/deployment.md`: both sides rewrote the `TOKENHUB_RESOURCE_COOLDOWN_SECONDS`
  row. Kept this branch's wording, since "base cooldown before a parked resource is
  given a half-open retry" is what the variable now actually means, and appended #57's
  metrics rows.
- `docs/*/administrator-guide.md`: kept both new sections, with recovery placed before
  metrics so it sits next to the routing material above it.

Re-verified on the rebased tree, against the linter that #58 just added:

- `golangci-lint run ./...` — **0 issues**
- `gofmt -l internal/ cmd/` clean, `go vet ./...` clean
- `go test ./...` — `ok tokenhub/backend/internal/server 16.4s`
- `git diff --check` clean

PR #103, opened 2026-08-01
fix(observability): measure request latency with the local clock

### LINKED ISSUE #91: fix(postgres): prevent negative request latency from clock skew
In PostgreSQL deployments, `request_logs.latency_ms` can be written as a negative value. This does not occur with SQLite.

Investigation found that the PostgreSQL host in the affected environment was approximately four minutes ahead of the actual time. This clock skew directly affects backend latency calculation.

The relevant code is approximately:

```go
// Start time obtained in StartCall
now, err := s.databaseNow(tx)
...
StartedAt: now
```

Under PostgreSQL, `databaseNow` uses database time:

```go
SELECT EXTRACT(EPOCH FROM clock_timestamp())::double precision
```

When the request completes and the log is written, however, the application process's own time is used:

```go
LatencyMS: time.Since(call.StartedAt).Milliseconds()
```

The effective calculation is therefore:

```text
latency_ms = current application-server time - start time recorded by PostgreSQL
```

If PostgreSQL is four minutes ahead of the application server, a request that actually takes only a few seconds may be recorded near `-240000ms`.

SQLite behaves normally, likely because it shares the same machine/process environment as the application and therefore effectively uses the same time source.

The current workaround is to force the PostgreSQL host's time/time-zone configuration to synchronize with the application server. This avoids negative latency, but it depends on deployment configuration while the code still mixes two time sources.

Expected result: `request_logs.latency_ms` should never be negative, even if the application and PostgreSQL run on different machines or have minor clock skew.

legendtkl:
## Summary

`request_logs.latency_ms` could be written as a large negative value on PostgreSQL deployments (issue #91 reports values near `-240000`).

`CallContext.StartedAt` is a **database** clock reading — on PostgreSQL `SELECT EXTRACT(EPOCH FROM clock_timestamp())` — because `StartCall` derives the quota day/month buckets and the in-flight lease expiry from it, and every replica has to agree on those. Latency, however, was computed as `time.Since(call.StartedAt)`, i.e. against the **application process** clock:

```text
latency_ms = application host now − PostgreSQL host start reading
```

Any skew between the two hosts therefore landed directly in the stored latency. The reporter's database host ran about four minutes ahead, so a request that really took a few seconds was recorded as roughly `-240000ms`. SQLite deployments never showed it because the database and the application share one clock.

## Related Issue

Closes #91

## Changes

- `CallContext` gains an unexported `measuredAt` reference, taken next to the `databaseNow` reading in `StartCall` (and once in the admin playground handler) so both describe the same instant. It keeps its monotonic clock value, so the measurement also survives wall-clock adjustments on the application host.
- New `CallContext.elapsed()` measures against `measuredAt`, falls back to `StartedAt` for contexts assembled by hand, and never returns a negative duration.
- `FinishCall` threads the measured interval into `finishCallTransaction`, so `request_logs.latency_ms`, `provider_observations.latency_ms`, and the `tokenhub_gateway_request_duration_seconds` metric all describe the same interval. Previously each of the three re-read the clock at a different point, and the two persisted ones did so *inside* the store lock and transaction.
- `RecordPlaygroundRequest` measures before taking the store lock, matching `FinishCall`.
- New `latencyMillis` clamps latency derived from a persisted timestamp (the image-job restart-recovery path), where the replica that wrote the row may have been running ahead of the replica reading it back.
- `backend/internal/server/request_latency_test.go` covers the gateway, playground, image-completion, and image-recovery paths plus the `elapsed`/`latencyMillis` boundaries.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor or maintenance
- [ ] Documentation

## Verification

- [ ] Frontend: `npm run typecheck` and `npm run build`
- [ ] SDK smoke tests against a compatible backend
- [ ] Docker Compose configuration rendered successfully

Verification details:

- `gofmt -l backend/internal` — clean.
- `go vet ./...` — clean.
- `go test ./...` — pass (`ok tokenhub/backend/internal/server`).
- Repository gates from the root: `node --test tools/*.test.mjs` (80 pass), `node tools/check-doc-translations.mjs`, `node tools/check-env-contract.mjs`, `node tools/check-source-lines.mjs`, `git diff --check` — all pass. The translation gate ran its existence half only; there are no diff endpoints outside CI.
- Red check: with the fix reverted but the tests kept, `TestFinishCallLatencyIsNotNegativeWhenDatabaseClockRunsAhead` fails with `latency_ms = -239999`, reproducing the value reported in the issue.
- Frontend and SDK checks were skipped because no frontend, SDK, or deployment file is touched. No Docker or Compose file changed, so the Compose render check does not apply.

- OpenAI-compatible `/v1` API impact: None. No request or response shape changes.
- Security or credential-handling impact: None. No credential, key, or audit payload handling is touched.
- Database, environment, or deployment impact: No schema change, no migration, no new environment variable. One observable change in recorded values: `request_logs.latency_ms` and `provider_observations.latency_ms` are now measured before the store lock rather than inside the finish transaction, so on a contended gateway they no longer include lock and transaction time. That aligns them with the existing gateway duration metric, but latency percentiles on dashboards may step down slightly after the upgrade.
- Rollout and rollback considerations: Rollback is a plain revert; no state is written that an older build cannot read. Rows already written with a negative `latency_ms` are deliberately left untouched — this PR stops new ones from being produced but does not migrate history. Operators who want the old rows cleaned can run `UPDATE request_logs SET latency_ms = 0 WHERE latency_ms < 0;` themselves.

### Known related issue, not fixed here

Review of this change surfaced the same database-clock/application-clock split in `provider_resources.cooldown_until`: it is written with the application clock in `FinishProviderResourceAttempt` (`store_providers.go`) but written and read with the database clock on the half-open recovery path, and read with the application clock in `CheckProviderResourceCapacity`. Under the same four-minute skew a circuit-breaker backoff window can be skipped entirely. That is provider routing and circuit-breaking behavior rather than logging, so it is left for a separate change.

## Checklist

- [x] Environment variable changes are synchronized across examples, Compose, `start.sh`, and deployment documentation where applicable. (No environment variable changed.)
- [x] Shared user-facing behavior is documented consistently in English, Simplified Chinese, and Japanese where applicable. (No user-facing documentation describes `request_logs.latency_ms`, so no translation update applies.)

### DIFF
--- a/backend/internal/server/gateway_http.go
+++ b/backend/internal/server/gateway_http.go
@@ -509,6 +509,7 @@ func (s *Server) handleAdminPlaygroundChat(w http.ResponseWriter, r *http.Reques
 		return
 	}
 	requestID := NewID("pg")
+	startedAt := time.Now()
 	routed := RoutedCall{
 		Call: CallContext{
 			RequestID: requestID,
@@ -517,8 +518,9 @@ func (s *Server) handleAdminPlaygroundChat(w http.ResponseWriter, r *http.Reques
 			// The catalog model, not a bare name: pricing lives on it, and per-attempt
 			// cost is computed from whatever this carries. A name-only model silently
 			// reports every playground generation as free.
-			Model:     playgroundModel(s.store, req.Model),
-			StartedAt: time.Now().UTC(),
+			Model:      playgroundModel(s.store, req.Model),
+			StartedAt:  startedAt.UTC(),
+			measuredAt: startedAt,
 		},
 	}
 	routed.Routes = s.planRouteOrder(routed.Call, routes)
--- a/backend/internal/server/request_latency_test.go
+++ b/backend/internal/server/request_latency_test.go
@@ -1,0 +1,225 @@
+package server
+
+import (
+	"context"
+	"net/http"
+	"testing"
+	"time"
+)
+
+// skewedCall reproduces what StartCall produces on a PostgreSQL deployment whose
+// database host runs ahead of the application host: StartedAt is the database
+// clock reading, measuredAt is the local reading taken at the same moment.
+func skewedCall(requestID string, skew time.Duration) CallContext {
+	return CallContext{
+		RequestID:  requestID,
+		Project:    Project{ID: "project_clock_skew"},
+		Key:        APIKey{ID: "key_clock_skew"},
+		Model:      Model{Name: "skewed-chat", Modality: "chat", Status: StatusActive},
+		StartedAt:  time.Now().UTC().Add(skew),
+		measuredAt: time.Now(),
+	}
+}
+
+func TestFinishCallLatencyIsNotNegativeWhenDatabaseClockRunsAhead(t *testing.T) {
+	store := NewMemoryStore()
+	route := RouteSelection{Provider: Provider{ID: "provider_clock_skew", Type: ProviderOpenAI}}
+
+	store.FinishCall(skewedCall("req_clock_skew", 4*time.Minute), route, Usage{
+		PromptTokens:     10,
+		CompletionTokens: 5,
+		TotalTokens:      15,
+	}, 200, "", "127.0.0.1", "latency-test")
+
+	logs := store.ListRequestLogs()
+	if len(logs) != 1 {
+		t.Fatalf("request logs = %d, want 1", len(logs))
+	}
+	if logs[0].LatencyMS < 0 {
+		t.Fatalf("request log latency_ms = %d, want a non-negative value", logs[0].LatencyMS)
+	}
+	if logs[0].LatencyMS > time.Minute.Milliseconds() {
+		t.Fatalf("request log latency_ms = %d, want the locally measured duration", logs[0].LatencyMS)
+	}
+
+	observations := store.ListProviderObservations(time.Time{})
+	if len(observations) != 1 {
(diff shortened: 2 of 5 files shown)


legendtkl:
Rebased onto `main` (the OTLP tracing change touched the same files) and extended the fix to cover it.

`otlpTraceEmitter.record` started the root span at `completion.Call.StartedAt` — the database clock reading — while ending it at `completion.FinishedAt`, which this process stamps. Under the same four-minute skew that produced the negative `latency_ms`, that placed the whole Langfuse trace minutes in the future; the existing `finishedAt.Before(startedAt)` clamp hid the resulting negative duration but not the wrong timestamps. The span start (and the per-attempt fallback) now use the same local reference the elapsed time is measured against.

Covered by a new `TestTracingSpanStartsOnTheLocalClockWhenDatabaseClockRunsAhead`, which fails on the previous code with `span starts at ... after the call finished at ...`.

Re-ran after the rebase: `gofmt -l`, `go vet ./...`, `go test ./...` (pass), plus `node --test tools/*.test.mjs`, `node tools/check-doc-translations.mjs`, `node tools/check-env-contract.mjs`, `node tools/check-source-lines.mjs`, `git diff --check`.

PR #110, opened 2026-08-01
feat(gateway): support codex-backed cli clients

Mr0bean:
## Summary

Add direct Codex subscription support for Codex CLI, Claude Code, and Gemini CLI without requiring a local protocol switch, and add guarded Codex quota reset controls.

## Related Issue

N/A.

## Changes

- Add Codex Responses compatibility for Anthropic Messages, tool calls, reasoning signatures, streaming, compact requests, and session affinity.
- Add Gemini native `v1beta` model, token-count, generate, and streaming endpoints backed by Codex subscription routes.
- Add Codex quota reset-credit discovery and an idempotent, administrator-only reset workflow with explicit UI confirmation and quota refresh.
- Add English, Simplified Chinese, and Japanese client configuration documentation.
- Preserve the latest `main` routing, tracing, timeout, migration, and deployment changes, including Compose capability fallback handling.

## Type of Change

- [x] Bug fix
- [x] New feature
- [x] Refactor or maintenance
- [x] Documentation

## Verification

- [x] `go test ./...`
- [x] `go vet ./...`
- [x] `npm run typecheck`
- [x] `npm run build`
- [x] `node --test tools/*.test.mjs`
- [x] `bash deploy/install_test.sh`
- [x] `docker compose --env-file deploy/.env.example -f deploy/docker-compose.yml config`
- [x] Real Codex CLI 0.145.0 call against local TokenHub Responses
- [x] Real Claude Code 2.1.37 call against local TokenHub Messages
- [x] Real Gemini CLI 0.1.9 call against local TokenHub `v1beta`
- [x] Verified all three CLI requests in TokenHub audit logs; the temporary scoped verification key was deleted

The existing OpenAI-compatible `/v1` API remains compatible. This change adds direct Anthropic Messages and Gemini `v1beta` compatibility for Codex subscription routes. Codex quota reset remains administrator-only, uses idempotency protection, requires explicit client confirmation, and refreshes authoritative quota state after completion.

Quota-reset idempotency state is persisted as `AdminResource` records of kind `codex-quota-reset-operations` in the existing admin-resources table; no schema migration is required. Succeeded and failed records are retained without automatic cleanup because they prevent a replay from consuming another credit, so they remain part of normal database backup and retention. Upgrades must preserve the database. `pending` and `unknown` records survive restarts, block a different reset for the same account, and can only be retried with the same idempotency key, expected count, and credit ID until OpenAI returns a definitive outcome.

No credentials, local environment files, databases, runtime logs, competitor source trees, or private server addresses are included. The `papers/` research directory remains ignored. Deployment rollback can use the previous image or revert the feature commits.

## Checklist


backend/internal/server/gemini_native_http.go
```diff
@@ -0,0 +1,308 @@
...
+	if r.Method != http.MethodGet {
+		writeError(w, r, NewHTTPError(http.StatusMethodNotAllowed, "method_not_allowed", "Method not allowed"))
+		return
+	}
+	_, key, err := s.authenticate(r)
+	if err != nil {
+		writeError(w, r, err)
+		return
+	}
+	models := s.store.AccessibleModels(key)
```
astaxie: [P2] Filter the Gemini catalog to models this surface can actually serve

`AccessibleModels` includes every active routed model allowed by the key. This therefore advertises chat models whose only route is, for example, Anthropic (and the single-model GET below does the same), but `generateContent` later removes routes without the Responses capability and returns 501. Gemini CLI can select a model that this endpoint has just claimed is supported. Derive both model endpoints from active compatible routes—per this feature's contract, Codex subscription Responses routes—and add a regression test with an accessible chat-only model.

backend/internal/server/gemini_native_http.go
```diff
@@ -0,0 +1,308 @@
...
+		usage, err := s.streamCodexCompatibility(ctx, prepared, upstream, geminiCodexCompatibilityHeaders(r.Header, payload), sink)
+		return struct{}{}, usage, classifyStreamError(ctx, err, tracker.Wrote())
+	})
+	status, code := statusAndCode(streamErr)
+	if streamErr == nil {
+		tracker.ensureStarted()
+		s.store.MarkRouteUsed(route.Route.ID)
+		s.store.MarkProviderResourceUsed(routeResourceID(route))
+	}
+	s.store.RecordRouteAttempts(routed.Call.RequestID, attempts)
```
astaxie: [P2] Complete Gemini calls through the tracing funnel

This streaming completion path (and `finishSuccessfulGeminiCall` below) writes attempts/call/payload records directly. It never reaches `finishCall`, which is the only place that prices usage and calls `emitGatewayTrace`; consequently successful Gemini requests, plus streaming failures after output starts, produce no gateway trace. Use `finishRoutedCall`/`finishSuccessfulRoutedCall` here and add a trace-emitter regression test so this protocol preserves the unified completion semantics.

frontend/features/admin/views/provider-account-quota-reset.tsx
```diff
@@ -0,0 +1,391 @@
...
+      });
+      if (!resp.ok) throw await readResetError(resp);
+      const result = await resp.json().catch(() => ({})) as { code?: string; windows_reset?: number };
+      if (result.code !== "reset" && result.code !== "already_redeemed") {
+        throw new Error(tx("重置请求返回未知结果。请保留当前弹窗并直接重试。"));
+      }
+      clearStoredResetConfirmation(resource.id);
+      const completed = result.code === "already_redeemed" ? tx("该重置请求此前已完成，正在刷新额度。") : tx("重置请求已完成，正在刷新额度。");
+      const windowsReset = typeof result.windows_reset === "number" && Number.isFinite(result.windows_reset) ? result.windows_reset : null;
+      await Promise.allSettled([loadResetCredits(), onRefreshQuota()]);
```
astaxie: [P2] Do not treat failed authoritative refreshes as success

Both refresh functions catch their own errors and resolve, and `allSettled` discards rejection status anyway. The next lines therefore close the operation and show the completed notice even when the quota refresh failed, while `queryAccountQuota` leaves the previous quota object on screen. That contradicts the promised authoritative post-reset refresh and can make an administrator act on stale limits. Propagate/detect refresh failure, invalidate stale quota state, and expose a clear retry/refresh-failed state.

backend/internal/server/provider_account_quota_reset.go
```diff
@@ -0,0 +1,788 @@
...
+		httpErr := AsHTTPError(err)
+		s.recordAdminAuditWithStatus(r, user, "query_quota_reset_credits", "provider_resource", resourceID, "failed", httpErr.Code, "", map[string]any{"error_code": httpErr.Code})
+		writeError(w, r, err)
+		return
+	}
+	s.recordAdminAudit(r, user, "query_quota_reset_credits", "provider_resource", resourceID, "", map[string]any{"available_count": details.AvailableCount})
+	writeJSON(w, http.StatusOK, details)
+}
+
+func (s *Server) handleAdminOpenAIAccountQuotaReset(w http.ResponseWriter, r *http.Request, user AdminUser, resourceID string) {
```
astaxie, line 98: [P2] Add automated coverage for the destructive reset state machine

No backend test references these reset endpoints or handlers, despite this adding an administrator-only destructive operation with persisted idempotency states. Please cover the confirmation headers/RBAC, preflight count and credit checks, replay and payload mismatch, pending/unknown recovery, 401 credential refresh, upstream outcome mapping, and audit records using an in-process fake upstream. This is required by the repository's backend behavior-change test policy and is important protection against consuming a reset credit twice.

backend/internal/server/codex_compat_route.go
```diff
@@ -0,0 +1,232 @@
...
+	"net/http"
+	"strings"
+)
+
+const (
+	codexBridgeProtocolAnthropic = "anthropic_messages"
+	codexBridgeProtocolChat      = "chat_completions"
+)
+
+func (s *Server) executeCodexAnthropicMessages(
```
astaxie, line 15: [P2] Exercise the new Codex protocol bridges end to end

The PR adds the Codex-backed Anthropic Messages and Chat Completions routing/conversion/streaming paths, but the only changed backend test file covers Gemini. Please add fake-Codex route-level tests through `/v1/messages` and `/v1/chat/completions` for non-streaming and streaming text, tools/reasoning signatures, incompatible-route filtering, failover, and session affinity. These are public backend behavior changes and the repository explicitly requires tests for them.

astaxie (changes requested):
Requesting changes for three behavior regressions and two material test gaps:

- Successful Gemini calls bypass the unified completion funnel, so gateway traces are never emitted.
- The Gemini model endpoints advertise accessible models even when no compatible Responses/Codex route can serve them.
- The quota-reset UI reports completion after refresh failures while retaining stale quota state.
- The destructive quota-reset state machine and the new Codex Chat/Anthropic bridges have no automated coverage.

All applicable local checks passed (`go test ./...`, `go vet ./...`, frontend install/typecheck/build, repository gates, deployment install test, Compose render, and `git diff --check`). The skipped LiteLLM migration E2E is also skipped in GitHub CI.

Please also update the PR's operations section to document that reset operations are persisted as `AdminResource` records (no schema migration), how completed records are retained/cleaned up, and how pending/unknown operations behave across upgrades.

PR #120, opened 2026-08-01
feat(observability): attribute failover attempts and upstream latency

coldbrewtea:
## Summary

The gateway metrics introduced in #57 answer "how many requests, how fast, how many errors" — but they cannot answer where the latency comes from or how hard the router is working. A logical request that failed over three times looks identical to one that succeeded on the first try, and `tokenhub_gateway_request_duration_seconds` blends gateway overhead (auth, routing, DB, serialization) with upstream provider time.

This PR adds three metrics on the existing `GatewayMetrics` funnel to close that gap:

| Metric | Type | What it answers |
| --- | --- | --- |
| `tokenhub_gateway_route_attempts_total` | Counter (`model, provider_type, provider_id, resource_id, status_code, error_code, invoked`) | Physical candidate attempts. The ratio `rate(route_attempts_total) / rate(routed_requests_total)` is the failover depth; per-provider slices show who is being retried away from. |
| `tokenhub_gateway_routed_requests_total` | Counter (`model, provider_type, stream`) | Attempt-bearing logical requests: the denominator for the failover-depth ratio. `requests_total` also counts refusals that never reached a provider, which would dilute the ratio; this counter does not. |
| `tokenhub_gateway_attempt_duration_seconds` | Histogram (`model, provider_type, outcome`) | Duration of one invoked routed attempt. Measured around the whole attempt — upstream transport, stream translation, and writing to the client — so streaming calls include slow-client backpressure. Separated from end-to-end duration; the dedicated upstream-boundary signal is the TTFB metric. |
| `tokenhub_gateway_overhead_seconds` | Histogram (`stream`) | Gateway's own cost: `max(0, elapsed − Σ invoked attempt duration)`, computed per request server-side. A request admitted for routing that fails before any attempt contributes its full elapsed time. |

Design notes:

- **Overhead must be computed server-side.** Histogram quantiles cannot be subtracted in PromQL (`histogram_quantile(P99 end-to-end) − P99 attempt` is mathematically invalid), so the per-request subtraction happens before observation.
- **Single observation funnel.** Attempts ride the existing `GatewayCallCompletion.Attempts` slice; `finishCall` copies them into `CallContext.RouteAttempts` so the existing `FinishCall → observeGatewayCall` path can map them. Playground and rejected traffic are naturally excluded by `CompletionKindPlayground` / `CompletionKindRejected`.
- **Bounded cardinality.** `model` comes from the catalog-validated model name, `error_code` is the bounded internal enum from `statusAndCode`/`routeAttemptStatusAndCode`. The attempt counter's label set is the same order of magnitude as `requests_total`; the attempt histogram deliberately omits `provider_id`/`resource_id` (each histogram label layer costs ×14 series; per-resource latency is already queryable via `RouteAttemptLog` in the DB). `status_code` is the gateway-mapped status (an upstream 401 is reported as 502) to stay comparable with `requests_total`; the raw upstream status is in `RouteAttemptLog`.
- **Attempt duration semantics are documented, not overloaded.** The attempt window is the routed callback (upstream transport, stream translation, client writes), so for streaming calls it includes slow-client backpressure; the help text, docs and a slow-writer regression test pin that semantics, and the upstream boundary itself is the TTFB metric's job. The failover-depth denominator is `routed_requests_total`, which counts only attempt-bearing requests, so a rejection burst cannot dilute the ratio below 1 for healthy traffic.
- **Image jobs are wired too.** `processImageJob` threads its routed attempts into the call context on both the success and failure paths, so the failover-depth ratio is not diluted by image traffic. One caveat is documented in the metric help and docs: an image job's elapsed time includes queue wait, so its `overhead_seconds` value is an upper bound.

## Related Issue

N/A (follow-up to #57)

## Changes

- Backend
  - `types.go`: `CallContext.RouteAttempts` observability-only field.
  - `metrics.go`: 4 new collectors (`route_attempts_total`, `routed_requests_total`, `attempt_duration_seconds`, `overhead_seconds`), `GatewayAttemptSample`, `gatewayOverheadBuckets`; `ObserveGatewayCall` keeps the nil-safe and "zero value produces no series" conventions.
  - `store_routing_calls.go`: `observeGatewayCall` remains the single label-mapping point (`RouteAttempt → GatewayAttemptSample`).
  - `gateway_completion.go`: one-line `completion.Call.RouteAttempts = attemptsWithoutErrorText(completion.Attempts)` in `finishCall`, replacing the 10 per-handler assignments from the original plan and keeping upstream error text out of the observability path.
  - `image_generation.go`: `processImageJob` fills `work.call.RouteAttempts` from its routed attempts so image jobs report attempt and attempt-duration series on both completion paths.
- Tests (`metrics_test.go`): failover produces 2 attempt series while `requests_total` +1; non-invoked (capacity-rejected) attempts skip the attempt histogram; overhead clamps at 0 and is recorded for admitted zero-attempt failures (route selection, image claim); `routed_requests_total` counts only attempt-bearing requests; a slow-reading client blocks the stream copy and that backpressure lands in `attempt_duration_seconds` (documented semantics, pinned by regression) rather than overhead; rejected requests produce no new series; an end-to-end image job produces one request, one attempt, one attempt-duration and one overhead series; nil-safety extended.
- Docs: metric tables + PromQL examples (failover depth, overhead P99) + multi-instance aggregation note in `docs/administrator-guide.md`, `docs/zh-CN/…`, `docs/ja/…`.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor or maintenance
- [x] Documentation

## Verification

- [ ] Frontend: `npm run typecheck` and `npm run build` — not applicable, no frontend changes
- [ ] SDK smoke tests against a compatible backend — not applicable, no API contract changes
- [ ] Docker Compose configuration rendered successfully — not applicable, no config changes

Verification details:

- `cd backend && gofmt -l . && go vet ./... && go test ./...` — all pass.
- New tests: `TestMetricsAdmittedZeroAttemptFailureRecordsOverhead`, `TestMetricsRoutedRequestsCountsOnlyAttemptBearingRequests`, `TestMetricsStreamAttemptDurationIncludesClientBackpressure` (slow-writer regression).
- Existing `stream_failover_test.go` and `routing_policy_test.go` pass **unmodified** (behavior regression gates).
- `git diff --check` clean.

- OpenAI-compatible `/v1` API impact: None. Observability-only; no request/response changes.
- Security or credential-handling impact: None. No new data leaves the process; labels use bounded internal enums.
- Database, environment, or deployment impact: None. Reuses `TOKENHUB_METRICS_ENABLED`; no new env vars, no migrations, no compose changes.
- Rollout and rollback considerations: Metrics appear immediately after deploy when metrics are enabled; rollback simply removes the series. Existing metric names/labels/buckets untouched.

## Checklist

- [x] Environment variable changes are synchronized across examples, Compose, `start.sh`, and deployment documentation where applicable. (No env changes.)

### DIFF
--- a/backend/internal/server/metrics.go
+++ b/backend/internal/server/metrics.go
@@ -28,6 +28,11 @@ const (
 // 10s and would put almost every LLM request in +Inf.
 var gatewayDurationBuckets = []float64{0.1, 0.25, 0.5, 1, 2.5, 5, 10, 20, 30, 60, 120, 300}
 
+// gatewayOverheadBuckets is sized for gateway-only work: auth, routing, DB and
+// serialization. Values below a millisecond are rounded away by the attempt
+// latency clamp, and multi-second overhead is already a clear outlier.
+var gatewayOverheadBuckets = []float64{0.001, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10}
+
 // GatewayMetrics holds the Prometheus collectors for the model API.
 //
 // It owns its registry rather than using the default one, so tests get an isolated
@@ -52,6 +57,25 @@ type GatewayMetrics struct {
 	// like a gateway that received no traffic.
 	traceCompletions *prometheus.CounterVec
 	traceSpans       *prometheus.CounterVec
+
+	// routeAttempts counts every candidate the router considered, including
+	// capacity-skipped attempts. It is the physical attempt count behind the
+	// logical request count in routed_requests_total.
+	routeAttempts *prometheus.CounterVec
+	// attemptDuration records only invoked attempts, split by outcome. It measures
+	// the whole routed attempt — upstream transport, stream translation and
+	// writing to the client — so streaming calls include slow-client backpressure;
+	// the upstream boundary itself is not separately timed here.
+	attemptDuration *prometheus.HistogramVec
+	// routedRequests counts logical requests that made at least one candidate
+	// attempt. It is the denominator for the failover-depth ratio: requests_total
+	// also counts refusals that never reached a provider.
+	routedRequests *prometheus.CounterVec
+	// overhead is the gateway's own cost: max(0, elapsed - sum(invoked attempt
+	// duration)). A request admitted for routing that fails before any attempt
+	// contributes its full elapsed time. It is an approximation because attempt
+	// durations are rounded to whole milliseconds and clamped to a positive floor.
+	overhead *prometheus.HistogramVec
 }
 
 const (
@@ -61,6 +85,19 @@ const (
 	traceSpanOutcomeExported = "exported"
 	traceSpanOutcomeFailed   = "failed"
 )
+
+// GatewayAttemptSample is one candidate attempt, reduced to the fields the
+// metrics layer needs. Keeping it separate from RouteAttempt avoids pulling the
+// full routing structure into the metrics package.
+type GatewayAttemptSample struct {
+	ProviderType string
+	ProviderID   string
+	ResourceID   string
+	StatusCode   int
+	ErrorCode    string
+	Invoked      bool
+	Duration     time.Duration
+}
 
 // GatewayCallSample is one finished gateway request, successful or not.
 type GatewayCallSample struct {
@@ -74,6 +111,10 @@ type GatewayCallSample struct {
 	Stream       bool
 	Duration     time.Duration
 	Usage        Usage
+	// Attempts is the per-candidate outcome slice. Empty means no candidate was
+	// tried — the request was refused before routing or failed during route
+	// selection — and both still count in requests_total.
+	Attempts []GatewayAttemptSample
 }
 
 // NewGatewayMetrics builds the collectors. When projectLabel is true every metric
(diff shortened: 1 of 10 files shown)


backend/internal/server/store_routing_calls.go
```diff
@@ -927,10 +927,37 @@ func (s *GormStore) observeGatewayCall(call CallContext, route RouteSelection, u
...
+		sample := GatewayAttemptSample{
+			ProviderType: attempt.Selection.Provider.Type,
+			ProviderID:   attempt.Selection.Provider.ID,
+			ResourceID:   routeResourceID(attempt.Selection),
+			StatusCode:   attempt.Status,
+			ErrorCode:    attempt.ErrorCode,
+			Invoked:      attempt.Invoked,
+		}
+		if attempt.Invoked {
+			sample.Duration = time.Duration(attempt.LatencyMS) * time.Millisecond
```
astaxie, line 970: [P1] Measure upstream time before downstream streaming work

`attempt.LatencyMS` times the entire routed callback (`gateway_http.go:626-631`). For streaming calls that callback synchronously reads and translates each frame and writes or flushes it to the client, so slow-client backpressure and gateway serialization are included here, exported as `upstream_duration_seconds`, and then subtracted from `overhead_seconds`. `prepareRouteForUpstream` and similar pre-transport work are also inside the same interval. This defeats the promised pure upstream versus gateway split for streaming traffic. Please instrument at the upstream transport boundary, or rename and document the actual callback duration, and add a slow-writer regression.

backend/internal/server/metrics.go
```diff
@@ -275,6 +330,46 @@ func (m *GatewayMetrics) ObserveGatewayCall(sample GatewayCallSample) {
...
+			m.upstreamDuration.WithLabelValues(m.labels(
+				sample.ProjectID,
+				model,
+				attemptProviderType,
+				metricsOutcome(attempt.StatusCode),
+			)...).Observe(attempt.Duration.Seconds())
+		}
+	}
+
+	if sample.Duration > 0 && len(sample.Attempts) > 0 {
```
astaxie: [P2] Observe overhead for admitted zero-attempt failures

`len(sample.Attempts) > 0` does not only identify rejected traffic. Calls that have passed `StartCall` can fail during route selection or filtering, or during image claim or no-route handling, before an attempt is appended; `FinishCall` still increments `requests_total` with a positive `Duration`, but this gate drops their overhead sample. For those calls the upstream sum is zero and overhead should be the full elapsed time. Please distinguish rejected completion explicitly, or rely on the already-zero rejected duration, with failure-path coverage.

docs/administrator-guide.md
```diff
@@ -146,6 +149,25 @@ Requests refused before routing — a bad API key, an exhausted quota, an unknow
 
 Labels are `model`, `provider_type`, `provider_id`, `resource_id`, `status_code`, `error_code` and `stream`. Upstream failures no longer report a single `provider_error` at `status_code="502"`; they carry the codes and statuses listed under Upstream Error Classification, so dashboards and alerts that match on the old pair need updating. Upstream failures no longer report a single `provider_error` at `status_code="502"`; they carry the codes and statuses listed under Upstream Error Classification, so dashboards and alerts that match on the old pair need updating. Setting `TOKENHUB_METRICS_PROJECT_LABEL=true` adds `project_id`, which multiplies the series count of every gateway metric by the number of active projects; leave it off unless you need per-project dashboards, and use the usage reports for per-key attribution instead.
 
+Useful PromQL examples:
+
+```promql
+# Average failover depth per model.
+sum by (model) (rate(tokenhub_gateway_route_attempts_total[5m]))
+/
+sum by (model) (rate(tokenhub_gateway_requests_total[5m]))
```
astaxie: [P2] Use an attempt-bearing request denominator

`requests_total` includes quota, unknown or disallowed-model, and other `StartCall` rejections, while `route_attempts_total` correctly emits nothing for them. This query therefore reports attempts per all counted requests, not average failover depth; a rejection burst can make it less than 1 even when every routed request succeeds on its first attempt. Please provide or filter to an attempt-bearing logical-request denominator, or document the different semantics, and update the collector help plus all three translations consistently.

astaxie (commented):
## Standards

No findings.

## Spec

Found three metric-semantics issues; see the inline comments. All applicable local checks pass, but the current tests do not exercise these cases.

backend/internal/server/store_routing_calls.go
coldbrewtea: Fixed in e3d7a66 by adopting the rename-and-document option: the metric is now `tokenhub_gateway_attempt_duration_seconds`, and its help text, `gatewayAttemptSamples` comment and the docs state that the window is the whole routed callback — upstream transport, stream translation and writing to the client — so streaming calls include slow-client backpressure, and the upstream boundary itself is left to a dedicated first-byte signal. The attempt window comment at the timing site in `executeRoutedWithStore` was updated too. Added `TestMetricsStreamAttemptDurationIncludesClientBackpressure`: a real client that stalls after receiving headers blocks the gateway stream copy, and the regression asserts the backpressure lands in `attempt_duration_seconds` and not in `overhead_seconds`.

backend/internal/server/metrics.go
coldbrewtea: Fixed in e3d7a66 following the "rely on the already-zero rejected duration" option: the overhead gate is now just `sample.Duration > 0`. Rejected requests never reach it (their duration is zero by convention in `RecordRejectedRequest`), while a request admitted for routing that fails before any attempt — route selection, image claim, no-route — now records overhead equal to its full elapsed time, which is the correct value since the attempt sum is zero. Added `TestMetricsAdmittedZeroAttemptFailureRecordsOverhead` asserting the full-elapsed-time value, plus failure-path assertions in `TestMetricsRoutedRequestsCountsOnlyAttemptBearingRequests`.

docs/administrator-guide.md
coldbrewtea: Fixed in e3d7a66 by providing the attempt-bearing denominator: new counter `tokenhub_gateway_routed_requests_total` (labels `model`, `provider_type`, `stream`) increments only when `len(sample.Attempts) > 0`. The docs, the `route_attempts_total` help text and the PromQL example now use `rate(route_attempts_total) / rate(routed_requests_total)`, and the rejection paragraph explains why `requests_total` would dilute the ratio. All three translations and the collector help are updated consistently. Added `TestMetricsRoutedRequestsCountsOnlyAttemptBearingRequests` (attempt-bearing, admitted-failure and rejected samples) and extended the rejected-request and image-job tests to pin the new counter.

PR #132, opened 2026-08-03
fix(admin): expose language preferences to all users

assclb:
## Summary

The admin console currently falls back to English when no preference is saved, while its general language control is only available under system settings. This makes the initial language surprising for Chinese and Japanese browsers and prevents lower-privilege users from changing it.

## Related Issue

Supersedes #131.

## Changes

- Resolve the first-run language from the browser's supported locale list while keeping a saved browser preference authoritative.
- Add a compact language menu beside the theme control on login and password-reset pages.
- Add the same language menu to the global top bar so every signed-in role can change the UI language.
- Render the menu in a portal to avoid shell clipping and support outside-click, Escape, focus, and arrow-key interactions.
- Keep the existing system-settings preference entry and reuse the shared language components.
- Add regression tests for locale detection, fallback, and saved-preference precedence.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor or maintenance
- [ ] Documentation

## Verification

- `node --test tools/*.test.mjs` (102 tests passed)
- `node tools/check-ui-translations.mjs --base origin/main --head HEAD`
- `npm run typecheck`
- `npm run build`
- `git diff --check origin/main..HEAD`

No API, persistence, authorization, environment, or deployment changes. The preference remains local to the current browser, and existing saved language values continue to take precedence.

## Checklist

### DIFF
--- a/frontend/features/admin/i18n/language-switcher.tsx
+++ b/frontend/features/admin/i18n/language-switcher.tsx
@@ -1,0 +1,165 @@
+import { Check, ChevronDown, Globe2 } from "lucide-react";
+import { useEffect, useId, useRef, useState } from "react";
+import { createPortal } from "react-dom";
+import { type AppLanguage, languageOptions, tx } from "./runtime";
+
+export function LanguageSwitcher({
+  language,
+  onChange,
+  className,
+}: {
+  language: AppLanguage;
+  onChange: (language: AppLanguage) => void;
+  className?: string;
+}) {
+  return (
+    <div className={className ? `language-switcher ${className}` : "language-switcher"} role="radiogroup" aria-label={tx("界面语言")}>
+      {languageOptions.map((option) => (
+        <button
+          aria-checked={language === option.value}
+          className={language === option.value ? "active" : ""}
+          key={option.value}
+          onClick={() => onChange(option.value)}
+          role="radio"
+          type="button"
+        >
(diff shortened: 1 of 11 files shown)


assclb:
Current UI screenshots for this change:

1. Language selector on the login page

Language selector on the login page

2. Language menu on the login page

Language menu on the login page

3. Language menu in the authenticated console header

Language menu in the authenticated console header

frontend/features/admin/i18n/language-switcher.tsx
```diff
@@ -0,0 +1,165 @@
...
+        <span className="language-select-label">{current.nativeLabel}</span>
+        <ChevronDown aria-hidden="true" className="language-select-chevron" size={14} />
+      </button>
+      {open && typeof document !== "undefined" ? createPortal(
+        <div
+          className="language-select-options"
+          data-theme={menuTheme}
+          id={menuID}
+          ref={optionsRef}
+          role="listbox"
```
astaxie, line 137: [P2] Please give the portaled listbox an accessible name. The trigger’s `aria-label` and `aria-controls` do not label this separate `role="listbox"`, so after focus moves into an option, screen-reader users lose the “interface language” context. Adding `aria-label={tx("界面语言")}` here (or an `aria-labelledby` reference to a stable trigger ID) would preserve that context.

PR #139, opened 2026-08-04
fix(gateway): bridge Claude reasoning for OpenAI providers

### LINKED ISSUE #133: feat(gateway): adapt Claude Code reasoning parameters and thinking output
# feat(gateway): adapt Claude Code reasoning parameters and thinking output

## Background

When Claude Code connects through TokenHub's `/v1/messages` endpoint to an OpenAI-compatible provider (SGLang + GLM-5.2-FP8 in this case), two independent problems occur:

1. Claude Code reasoning parameters use a different value domain from the upstream provider. TokenHub currently forwards them unchanged, causing upstream rejection.
2. The upstream returns `reasoning_content`, but TokenHub's OpenAI Chat Completion → Anthropic Messages conversion does not generate a `thinking` content block, so Claude Code cannot recognize the model's reasoning.

This is not only an `xhigh` compatibility issue. It must cover Claude Code's complete reasoning configuration and differences among OpenAI-compatible provider value domains.

## Test environment

- TokenHub version: v0.4.0 (with only an unrelated timeout patch)
- Latest TokenHub `main` inspected at: `110f2d5d372de11cf19b13f7d1be41e71034e49a`
- Claude Code: 2.1.220
- SGLang: 0.5.15.post1
- Model: GLM-5.2-FP8
- Path: Claude Code → TokenHub `/v1/messages` → OpenAI-compatible provider → SGLang

All reproductions were performed in an isolated local TokenHub environment without modifying SGLang configuration.

## Problem 1: reasoning parameters are forwarded unchanged despite incompatible provider values

On the latest `main`, `anthropicToOpenAIChatRequest` reads `effort` and `output_config.effort`, then writes the value unchanged to `reasoning_effort`:

- [`anthropic_messages.go#L255-L276`](https://github.com/astaxie/TokenHub/blob/110f2d5d372de11cf19b13f7d1be41e71034e49a/backend/internal/server/anthropic_messages.go#L255-L276)

SGLang currently accepts:

```text
none, low, medium, high, max
```

Observed results:

| Messages input | Value forwarded by TokenHub | Result |
|---|---|---|
| `output_config.effort=none` | `reasoning_effort=none` | HTTP 200 |
| `output_config.effort=minimal` | `reasoning_effort=minimal` | Upstream HTTP 400; TokenHub returns 502 |
| `output_config.effort=low` | `reasoning_effort=low` | HTTP 200 |
| `output_config.effort=medium` | `reasoning_effort=medium` | HTTP 200 |
| `output_config.effort=high` | `reasoning_effort=high` | HTTP 200 |
| `output_config.effort=xhigh` | `reasoning_effort=xhigh` | Upstream HTTP 400; TokenHub returns 502 |
| `output_config.effort=max` | `reasoning_effort=max` | HTTP 200 |

The actual SGLang error is:

```text
Input should be 'none', 'low', 'medium', 'high' or 'max'
```

Claude Code also uses Anthropic `thinking` configuration. The generic OpenAI-compatible path currently ignores:

```json
{"thinking":{"type":"disabled"}}
{"thinking":{"type":"adaptive"}}
{"thinking":{"type":"enabled","budget_tokens":1024}}
```

The code handles only `effort` and `output_config.effort`; `thinking.type` and `thinking.budget_tokens` do not contribute to the OpenAI `reasoning_effort`. Consequently, `disabled` cannot reliably disable upstream reasoning, and `enabled + budget_tokens` cannot control the upstream reasoning level.

## Problem 2: SGLang reasoning_content is lost in Messages conversion

Buffered tests with the same model and prompt produced:

### Through TokenHub Chat Completions

```json
{
  "content": "1813",
  "reasoning_content": "...complete reasoning process...",
  "usage": {
    "completion_tokens": 205,
    "reasoning_tokens": 202
  }
}
```

### Through TokenHub Messages

```json
{
  "content": [
    {"type": "text", "text": "1813"}
  ],
  "stop_reason": "end_turn"
}
```

The response contains no `thinking` content block.

In streaming tests, SGLang returned 53 frames containing `delta.reasoning_content`. The Anthropic SSE converted by TokenHub contained only `text_delta`; the counts of `thinking` and `thinking_delta` were both zero.

The latest `main` code matches the observed behavior:

- Buffered conversion reads only `message.content` and `message.tool_calls`, not `message.reasoning_content`: [`anthropic_messages.go#L635-L675`](https://github.com/astaxie/TokenHub/blob/110f2d5d372de11cf19b13f7d1be41e71034e49a/backend/internal/server/anthropic_messages.go#L635-L675)
- Streaming conversion handles only `delta.content` and `delta.tool_calls`, not `delta.reasoning_content`: [`anthropic_messages_stream.go#L201-L282`](https://github.com/astaxie/TokenHub/blob/110f2d5d372de11cf19b13f7d1be41e71034e49a/backend/internal/server/anthropic_messages_stream.go#L201-L282)

When reasoning consumes the entire output-token budget, SGLang's `content` may be empty. TokenHub currently returns an empty text block, so callers see only a successful response with no body and cannot tell that the model generated reasoning.

## Direct Claude Code verification

After enabling Claude Code extended thinking, run:

```bash
MAX_THINKING_TOKENS=1024 \
ANTHROPIC_BASE_URL=http://127.0.0.1:13001 \
ANTHROPIC_AUTH_TOKEN='<TOKENHUB_KEY>' \
claude -p 'Think carefully and calculate 37×49. Do not call tools; output only the final number.' \
  --model glm-5.2-fp8 \
  --max-turns 1 \
  --tools '' \
  --bare \
  --setting-sources project,local \
  --output-format stream-json \
  --include-partial-messages \
  --verbose
```

Observed result:

- Claude Code exit code: 0
- Final text: `1813`
- `thinking` / `thinking_delta` / `reasoning_content`: 0 occurrences
- The Chat Completions response through the same route clearly contains the complete `reasoning_content`

This is therefore not a Claude Code display-setting problem. TokenHub's OpenAI → Anthropic response conversion does not emit thinking events Claude Code can recognize.

### 1. Add provider-level reasoning-parameter compatibility configuration

Do not hard-code only `xhigh -> max`. Allow providers or provider resources to configure:

- The set of upstream-supported `reasoning_effort` values.
- Mappings from Claude/Anthropic effort to upstream effort.
- Mapping policy for `thinking.type`.
- Thresholds or mappings from `thinking.budget_tokens` to reasoning levels.
- Behavior when the upstream does not support reasoning parameters: remove the parameter, use a default, or return an explicit client error.

SGLang could provide default compatibility such as:

| Claude/unified value | SGLang value |
|---|---|
| `none` | `none` |
| `minimal` | `low`, or configurable as `none` |
| `low` | `low` |
| `medium` | `medium` |
| `high` | `high` |
| `xhigh` | `max` |

Also handle:

- `thinking.type=disabled` → `reasoning_effort=none`
- `thinking.type=adaptive/auto` → the provider's default reasoning level
- `thinking.type=enabled + budget_tokens` → a reasoning level mapped by provider configuration
- Define explicit precedence when `output_config.effort`, top-level `effort`, and `thinking` appear together.

### 2. Complete reasoning_content → Anthropic thinking response conversion

Buffered:

- `message.reasoning_content` → `content[].type=thinking`
- Support optional `reasoning_signature` and `redacted_reasoning_content`

Streaming:

- `delta.reasoning_content` → `content_block_start(type=thinking)`
- Emit consecutive `content_block_delta(type=thinking_delta)`
- End the thinking block before text or tool calls begin

If Claude Code replay requires a `signature`, reuse TokenHub's existing Provider Signature mechanism to generate and validate a gateway signature so the next turn's assistant thinking block can be converted safely back to OpenAI-compatible fields. If safe replay is impossible, use an explicit, testable fallback policy.

## Acceptance criteria

1. Claude Code `none/minimal/low/medium/high/xhigh` values convert deterministically according to provider configuration and unsupported values are never forwarded unchanged.
2. `thinking.type=disabled/adaptive/enabled` and `budget_tokens` take effect on the generic OpenAI-compatible path.
3. Buffered `reasoning_content` returned by SGLang converts to an Anthropic `thinking` content block.
4. Streaming `delta.reasoning_content` returned by SGLang converts to a valid `thinking_delta` event sequence.
5. With extended thinking enabled, Claude Code can recognize thinking content in a real session instead of receiving only final text.
6. Reasoning-only output with empty final `content` no longer converts to an uninformative successful response with an empty text block.
7. When reasoning, text, and tool use are mixed, Anthropic content-block order and SSE lifecycles are valid.
8. Add buffered, streaming, tool-call, and multi-turn thinking replay tests without affecting native Anthropic providers or ordinary OpenAI Chat clients.

hisuL:
## Summary

修复 Claude Code 经 `/v1/messages` 接入 OpenAI 兼容 Provider 时的两类兼容问题：推理参数值域不一致，以及上游 `reasoning_content` 在 Anthropic Messages 转换中丢失。实现保持通用，不对 SGLang 或单一模型硬编码，并在管理控制台提供 Provider / Provider Resource 两级配置界面。

## Related Issue

Closes #133.

## Changes

- 通过 Provider / Provider Resource `options` 配置 effort 映射、支持值集合、不支持值策略和 `budget_tokens` 阈值映射。
- 转换 `thinking.type=disabled/adaptive/enabled`、`thinking.budget_tokens`、顶层 `effort` 与 `output_config.effort`。
- allowlist 未命中时默认删除参数；只有显式配置 `passthrough` 才透传，`reject` 会过滤当前不兼容路由并继续尝试其他路由。
- 将非流式 `message.reasoning_content` 转为 Anthropic `thinking` 内容块。
- 将流式 `delta.reasoning_content` 转为合法的 `thinking_delta` / `signature_delta` 事件序列，并保证 thinking、text、tool use 顺序。
- 生成并校验 Provider 范围的 TokenHub 回放标记；仅在显式允许时把 reasoning 内容回传给支持该字段的上游，回退到原生 Anthropic 时会剥离 TokenHub 标记。
- 纯 reasoning、正文为空时不再补无信息的空文本块；Token 兜底估算纳入 reasoning 内容。
- 管理控制台的 Provider 高级页新增 Claude Code 推理兼容折叠区，并为每个 Provider Resource 提供独立覆盖配置、JSON 校验、保存反馈和中英日文案。
- 增加参数映射、allowlist 默认行为、异构路由过滤、原生 Anthropic 回退、非流式、流式、工具混合、多轮回放及跨 Provider 拒绝测试，并同步英中日文档。

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Refactor or maintenance
- [x] Documentation

## Verification

- `GOTOOLCHAIN=auto go test ./...`
- `GOTOOLCHAIN=auto go vet ./...`
- `npm run typecheck`
- `npm run build`
- `git diff --check`
- 本地浏览器验证 Provider / Provider Resource 两级表单：中英文渲染、非法 JSON 拦截、保存、数据库写入、重新打开回显、清空恢复和未知 options 保留均通过。
- 本地隔离链路：Claude Code 2.1.221 -> 本 PR TokenHub -> SGLang 0.5.15.post1 -> GLM-5.2-FP8。
- 非流式 `output_config.effort=xhigh` 经配置映射为 `max`，HTTP 200，并返回 thinking。
- 非流式 `thinking.enabled + budget_tokens=1024`，HTTP 200，返回 thinking + text，最终文本为 `1813`。
- 流式响应得到 51 个 `thinking_delta`、1 个 `signature_delta`、2 个 `text_delta`，且 message 生命周期完整。
- Claude Code 实测得到 44 个事件、15 个 thinking 事件、最终结果 `1813`、`subtype=success`。验证过程未使用代理，也未修改 SGLang 配置。

- 未配置 allowlist 时保持原有 effort 透传；配置 allowlist 后，未命中值默认删除，只有显式选择 `passthrough` 才原样转发。
- 所有新配置均为可选项，不需要数据库迁移或新增环境变量。
- OpenAI 兼容上游返回 reasoning 时，Messages 客户端会新增 Anthropic thinking 内容块；普通 OpenAI Chat Completions 路径不受影响。
- 回放标记绑定 Provider ID 与 reasoning 内容，用于一致性校验，不是凭据或安全边界；跨 Provider 标记不会被回放，原生 Anthropic 路由不会收到 TokenHub 本地标记。
- `preserve_reasoning_content` 默认关闭（DeepSeek 延续原有行为），避免向未知上游发送扩展字段。
- 回滚只需撤销本提交；已配置的未知 options 会被旧版本忽略。

## Checklist


backend/internal/server/anthropic_reasoning.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+	}
+
+	if effort == "" || effort == "omit" {
+		chatReq.ReasoningEffort = nil
+		return nil
+	}
+	if supported := reasoningEffortValues(provider.Options[reasoningEffortValuesOption]); len(supported) > 0 {
+		if _, ok := supported[effort]; !ok {
+			switch strings.ToLower(strings.TrimSpace(provider.Options[reasoningEffortUnsupportedOption])) {
```
astaxie, line 52: [P1] Do not default an allowlist miss to passthrough

When `reasoning_effort_values=none,low,medium,high,max` is configured and the request contains `xhigh`, an empty `reasoning_effort_unsupported` falls out of this switch and line 66 still sends `xhigh` upstream. That reproduces the original SGLang 400/TokenHub 502 even though the administrator supplied an allowlist, and conflicts with issue #133’s acceptance criterion that unsupported values are not forwarded unchanged. Please make an allowlist miss default to `omit` or `reject` (leaving passthrough available only when explicitly selected), and cover the allowlist-only configuration in a regression test.

backend/internal/server/anthropic_reasoning.go
```diff
@@ -0,0 +1,169 @@
...
+	"strings"
+)
+
+const (
+	reasoningEffortMapOption         = "reasoning_effort_map"
+	reasoningEffortValuesOption      = "reasoning_effort_values"
+	reasoningEffortUnsupportedOption = "reasoning_effort_unsupported"
+	reasoningBudgetMapOption         = "reasoning_budget_map"
+	reasoningContentOption           = "preserve_reasoning_content"
+	reasoningSignatureProvider       = "tokenhub-openai-compatible"
```
astaxie, line 20: [P1] Strip this marker before native Anthropic fallback

Responses now expose signatures in this new `tokenhub-openai-compatible:` namespace, but `nativeAnthropicPayload` only recognizes `codex` and `gemini` in `foreignProviderSignature`. If the next tool turn routes or fails over from this OpenAI-compatible provider to native Anthropic, TokenHub forwards this local marker verbatim as an Anthropic thinking signature, which the upstream will reject. Please register this namespace as foreign for native Anthropic (or make the foreign-marker check generic) and add an OpenAI-compatible → native Anthropic multi-turn fallback test.

backend/internal/server/codex_compat_anthropic_route.go
```diff
@@ -97,7 +97,7 @@ func validateAnthropicRouteCompatibility(route RouteSelection, req anthropicMess
 			"Provider does not support the Anthropic Messages gateway",
 		)
 	}
-	_, err := anthropicToOpenAIChatRequest(req)
+	_, err := anthropicToOpenAIChatRequest(req, route.Provider)
```
astaxie, line 100: [P1] Filter a rejecting provider instead of aborting all routes

This route-specific conversion can now return `unsupported_reasoning_effort`, but `compatibleAnthropicRoutes` does not classify that code as a route incompatibility. Consequently, if one provider is configured with `reasoning_effort_unsupported=reject`, the compatibility scan returns its 400 immediately and never uses another route whose allowlist supports the requested effort. That breaks the heterogeneous-provider use case this feature is meant to support. Please treat this code like the other provider-specific incompatibilities so the first error is returned only when no compatible route remains, and add a two-provider regression test.

astaxie (commented):
I found three issues in the provider-specific reasoning compatibility path: an allowlist miss is still passed upstream by default, the new replay marker is forwarded incorrectly during native Anthropic fallback, and one provider rejecting an effort prevents other compatible routes from being considered.

backend/internal/server/anthropic_reasoning.go
hisuL: Fixed in 47a1cca. An allowlist miss now defaults to omit; passthrough is honored only when explicitly configured. Added regression coverage for both the allowlist-only default and explicit passthrough.

backend/internal/server/anthropic_reasoning.go
hisuL: Fixed in 47a1cca. The TokenHub OpenAI-compatible reasoning signature namespace is now classified as foreign by nativeAnthropicPayload and stripped before native Anthropic fallback. Added a regression test that preserves portable content while removing the local thinking block.

backend/internal/server/codex_compat_anthropic_route.go
hisuL: Fixed in 47a1cca. unsupported_reasoning_effort is now a route incompatibility, so a rejecting route is filtered and later compatible routes remain eligible. Added a two-provider regression test.

frontend/features/admin/domain/provider-reasoning.ts
```diff
@@ -0,0 +1,110 @@
...
+    .map((value) => value.trim().toLowerCase())
+    .filter(Boolean)
+    .join(",");
+  if (allowed) options.reasoning_effort_values = allowed;
+
+  const effortMap = normalizedJSONObject(values.reasoning_effort_map, "推理强度映射");
+  if (effortMap) options.reasoning_effort_map = effortMap;
+
+  const unsupported = values.reasoning_effort_unsupported?.trim().toLowerCase();
+  if (unsupported && unsupported !== "omit") options.reasoning_effort_unsupported = unsupported;
```
astaxie: [P1] Persist resource-level opt-outs instead of inheriting Provider values

On a Provider Resource, the form displays missing values as `omit`, `false`, and empty allowlist/maps, but this serializer removes all of those values. Since route selection only overlays keys actually stored in `resource.Options`, saving these Resource settings still inherits a Provider-level `reasoning_effort_unsupported=reject`, `preserve_reasoning_content=true`, or non-empty allowlist/map. For example, an administrator can select `omit` or turn off reasoning replay for a particular endpoint, receive the saved confirmation, and the endpoint will continue rejecting the effort or receiving `reasoning_content`. Please give Resource settings explicit inherit/override (or tombstone) semantics and persist `omit`, `false`, and explicit empty values when they are intended to override the Provider; add regressions for Provider=`reject/true` with Resource=`omit/false`.

astaxie (commented):
One blocking resource-override issue remains after the latest fixes.

frontend/features/admin/domain/provider-reasoning.ts
hisuL: Addressed in 88ddfd8. Resource reasoning settings now use explicit inherit/override semantics; custom overrides persist `omit`, `false`, and explicit empty allowlist/maps, while switching back to inherit removes only the resource-level reasoning keys. Added CI regression coverage for Provider=`reject/true` with Resource=`omit/false`, explicit empty values, and clearing overrides.

frontend/features/admin/resources/payloads.tsx
```diff
@@ -137,6 +139,7 @@ export function providerResourceToForm(item: ProviderResource) {
     environment: item.environment ?? "",
     status: item.status,
     healthy: String(item.healthy),
+    ...providerReasoningFormValues(item.options),
```
astaxie: [P1] Preserve explicit Resource tombstones in the normal edit flow

The dedicated reasoning editor now stores empty allowlists/maps, omit, and false so they override Provider defaults, but the normal Provider Resource editor still reaches this path through providerResourceToForm, which uses providerReasoningFormValues without setting _reasoning_override. On any unrelated Resource edit, serializeProviderReasoningOptions deletes all five keys and only re-adds non-empty/non-default values. I reproduced a Resource containing the five explicit override markers plus an unrelated option round-tripping to only the unrelated option, after which routing silently inherits the Provider allowlist/maps/reject/replay settings again. Please preserve the override marker in this generic form path and add a round-trip regression.

backend/internal/server/anthropic_reasoning.go
```diff
@@ -0,0 +1,172 @@
...
+		if value = strings.ToLower(strings.TrimSpace(value)); value != "" {
+			values[value] = struct{}{}
+		}
+	}
+	return values
+}
+
+func providerPreservesReasoningContent(provider Provider) bool {
+	value := strings.ToLower(strings.TrimSpace(provider.Options[reasoningContentOption]))
+	return provider.Type == "deepseek" || value == "true" || value == "1" || value == "yes"
```
astaxie: [P2] Let an explicit false override DeepSeek's legacy default

DeepSeek is included in the new reasoning settings UI, and a Resource custom override now deliberately persists preserve_reasoning_content=false. This predicate nevertheless returns true for every DeepSeek Provider, so the toggle and Resource opt-out have no effect and assistant reasoning is still replayed upstream. Please distinguish an absent option, where keeping DeepSeek's legacy default-on behavior is reasonable, from an explicit false, or make the UI show this setting as forced; add a DeepSeek false regression.

frontend/features/admin/domain/provider-reasoning-options.ts
```diff
@@ -0,0 +1,93 @@
...
+    preserve_reasoning_content: source.preserve_reasoning_content === "true" ? "true" : "false",
+  };
+}
+
+export function providerReasoningOverrideFormValues(
+  resourceOptions?: Record<string, string>,
+  providerOptions?: Record<string, string>,
+) {
+  const overridden = providerHasReasoningOverrides(resourceOptions);
+  const displayed = overridden ? resourceOptions : providerOptions;
```
astaxie: [P2] Display the effective values for partial Resource overrides

If a Resource contains any one reasoning key, this selects only resourceOptions, while backend routing actually overlays the Resource keys on top of the Provider options. A partial override created through the API therefore appears here with every inherited field blank/omit/false; saving the form then materializes those displayed defaults and changes behavior. Please initialize an existing custom override from the merged Provider and Resource options while keeping _existing_options Resource-only, and cover a partial override regression.

astaxie (commented):
The dedicated Resource reasoning editor now handles explicit inheritance and overrides, but three configuration round-trip gaps remain.

frontend/features/admin/resources/payloads.tsx
hisuL: Fixed in 96e5001. The normal Provider Resource editor now initializes reasoning fields through the explicit Resource override form path and carries the reasoning override marker, with Provider options supplied when available. Unrelated Resource edits therefore preserve omit, false, and explicit empty allowlist/map tombstones. Added a normal-edit round-trip regression covering all five markers plus an unrelated option.

backend/internal/server/anthropic_reasoning.go
hisuL: Fixed in 96e5001. DeepSeek now keeps the legacy default-on behavior only when preserve_reasoning_content is absent. An explicitly configured false overrides that default and strips reasoning_content. Added an adapter-level DeepSeek false regression that verifies the upstream wire payload.

frontend/features/admin/domain/provider-reasoning-options.ts
hisuL: Fixed in 96e5001. Existing partial Resource overrides are now displayed from the effective Provider plus Resource overlay, while the existing-options snapshot remains Resource-only. Saving the form preserves effective behavior without misrepresenting inherited fields. Added a partial-override regression covering display values, Resource-only existing options, and serialization.

astaxie (commented):
Re-review of current head 96e5001: the four outstanding resource-override findings are addressed. I verified the normal edit round trip, explicit DeepSeek false handling, and partial effective-value overlay, and resolved those threads. Local validation passes: go test ./..., go vet ./..., npm ci, npm run typecheck, npm run build, repository gates, explicit translation co-change check, and git diff --check. This PR is not merge-ready only because GitHub reports no checks: Actions run 31138944560 completed with action_required and needs maintainer approval before CI can run.

PR #144, opened 2026-08-05
feat(server): enforce configurable JSON request body limits

### LINKED ISSUE #125: fix(gateway): make the 4 MiB JSON request-body limit configurable
## Background

On the current `main` branch, `decodeJSON` in `backend/internal/server/http_transport.go` hard-codes a **4 MiB** JSON request-body limit:

```go
const maxRequestBodyBytes = 4 << 20
```

This affects every OpenAI-compatible JSON endpoint that reuses this decoding logic.

## Impact

Codex multimodal requests may inline image data URLs or large structured context in JSON. Base64 encoding adds roughly 33% overhead, so fairly ordinary image input can exceed 4 MiB and fail with:

```
request body exceeds 4194304 bytes
```

The multi-instance Nginx configuration currently permits `client_max_body_size 32m`, which is inconsistent with the backend's actual 4 MiB limit.

## Expected behavior

Administrators should be able to configure a reasonable JSON request-body limit, or endpoints carrying inline images/multimodal input should have separate limits.

Additionally:

- Return an explicit `413 Payload Too Large` when the limit is exceeded instead of a generic JSON-decoding error.
- Keep reverse-proxy `client_max_body_size` aligned with the backend limit or document the difference clearly.
- Continue to account for memory and DoS risk in the default; do not simply raise every endpoint's permanent fixed limit to 128 MiB.

## Current workaround

In a self-hosted environment, I raised both the backend JSON limit and the multi-instance Nginx limit to 128 MiB, allowing large Codex requests to pass.

However, the current implementation reads the entire request into memory through `io.ReadAll`, so a high global fixed limit amplifies memory usage under concurrency. A configurable or endpoint-specific solution is preferable.

## Suggested direction

For example:

- Add a setting such as `TOKENHUB_MAX_JSON_REQUEST_BYTES`.
- Change decoding to accept endpoint-specific size limits.
- Use a higher independent limit for multimodal/image endpoints.
- Document defaults, proxy configuration, and security boundaries.

samz406:
## Summary

`decodeJSON` capped every JSON request body at a hard-coded 4 MiB. Multimodal requests (Codex/vision) inline base64-encoded images (~33% overhead), so common image inputs exceeded 4 MiB and failed. The limit could not be changed without a code patch, and over-limit bodies surfaced as a confusing generic `400` (silent truncation → mid-stream parse error) rather than a `413`.

This makes the limit admin-configurable, with a separate higher ceiling for multimodal endpoints, and returns a clear `413 Payload Too Large` on over-limit while preserving streaming decode.

## Related Issue

Closes astaxie/TokenHub#125

## Changes

- **Config:** `TOKENHUB_MAX_JSON_REQUEST_BYTES` (default 8 MiB) and `TOKENHUB_MAX_MULTIMODAL_REQUEST_BYTES` (default 32 MiB), parsed by `getenvBytes` — raw bytes or a binary suffix (`8m`, `8mib`, `512k`), clamped at a 512 MiB ceiling.
- **Decode:** `decodeJSON` becomes `*Server` methods (`decodeJSON` / `decodeJSONLimit` / `decodeJSONOptional`) using `http.MaxBytesReader`: streaming (bounded memory), `413 payload_too_large` on over-limit, `400 invalid_request` on parse errors, strict single-value check preserved. A zero-value `Config` falls back to the default limit.
- **Multimodal endpoints** use the higher limit: `/v1/chat/completions`, `/v1/responses`, responses-compact, Anthropic `/v1/messages`, and admin playground (chat + stream). `/v1/embeddings` and admin endpoints keep the default. Call sites carrying intentional custom error messages are preserved.
- **Tests:** `getenvBytes`/`parseByteSize`, plus under-limit / over-limit-413 / malformed-400 / empty-body / multimodal-limit decode behavior.
- **Env & docs sync:** both vars added to `backend/.env.example`, `deploy/.env.example`, `deploy/docker-compose.yml`, `start.sh`, and en/zh/ja deployment docs (with a reverse-proxy `client_max_body_size` alignment note). `docs/superpowers/` marked English-only for the translation gate.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor or maintenance
- [ ] Documentation

## Verification

- `gofmt -l` clean; `go vet ./...` clean; `go test ./...` passes (full backend suite).
- Repo gates: `check-env-contract`, `check-doc-translations` (incl. `--base/--head` co-change mode), `check-source-lines`, and `node --test tools/*.test.mjs` (102/0) all pass.
- `docker compose --env-file deploy/.env.example -f deploy/docker-compose.yml config` renders.
- `git diff --check` clean.

- **API:** `/v1` contract unchanged for well-formed under-limit requests. Over-limit now returns `413 payload_too_large` (previously a generic `400`). Endpoints with bespoke validation messages keep their `400`.
- **Security/Ops:** higher limits raise worst-case buffered bytes per in-flight request; mitigated by streaming decode, the 512 MiB hard ceiling, and conservative defaults (8/32 MiB). Operators fronting the gateway should set the reverse-proxy `client_max_body_size` at least as large as `TOKENHUB_MAX_MULTIMODAL_REQUEST_BYTES`. Rollback: unset the env vars (defaults apply) or revert the commit.

## Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)


deploy/docker-compose.yml
```diff
@@ -42,6 +42,8 @@ services:
       TOKENHUB_TRACING_QUEUE_SIZE: ${TOKENHUB_TRACING_QUEUE_SIZE:-2048}
       TOKENHUB_UPSTREAM_NON_STREAM_TIMEOUT_SECONDS: ${TOKENHUB_UPSTREAM_NON_STREAM_TIMEOUT_SECONDS:-120}
       TOKENHUB_UPSTREAM_STREAM_IDLE_TIMEOUT_SECONDS: ${TOKENHUB_UPSTREAM_STREAM_IDLE_TIMEOUT_SECONDS:-300}
+      TOKENHUB_MAX_JSON_REQUEST_BYTES: ${TOKENHUB_MAX_JSON_REQUEST_BYTES:-8388608}
+      TOKENHUB_MAX_MULTIMODAL_REQUEST_BYTES: ${TOKENHUB_MAX_MULTIMODAL_REQUEST_BYTES:-33554432}
```
astaxie, line 46: [P1] Propagate the limits through every documented deployment topology

These mappings were added only to `docker-compose.yml`. Both documented production variants, `deploy/docker-compose.postgres.yml` and `deploy/docker-compose.remote-postgres.yml`, define their own `tokenhub-backend.environment` blocks and omit both new variables; `--env-file` only supplies Compose interpolation and does not inject unreferenced values. I rendered both variants and the `TOKENHUB_MAX_*` entries are absent, so operator settings silently fall back to 8/32 MiB. The remote topology also mounts `nginx.multi-instance.conf`, which still hard-codes `client_max_body_size 32m`, so any multimodal value above the default is rejected before TokenHub. Please wire both variables into both PostgreSQL Compose files and make the bundled proxy ceiling align with the configured limit (or provide an explicit supported override), with rendered-config coverage for all three topologies.

samz406:
All review feedback has been addressed in commit 5a31f8a. Summary of fixes:

## Build failure (merge conflict)
✅ Merged `origin/main` (3f1d968..f0d3cca) and migrated the new `analytics_http.go` call site to `s.decodeJSON`. Build now passes.

## Code fixes (Copilot inline comments)

**config.go — `parseByteSize` malformed suffix acceptance**
✅ Fixed. Now splits digits from suffix explicitly and validates against the exact documented set (`k`/`kb`/`kib`, `m`/`mb`/`mib`, `g`/`gb`/`gib`). Rejects `8kk`, `8ib`, `8big`. Added regression tests in `config_test.go`.

**http_transport.go:101 — multimodal endpoints falling back to 8 MiB**
✅ Fixed. `NewWithConfig` now normalizes both limit fields so `New(store)` and zero-value configs get the tier-specific defaults (8 MiB global, 32 MiB multimodal) rather than collapsing to 8 MiB via the `decodeJSONLimit` fallback.

**gemini_native_http.go — Gemini endpoints using default limit**
✅ Fixed. Both `generateContent` and `countTokens` now use `s.decodeJSONLimit(..., s.config.MaxMultimodalRequestBytes)` since they accept base64 `inlineData`.

**provider_account_quota_reset.go:109 — audit records `invalid_request` for 413**
✅ Fixed. Audit now records `AsHTTPError(err).Code` so `payload_too_large` appears in logs when the decoder returns 413.

**Design/plan docs — stale paths and behavior description**
✅ Fixed. Corrected the pre-change behavior description (it used `io.ReadAll` + explicit length check, not silent truncation). Updated all file paths: `http.go` → `http_transport.go`, `decode_test.go` → `decode_json_test.go`. Added a reconciliation note to the plan doc.

## Deploy fixes (astaxie + Copilot)

**docker-compose.yml:46 / postgres / remote-postgres topologies**
✅ Verified. Codex's earlier commits propagated both limit vars to all three topologies. CI now asserts they render in every topology.

**nginx.multi-instance.conf — fragile `client_max_body_size` substitution**
✅ Fixed. Introduced a dedicated nginx-format variable `TOKENHUB_NGINX_CLIENT_MAX_BODY_SIZE` (default `32m`) for the load balancer's `client_max_body_size`. This decouples the proxy ceiling from the backend's byte-format variable so backend-valid values like `32mb`/`32mib` never render as invalid nginx syntax. Registered in `KNOWN_CONSUMERS`, documented in en/zh/ja deployment docs, and asserted in CI.

## Verification
- gofmt / go vet / go build / **full backend test suite** ✅
- All repo gates (env-contract, doc-translations, source-lines, tools tests) ✅
- All three compose topologies render ✅
- `git diff --check` clean ✅

CI is running: Repository gates and Deployment already passed.

deploy/docker-compose.yml
samz406: done

astaxie:
删除 superpowers 的文档目录

backend/internal/server/billing_http.go
```diff
@@ -32,7 +31,7 @@ func (s *Server) handleAdminBillingConnectors(w http.ResponseWriter, r *http.Req
 		writeJSON(w, http.StatusOK, map[string]any{"data": s.store.ListBillingConnectors()})
 	case http.MethodPost:
 		var request BillingConnectorRequest
-		if err := decodeJSON(r, &request); err != nil {
+		if err := s.decodeJSON(w, r, &request); err != nil {
```
astaxie, line 34: [P2] Preserve 413 for over-limit request bodies

`decodeJSON` now returns a typed `payload_too_large` error, but this handler replaces every decode error with `400 invalid_billing_connector`. The same pattern remains at `billing_http.go:85`, `billing_http.go:136`, `reconciliation_http.go:28`, `reconciliation_http.go:60`, `reconciliation_http.go:93`, and `version_update.go:704`. Those endpoints therefore do not honor the PR’s clear-413 contract. Please pass through `413/payload_too_large` and keep the bespoke 400 errors only for malformed or invalid payloads.

backend/internal/server/config.go
```diff
@@ -400,3 +405,81 @@ func getenvBool(key string, fallback bool) bool {
...
+	case "g", "gb", "gib":
+		multiplier = 1 << 30
+	default:
+		return 0, false
+	}
+	n, err := strconv.ParseInt(digits, 10, 64)
+	if err != nil || n < 0 {
+		return 0, false
+	}
+	if multiplier > 1 && n > (int64(1)<<62)/multiplier {
```
astaxie: [P3] Check multiplication against the actual int64 maximum

The `1<<62` threshold rejects some values whose multiplied byte count still fits in `int64`. For example, `5000000000g` is representable, but this branch returns `false`, so `getenvBytes` silently falls back to 8/32 MiB instead of applying the documented 512 MiB clamp. Please compare against `math.MaxInt64/multiplier` and add a regression case for a representable value in this range.

astaxie (commented):
Two follow-up findings from the merge-readiness review.

samz406:
已删除 docs/superpowers/ 目录（设计文档和实施计划两个文件），并回退了翻译门禁里对应的 docs/superpowers/ 豁免配置。提交 5ab621b 已推送。

astaxie (commented):
Re-review of current head 5ab621b: this is not merge-ready yet. The latest commit removes docs/superpowers, but it does not address the two findings from the 5a31f8a review. The billing, reconciliation, and rollback handlers still replace payload_too_large from s.decodeJSON with bespoke 400 errors, and parseByteSize still checks overflow against (1<<62)/multiplier rather than math.MaxInt64/multiplier. The green CI run therefore does not cover the outstanding 413 contract and representable-large-value cases. Please fix both current-head threads and add the requested regressions.

samz406:
Both merge-readiness findings and the two additional Copilot findings are addressed in commit c894882 (also merged latest main / v0.5.0).

## P2 — Preserve 413 for over-limit bodies
✅ Fixed. Added a shared `isPayloadTooLarge(err)` helper; the bespoke-400 handlers now let the typed `413 payload_too_large` pass through and keep their 400 only for malformed/invalid payloads. Applied at all named sites: `billing_http.go` (connector create/patch + sync), `reconciliation_http.go` (rule create/patch + run), and `version_update.go` (rollback). The audit-recording paths (reconciliation run, rollback) now record `payload_too_large` in the audit event when the body is over-limit, so logs match the response.
- Regression: `TestBillingConnectorOverLimitReturns413` (end-to-end, asserts 413 + `payload_too_large`) and `TestIsPayloadTooLarge`.

## P3 — Overflow guard against the real int64 max
✅ Fixed. `parseByteSize` now checks `math.MaxInt64/multiplier` instead of `(1<<62)/multiplier`, so representable values above the ceiling reach the documented 512 MiB clamp instead of falling back.
- Regression: `5000000000g` added to both `TestParseByteSize` (parses) and `TestGetenvBytes` (clamps to ceiling).

## Copilot — Anthropic 413 misclassification
✅ Fixed. `anthropicErrorType` now maps `413` to `request_too_large` instead of the generic `api_error`, so `/v1/messages` over-limit responses use the correct Anthropic error type.
- Regression: `TestAnthropicErrorTypeForPayloadTooLarge`.

## Copilot — misleading memory comment
✅ Fixed. Corrected the `decodeJSONLimit` comment: `json.Decoder` buffers the full top-level value, so `MaxBytesReader`'s guarantee is capping total bytes read (rejecting over-limit early), not bounding peak memory below the body size.

## Verification
- gofmt / go vet / go build / full backend test suite ✅
- All repo gates (env-contract, doc-translations incl. PR-mode co-change, ui-translations, source-lines, tools tests) ✅
- All three compose topologies render ✅
- `git diff --check` clean ✅

astaxie (approved):
Current head c894882 reviewed. The outstanding 413 propagation and int64 overflow findings are fixed with regressions, the remaining current-line bot concerns are already addressed by config normalization and typed audit recording, and all available CI plus full applicable local backend/frontend/repository/Compose checks pass.

PR #149, opened 2026-08-06
fix(server,migration): correctness and robustness fixes

susunola:
## Summary

Fixes a collection of correctness and robustness bugs found in the backend server and migration tooling, and addresses the five review findings on the original implementation gaps.

## Related Issue

N/A

## Changes

- Use escaped path parsing for admin provider and billing connector endpoints so IDs containing `%2F` are handled correctly, and decode the single provider-catalog segment so a catalog ID containing `/` is looked up under its real value.
- Mark external model roles only after route create/update succeeds to avoid catalog inconsistency when the store operation fails; the store now exposes `CreateRoute`, which propagates the database write error.
- Reject requests when the Anthropic adapter is missing instead of falling back to an unconfigured HTTP client.
- Replace `sync.Once` in the billing scheduler with explicit run-state tracking so `StartScheduler` works after `Shutdown`.
- Make the Aliyun billing range filter inclusive on the upper boundary and reject invalid upstream decimal amounts, including discount components, instead of silently coercing them to zero.
- Preserve precision for large integer OneAPI billing fields by parsing through `Int64` instead of `float64`.
- Give the migration Admin API client an explicit timeout, validate the base URL in the constructor so a malformed `--to` value is a controlled error instead of a panic, and cap response body size; the production CLI now passes `nil` so it receives the default timeout.
- Trim whitespace from migration target values and secret-file values.
- Reject negative schema version components and skip LiteLLM users with empty emails that would fail bundle validation.

## Type of Change

- [x] Bug fix

## Verification

- `gofmt -w` on all changed Go files
- `go vet ./...` passes
- `go test ./...` passes
- `git diff --check` passes
- Regression tests added for: malformed base URL rejection and default client timeout, escaped-path endpoint encoding, response body cap, provider-catalog `%2F` ID decoding, route create failure ordering, missing Anthropic adapter, OneAPI large integers, and billing scheduler restart.
- Note: the Frontend and Migration Unit Tests failures seen on the previous head were GitHub Actions infrastructure errors (`Failed to resolve action download info: Service Unavailable`), not code failures; local `go test ./...` and `go vet ./...` pass on this head.

None.

## Checklist


backend/internal/migration/sink/tokenhub/http_client.go
```diff
@@ -44,7 +46,10 @@ func NewAdminAPIClient(baseURL string, token string, httpClient *http.Client) *A
 }
 
 func (c *AdminAPIClient) endpoint(parts ...string) string {
-	base, _ := url.Parse(c.baseURL)
+	base, err := url.Parse(c.baseURL)
+	if err != nil || base == nil {
+		panic(fmt.Sprintf("invalid admin api base url %q: %v", c.baseURL, err))
```
astaxie: A malformed user-supplied `--to` value now terminates the migration process via `panic`. The PR says malformed base URLs are guarded, but this should be a controlled CLI/request error. Please validate in the constructor or make endpoint construction return an error and propagate it through the client methods.

backend/internal/server/admin_providers_http.go
```diff
@@ -82,7 +82,7 @@ func (s *Server) handleAdminProviderCatalogItem(w http.ResponseWriter, r *http.R
 	if _, ok := s.requireAdmin(w, r, "provider", r.Method); !ok {
 		return
 	}
-	id := strings.Trim(strings.TrimPrefix(r.URL.Path, "/api/admin/provider-catalog/"), "/")
+	id := strings.Trim(strings.TrimPrefix(r.URL.EscapedPath(), "/api/admin/provider-catalog/"), "/")
```
astaxie: Switching to `EscapedPath` preserves `%2F`, but this handler never decodes the segment. A catalog ID such as `foo/bar` arrives here as `foo%2Fbar` and is then looked up under the encoded value, so the stated escaped-ID fix is incomplete. Please unescape the single segment (and reject malformed escapes), as the other updated handlers do.

backend/internal/server/billing_aliyun.go
```diff
@@ -200,17 +200,29 @@ func normalizeAliyunBillingRecord(item map[string]any, connector BillingConnecto
 	if err != nil {
 		usageEnd = usageStart
 	}
-	gross := billingDecimalValue(firstValue(item, "PretaxGrossAmount", "pretax_gross_amount", "PretaxAmount"))
+	gross, err := billingDecimalValue(firstValue(item, "PretaxGrossAmount", "pretax_gross_amount", "PretaxAmount"))
+	if err != nil {
+		return BillingRecord{}, err
+	}
 	discount := billingDecimalAdd(
```
astaxie: Invalid discount fields are still silently accepted here. `billingDecimalAdd` skips every value that `big.Rat.SetString` cannot parse, so an invalid `InvoiceDiscount` or coupon amount becomes an understated discount instead of rejecting the upstream record. Please make the addition helper return an error and propagate it.

backend/internal/migration/sink/tokenhub/http_client.go
```diff
@@ -34,7 +36,7 @@ func (e *ApprovalRequiredError) Error() string {
 
 func NewAdminAPIClient(baseURL string, token string, httpClient *http.Client) *AdminAPIClient {
 	if httpClient == nil {
```
astaxie, line 52: This timeout branch is not used by the production migration CLI: `newHTTPSink` still passes `http.DefaultClient`, whose total timeout is zero. Because the CLI also uses `context.Background`, requests can still hang indefinitely. Please pass `nil` from the CLI or otherwise ensure the production client receives the intended timeout.

backend/internal/server/admin_providers_http.go
```diff
@@ -1004,11 +1004,16 @@ func (s *Server) handleAdminRoutes(w http.ResponseWriter, r *http.Request) {
 			writeError(w, r, NewHTTPError(http.StatusConflict, "model_route_conflict", "This external model is already mapped to the selected provider model"))
 			return
 		}
+		route := s.store.AddRoute(req)
```
astaxie: This still does not guarantee that route creation succeeded before marking the model. `GormStore.AddRoute` discards the error returned by `createRouteRecord` and returns a generated route even when the database write fails, so `markExternalModel` still runs and preserves the catalog inconsistency this change is meant to fix. The store API needs to return and propagate the create error before this call.

astaxie (changes requested):
Thanks for the robustness pass. I found five implementation gaps where the behavior described in the PR is still incomplete. The most important ones are that the migration CLI still uses an HTTP client without a total timeout, and route creation can still mark a model external after the database write failed.

Repository standards also require regression tests for backend behavior changes. The new timeout/URL/response-cap behavior, escaped-ID handlers, route failure ordering, missing Anthropic adapter, OneAPI large integers, and scheduler restart path are not covered. The PR body also omits required template sections and does not report the full applicable validation set.

Local verification at this commit passed `go test ./...`, `go vet ./...`, repository gates, and `git diff --check`. The current Frontend and Migration Unit Tests failures are GitHub Actions infrastructure failures (`Failed to resolve action download info: Service Unavailable`), not code failures.

backend/internal/migration/sink/tokenhub/http_client.go
susunola: Fixed in aa91897: NewAdminAPIClient now validates the base URL in the constructor and returns an error (scheme + host required), and endpoint() uses the pre-parsed URL so it can no longer panic. The CLI surfaces the constructor error before any request is sent.

backend/internal/server/admin_providers_http.go
susunola: Fixed in aa91897: handleAdminProviderCatalogItem now unescapes the single segment with url.PathUnescape and rejects malformed escapes before lookup, matching the other escaped handlers. Regression test added (catalog entry with a '/' in its ID resolved via %2F).

backend/internal/server/billing_aliyun.go
susunola: Fixed in aa91897: billingDecimalAdd now returns an error for any present-but-unparseable component, and normalizeAliyunBillingRecord propagates it instead of understating the discount. Regression test added.

backend/internal/migration/sink/tokenhub/http_client.go
susunola: Fixed in aa91897: the production CLI now passes nil to NewAdminAPIClient so the 30s default total timeout applies, and newHTTPSink returns the constructor error instead of building a sink with a zero-timeout client. Test added for the default timeout.

backend/internal/server/admin_providers_http.go
susunola: Fixed in aa91897: GormStore.CreateRoute propagates the createRouteRecord error (AddRoute is unchanged for best-effort callers); the admin route handler now returns the write error before markExternalModel. Regression test added using a forced SQLite trigger failure on the routing-rules endpoint.

astaxie (commented):
Re-reviewed the complete current diff at aa91897. The five requested fixes are present with regression coverage, all five threads are resolved, and full backend tests, go vet, repository tests/gates, translation/environment/source-line checks, and diff check pass locally. GitHub has no CI checks for the current head, which is a merge blocker under the repository policy; no approval until the required workflow runs and succeeds.

susunola:
Rebased the branch onto the current `astaxie/TokenHub` main (merge commit `2238279`) so this pull request head is no longer 118 commits behind the base.

This also explains why the previous head had no CI checks at all: the branch was still carrying an older `.github/workflows/ci.yml` that predated the `Backend PostgreSQL` job and the `EXPECTED_JOBS: 5` aggregate gate. The workflow definitions on this head are now identical to upstream, so the required `CI` aggregate check can resolve correctly.

Local verification on the merged head:

- `gofmt -l backend` — clean
- `go vet ./...` — passes
- `go test ./...` — passes
- `node --test tools/*.test.mjs` — 117 tests, 0 failures
- `node tools/check-doc-translations.mjs` — passes
- `node tools/check-ui-translations.mjs` — passes (0 new literal keys)
- `node tools/check-env-contract.mjs` — passes (68 backend variables)
- `node tools/check-source-lines.mjs` — passes
- `git diff --check` — passes

The merge was conflict-free, and this branch only touches backend files, so none of the upstream frontend work is affected.

@astaxie the `CI` and `Migration CI` runs for this head are currently sitting in `Action required` because the pull request comes from a fork. Could you approve the pending workflow runs so the required checks can execute? Nothing further is blocking from my side.

astaxie (commented):
Re-reviewed the complete current diff at db77697. The clean upstream merge preserves the verified correctness fixes, and the bot-generated contributor refresh updates all three README variants consistently. Local go test/go vet, repository tests and gates, full documentation translation diff check, UI/environment/source-line checks, and git diff --check all pass. However, this exact head currently has zero GitHub check runs/statuses, so missing required CI blocks approval; because db77697 was generated by github-actions[bot], please ensure the required workflows run on this exact SHA through a trusted author/maintainer update or another trusted path. Also, the PR title summary “correctness and robustness fixes” is a noun phrase; please rename it to an imperative summary such as “improve correctness and robustness” to satisfy the repository PR-title rule.

PR #162, opened 2026-08-08
fix(providers): validate custom provider base URL against SSRF

deepjerry-ai:
## Summary

Custom provider upstream base URLs were accepted without any validation. An admin (or a compromised admin session) could point the gateway at link-local cloud metadata endpoints (e.g. 169.254.169.254, Alibaba Cloud 100.100.100.200) or arbitrary internal services, turning provider requests into SSRF primitives.

This PR adds defense-in-depth validation at save time and at connection time for **all** provider traffic (model discovery, chat/streaming inference, and Codex subscription calls), plus an explicit operator allowlist for in-house model servers on private networks.

## Related Issue

N/A

## Changes

**Persistence-time validation (create and update)**
- `providerFromCreateRequest` validates the normalized base URL, so `POST /api/admin/providers` and PATCH can no longer save a disallowed URL.
- Resource-level URLs are validated too (`ProviderResource.BaseURL` overrides the provider URL in `routeSelection`): create, update and import paths all enforce the same guard, with per-row failure semantics on import.

**Connection-time guards (all provider HTTP clients)**
- Streaming, non-streaming and Codex subscription inference clients all dial through `ssrfGuardedProviderTransport` and follow the shared strict redirect policy (`strictProviderUpstreamRedirect`).
- Dial-time DNS resolution filtering closes DNS rebinding: only validated addresses are dialed inside one shared resolution+dial deadline, with staggered fallback between candidates (a stalled resolver or one black-holing address cannot starve healthy candidates or multiply the timeout).
- Proxying is disabled so the checks cannot be bypassed via proxy environment variables.
- Redirect re-validation runs in strict mode: no private-range allowlist and **no localhost exception**, so a public URL can never bounce requests into the internal network or onto a loopback service.

**Address classification**
- Rejects trust-boundary-sensitive destinations: private (RFC1918/ULA), loopback, link-local/metadata, multicast, unspecified and CGNAT, plus a curated set of ranges that are unsafe or never meaningful provider endpoints (documentation/benchmarking/reserved ranges, deprecated 6to4 relay-anycast `192.88.99.0/24`, deprecated IPv6 site-local, IPv6 discard/dummy and SRv6 SID ranges). This is intentionally not an exhaustive mirror of the IANA special-purpose registries; deployment-specific reachability must be constrained by network egress policy.
- RFC 6052 NAT64 addresses use their embedded IPv4 target for the same decision when the well-known prefix or an operator-configured network-specific prefix is in use. IPv4-mapped IPv6 spellings are classified through To4.
- URL scheme restricted to http/https; embedded credentials rejected.

**Operator allowlist**
- `TOKENHUB_PROVIDER_UPSTREAM_ALLOWED_CIDRS` (comma-separated CIDRs) allows literal private IPs for in-house model servers. It applies **only** to administrator-typed literal IPs in RFC1918/ULA ranges — never to resolved hostnames or redirect targets — and can never allowlist link-local/metadata or curated high-risk/non-provider ranges. Loopback provider URLs are rejected by default; local Ollama/LM Studio development requires `TOKENHUB_PROVIDER_UPSTREAM_ALLOW_LOOPBACK=true`. Redirect hops always reject loopback.
- Synchronized across `deploy/.env.example`, `backend/.env.example`, the Compose files, `start.sh`, and deployment docs (EN/zh-CN/ja).

**Tests** — save-time create/update rejection, resource-level create/update/import rejection, allowlisted literal acceptance, strict-mode loopback rejection, redirect-to-loopback/private/non-http rejection, dial-time literal classification (network-free), guarded inference-client wiring, and category coverage for the curated classifier, including save/dial regression cases for `192.88.99.1` and `.2`.

## Type of Change

- [x] Bug fix

## Verification

- `go test ./...` — pass
- `go vet ./...` — pass
- `golangci-lint run` — 0 issues
- `node --test tools/*.test.mjs` — pass
- `node tools/check-doc-translations.mjs --base aebf538c55aa04042058b828a36a5960d22ef217 --head HEAD` — pass
- `node tools/check-ui-translations.mjs` — pass
- `node tools/check-env-contract.mjs` — pass
- `node tools/check-source-lines.mjs` — pass
- `docker compose --env-file deploy/.env.example -f deploy/docker-compose.yml config` — renders successfully
- `git diff --check` — pass

Security-sensitive. Previously accepted malicious or intranet base URLs are now rejected at save time and blocked at connection time; legitimate public endpoints are unaffected; localhost development providers require the explicit `TOKENHUB_PROVIDER_UPSTREAM_ALLOW_LOOPBACK` opt-in. Operators with in-house model servers on private networks must set `TOKENHUB_PROVIDER_UPSTREAM_ALLOWED_CIDRS` (e.g. `192.168.0.0/16`) and use literal IPs from those ranges. The escape hatch is deliberately scoped to literal IPs; hostname-based private routing is outside this bug fix. Default behavior (empty allowlist) is the strictest posture. The application classifier is defense in depth rather than a substitute for firewall, security-group, NetworkPolicy, or trusted egress-proxy controls. No API contract change for valid configurations; no migration impact.

## Checklist

- [x] Shared user-facing behavior is documented consistently in English, Simplified Chinese, and Japanese where applicable. (Deployment variables documented in all three languages; no UI copy changes.)
- [x] `data/model-catalog.yaml` remains tracked and catalog changes were reviewed where applicable. (No catalog changes.)


backend/internal/server/provider_catalog.go
```diff
@@ -522,9 +524,10 @@ func CustomProviderCatalogFromUpstream(ctx context.Context, client *http.Client,
 	if err != nil || endpoint.Scheme == "" || endpoint.Host == "" {
 		return ProviderCatalogEntry{}, NewHTTPError(http.StatusBadRequest, "provider_base_url_invalid", "Base URL is invalid")
 	}
-	if client == nil {
-		client = http.DefaultClient
+	if err := validateProviderUpstreamBaseURL(endpoint, allowedProviderUpstreamCIDRs()); err != nil {
+		return ProviderCatalogEntry{}, err
 	}
+	client = ssrfGuardedProviderClient(client)
```
astaxie, line 528: [P1] This guard is installed only inside `CustomProviderCatalogFromUpstream`, so it protects the `/models` discovery/test request rather than provider persistence or real inference traffic. `POST /api/admin/providers` and PATCH still reach `providerFromCreateRequest`, which only normalizes `BaseURL`; I reproduced a direct POST with `http://169.254.169.254/latest/meta-data` returning 201. Runtime calls are also built by `newUpstreamClients` with ordinary default transports, so DNS rebinding, redirects, and proxy resolution remain available after save. Please validate create/update and apply the guarded transport to both streaming and non-streaming inference clients.

backend/internal/server/provider_catalog.go
```diff
@@ -574,6 +577,263 @@ func CustomProviderCatalogFromUpstream(ctx context.Context, client *http.Client,
...
+func ssrfGuardedProviderClient(client *http.Client) *http.Client {
+	guard := &http.Client{
+		// Re-validate every redirect target before following it.
+		CheckRedirect: func(req *http.Request, via []*http.Request) error {
+			if len(via) >= 10 {
+				return fmt.Errorf("stopped after 10 redirects")
+			}
+			// Redirect targets never get the private-range allowlist: a public
+			// URL must not be able to bounce requests into the internal network.
+			return validateProviderUpstreamBaseURL(req.URL, nil)
```
astaxie: [P1] Passing `nil` here disables only the private-CIDR allowlist; `validateProviderUpstreamBaseURL` still returns success for `localhost`, `127.0.0.1`, and `::1` at lines 612-613, and the guarded dialer connects to them. A public endpoint can therefore redirect to a loopback service. The new `TestCustomProviderCatalogFromUpstreamAllowsRedirectToPublic` currently demonstrates this with a 302 to `127.0.0.1`. Redirect validation needs a strict mode that also disables the localhost exception.

backend/internal/server/provider_catalog_ssrf_test.go
```diff
@@ -0,0 +1,400 @@
...
+	// hostname resolution never get the private-range allowlist).
+	return validateProviderUpstreamBaseURL(endpoint, nil)
+}
+
+func TestValidateProviderUpstreamBaseURLAllowsPublicHTTPS(t *testing.T) {
+	for _, raw := range []string{
+		"https://api.example.com/v1",
+		"https://api.openai.com/v1",
+		"http://api.example.com:8080/v1",
+		"https://203.0.113.10/v1",
```
astaxie: [P2] `203.0.113.0/24` is IANA TEST-NET-3 special-purpose space, so accepting it conflicts with the stated claim that reserved ranges are rejected. The denylist also omits other relevant special-use ranges such as `192.0.2.0/24`, `198.18.0.0/15`, `198.51.100.0/24`, IPv4 multicast, and IPv6 documentation space. Please classify the full set of non-routable/special-purpose ranges you intend to reject and add matching save- and dial-time tests.

backend/internal/server/provider_catalog.go
deepjerry-ai: Fixed in 8b9d95e. Save-time validation now lives in `providerFromCreateRequest`, so both `POST /api/admin/providers` and PATCH are guarded (covered by `TestAdminCreateProviderRejectsSSRFBaseURL` and `TestAdminPatchProviderRejectsSSRFBaseURL`, which assert the metadata URL now returns 400). The inference path is guarded too: `newUpstreamClients` builds both the streaming and non-streaming clients on `ssrfGuardedProviderTransport` (dial-time DNS filtering, proxy disabled, one shared resolution+dial budget with per-candidate fallback slices) plus the strict redirect policy, and the Codex subscription adapter gets the same transport. Resource-level URLs (`routeSelection` overrides `Provider.BaseURL` with `ProviderResource.BaseURL`) are validated on create/update/import as well.

backend/internal/server/provider_catalog.go
deepjerry-ai: Fixed in 8b9d95e. `validateProviderUpstreamBaseURL` now takes an `allowLocalhost` flag, and redirect re-validation (`strictProviderUpstreamRedirect`, shared by the catalog guard and all inference clients) passes `nil, false` — so redirect hops reject loopback in every spelling (`localhost`, `127.0.0.1`, `::1`) as well as allowlisted private ranges. The test you cited was inverted: `TestCustomProviderCatalogFromUpstreamRejectsRedirectToLoopback` now asserts a 302 onto `127.0.0.1` fails, and `TestValidateProviderUpstreamBaseURLStrictModeRejectsLoopback` pins the unit behavior.

backend/internal/server/provider_catalog_ssrf_test.go
deepjerry-ai: Fixed in 8b9d95e. The special-use classification now covers the documentation/benchmarking ranges (192.0.2.0/24, 198.51.100.0/24, 203.0.113.0/24, 198.18.0.0/15, 2001:db8::/32), IPv4 and IPv6 multicast (via `IsMulticast`), limited broadcast, the deprecated IPv6 site-local range (fec0::/10), and the NAT64 translation prefixes (64:ff9b::/96 and local-use 64:ff9b:1::/48, which can encode RFC1918 targets). The public-literal test case now uses 8.8.8.8 instead of 203.0.113.10, and save-time plus dial-time tests cover each added range.

backend/internal/server/provider_catalog.go
```diff
@@ -574,6 +578,341 @@ func CustomProviderCatalogFromUpstream(ctx context.Context, client *http.Client,
...
+		"0.0.0.0/8",
+		"100.64.0.0/10",
+		"192.0.2.0/24",
+		"198.18.0.0/15",
+		"198.51.100.0/24",
+		"203.0.113.0/24",
+		"240.0.0.0/4",
+		"2001:db8::/32",
+		"fec0::/10",
+		"64:ff9b::/96",
```
astaxie: [P1] Blocking the entire well-known NAT64 prefix breaks valid public upstreams on IPv6-only DNS64 networks. For example, `64:ff9b::808:808` is the standard synthesis of public `8.8.8.8`, but `checkProviderUpstreamLiteralDial` rejects it before dialing; now that every inference transport uses this classifier, IPv4-only providers become unreachable in those deployments. Please decode the embedded IPv4 for `64:ff9b::/96` and apply the existing private/special-use classification to that address, rejecting unsafe targets while allowing public ones. The local-use prefix can retain a stricter policy. Add a regression case for a synthesized public IPv4 address.

backend/internal/server/provider_catalog.go
```diff
@@ -574,6 +578,341 @@ func CustomProviderCatalogFromUpstream(ctx context.Context, client *http.Client,
...
+		}
+		if len(allowed) == 0 {
+			return nil, fmt.Errorf("provider base URL host %q resolves only to disallowed addresses", host)
+		}
+		// Dial each validated address directly, falling back to the next one, so
+		// the connected address is always one that passed the check. Every
+		// candidate gets an even slice of the remaining budget: one silently
+		// black-holing address must not starve the healthy candidates behind
+		// it (the default transport's Happy Eyeballs behavior this replaces).
+		budget := dialer.Timeout / time.Duration(len(allowed))
```
astaxie: [P2] This slice is based on the original 30-second timeout rather than the time remaining after DNS resolution. If lookup consumes 20 seconds and returns two addresses, the first black-holing address receives a nominal 15-second child context but is capped by the 10 seconds left on `dialCtx`, consuming the entire shared deadline before the healthy second candidate is attempted. That violates the stated guarantee that a stalled resolver or candidate cannot starve later addresses. Please recompute each slice inside the loop from the parent deadline and the number of candidates remaining, with a focused slow-DNS plus black-hole-first regression test.

backend/internal/server/provider_catalog.go
deepjerry-ai: Fixed. `64:ff9b::/96` is no longer blocked wholesale: `isDisallowedProviderUpstreamIP` decodes the embedded IPv4 from the well-known prefix and recursively applies the private/special-use classification to that address, so a synthesized public target like `64:ff9b::808:808` (= 8.8.8.8) is dialed while private/loopback/metadata syntheses are rejected. The local-use prefix `64:ff9b:1::/48` keeps the stricter whole-range rejection. Regression coverage: `64:ff9b::808:808` is in the allowed-literal test list, `64:ff9b::a00:1` (10.0.0.1 synthesis) and IPv4-mapped forms are in the rejection tests (`TestCheckProviderUpstreamLiteralDialAllowlist` and the base-URL validation cases). The helpers moved to `provider_upstream_ssrf.go` in the latest commit to stay under the source-line gate.

backend/internal/server/provider_catalog.go
deepjerry-ai: Fixed by replacing the slice-per-candidate approach entirely: validated candidates now dial in a staggered race inside the shared budget (first candidate immediately, later ones join after a 300ms fallback delay or as soon as an earlier attempt fails, first success wins, losers are cancelled and close their own connections). A stalled resolver or black-holing first candidate can no longer starve later addresses, and there is no fixed per-candidate share to recompute. Regression test: `TestDialGuardedUpstreamFallbackAfterSlowLookup` covers exactly the slow-DNS (500ms lookup) + black-hole-first scenario and asserts the healthy second candidate wins well inside the budget; `TestRaceValidatedUpstreamCandidatesClosesLosingConnections` and `TestRaceValidatedUpstreamCandidatesDrainsAfterCancellation` pin the race teardown. The code moved to `provider_upstream_ssrf.go` in the latest commit.

backend/internal/server/provider_upstream_ssrf.go
```diff
@@ -0,0 +1,493 @@
...
+
+// strictProviderUpstreamRedirect is the CheckRedirect policy shared by every
+// SSRF-guarded client: re-validate each hop with no private-range allowlist
+// and no localhost exception, so a validated URL can never redirect into the
+// internal network or onto a loopback service.
+func strictProviderUpstreamRedirect(req *http.Request, via []*http.Request) error {
+	if len(via) >= 10 {
+		return fmt.Errorf("stopped after 10 redirects")
+	}
+	return validateProviderUpstreamBaseURL(req.URL, nil, false)
```
astaxie: [P1] The redirect guard validates only the destination address class, so cross-origin public redirects and HTTPS-to-HTTP downgrades still pass. Go strips `Authorization` on unrelated hosts, but it copies custom headers such as `x-api-key`, `api-key`, and provider-defined headers; a redirected Anthropic request can therefore disclose its credential to the target host. Please reject redirects whose scheme or host differs from the original request, and reject HTTPS downgrades, or explicitly remove every credential and custom provider header before following the redirect.

backend/internal/server/provider_upstream_ssrf.go
```diff
@@ -0,0 +1,493 @@
...
+		"0.0.0.0/8",
+		"100.64.0.0/10",
+		"192.0.2.0/24",
+		"198.18.0.0/15",
+		"198.51.100.0/24",
+		"203.0.113.0/24",
+		"240.0.0.0/4",
+		"2001:db8::/32",
+		"fec0::/10",
+		"64:ff9b:1::/48",
```
astaxie: [P2] Rejecting the entire RFC 8215 local-use NAT64 allocation `64:ff9b:1::/48` also blocks valid DNS64 synthesis of public provider addresses. On IPv6-only deployments using this prefix, hostname resolution produces candidates in this range and the dial guard filters all of them, making every provider unreachable. Please decode and classify the embedded IPv4 according to the configured prefix format, or make the deployment NAT64 prefix configurable, instead of categorically denying the whole `/48`.

backend/internal/server/provider_upstream_ssrf.go
```diff
@@ -0,0 +1,474 @@
...
+		return NewHTTPError(http.StatusBadRequest, "provider_base_url_invalid_scheme", "Base URL must use http or https")
+	}
+	if endpoint.User != nil {
+		return NewHTTPError(http.StatusBadRequest, "provider_base_url_invalid", "Base URL must not embed credentials")
+	}
+	host := endpoint.Hostname()
+	if host == "" {
+		return NewHTTPError(http.StatusBadRequest, "provider_base_url_invalid", "Base URL is invalid")
+	}
+	if isLocalProviderHostname(host) {
```
astaxie, line 53: **[P1] Do not leave loopback enabled for every provider**

The threat model in this PR explicitly includes a compromised admin session, but this branch accepts localhost, 127.0.0.1, and ::1 whenever allowLocalhost is true. Every provider/resource persistence caller passes true, and dialGuardedUpstream independently bypasses filtering for the same hosts, so an attacker can still create any adapter or resource that targets an arbitrary loopback service and the production clients will connect to it. That is much broader than the documented Ollama/LM Studio compatibility exception and preserves the original SSRF primitive for local services. Please scope this exception to development/local-provider templates or an explicit operator opt-in that is disabled by default in production, and enforce the same decision at save and dial time.

backend/internal/server/provider_catalog_ssrf_test.go
astaxie: **[P2] This range coverage is still incomplete on the latest head.** `disallowedProviderUpstreamCIDRs` still omits the non-global IANA assignments inside 192.0.0.0/24, including 192.0.0.0/29 (IPv4 Service Continuity / DS-Lite) and 192.0.0.170/171 (NAT64 discovery). These addresses are neither private, loopback, link-local, multicast, nor in the current explicit blocks, so a literal such as http://192.0.0.1 passes both save-time validation and checkProviderUpstreamLiteralDial even though it may route to local network infrastructure. Please add the relevant non-global subranges (without unnecessarily blocking globally reachable PCP/TURN anycast addresses .9/.10) and matching save/dial tests.

backend/internal/server/provider_upstream_ssrf.go
```diff
@@ -0,0 +1,482 @@
...
+// 100.100.100.200), the IPv4/IPv6 documentation ranges (192.0.2.0/24,
+// 198.51.100.0/24, 203.0.113.0/24, 2001:db8::/32), the benchmarking range
+// (198.18.0.0/15), the reserved range (240.0.0.0/4, which includes the
+// limited-broadcast address), the deprecated IPv6 site-local range
+// (fec0::/10, still routable inside sites that kept it), and AWS's IPv6
+// instance-metadata endpoint. NAT64 ranges are handled separately so a
+// configured RFC 6052 prefix can distinguish public from private embedded
+// IPv4 targets. IPv4-mapped IPv6 spellings need no entry: To4-based
+// predicates already classify ::ffff:10.x as private.
+var disallowedProviderUpstreamCIDRs = func() []*net.IPNet {
```
astaxie, line 171: [P2] Complete the non-global special-purpose denylist

The current classifier still accepts several IANA special-purpose destinations that are not globally reachable: `192.0.0.8/32` (IPv4 dummy), `192.88.99.2/32`, `100::/64` and `100:0:0:1::/64` (discard/dummy), `2001:2::/48` (benchmarking), `3fff::/20` (documentation), and `5f00::/16` (SRv6 SIDs). I verified that each literal passes both `validateProviderUpstreamBaseURL` and `checkProviderUpstreamLiteralDial` on this head. The SRv6 range is especially relevant to the SSRF threat model because it is forwardable and may identify internal network functions. Please classify the remaining non-global IANA assignments and add save-time/dial-time cases, while retaining the explicitly globally reachable anycast exceptions.

backend/internal/server/provider_upstream_ssrf.go
deepjerry-ai: Fixed in 9d66963. TOKENHUB_PROVIDER_UPSTREAM_NAT64_PREFIX configures the RFC 6052 prefix (supported lengths: /32, /40, /48, /56, /64, /96); the embedded IPv4 is decoded and passed through the same special/private classification, allowing public syntheses and rejecting unsafe ones. Invalid prefixes fail closed, and env/Compose/EN-zh-JA docs plus RFC 6052 format tests are included.

backend/internal/server/provider_upstream_ssrf.go
deepjerry-ai: Fixed in 9d66963. strictProviderUpstreamRedirect now requires every hop to keep the original scheme and host (including port), so cross-origin redirects, port changes, and HTTPS-to-HTTP downgrades are rejected before provider credentials or custom headers can be forwarded. Covered by TestStrictProviderUpstreamRedirectRejectsOriginChanges.

backend/internal/server/provider_upstream_ssrf.go
deepjerry-ai: Fixed in 2dbb9cb. Loopback is now disabled by default and requires the explicit TOKENHUB_PROVIDER_UPSTREAM_ALLOW_LOOPBACK=true operator opt-in. The same decision is enforced at provider/resource persistence, model discovery, and dial time; redirect hops remain strict and never permit loopback. Regression coverage verifies default rejection and opted-in save/dial behavior.

backend/internal/server/provider_upstream_ssrf.go
deepjerry-ai: Fixed in 4552028. Added the remaining non-global assignments identified here: 192.0.0.8/32, 192.88.99.2/32, 100::/64, 100:0:0:1::/64, 2001:2::/48, 3fff::/20, and 5f00::/16. TestProviderUpstreamClassifiesSpecialPurposeAssignments covers save-time and dial-time rejection and pins the globally reachable anycast exceptions (192.0.0.9/.10 and 2001:1::1/2/3) as allowed. Full local suite and GitHub CI pass.

backend/internal/server/provider_catalog_ssrf_test.go
deepjerry-ai: Addressed in 2dbb9cb and completed in 4552028. The classifier now rejects 192.0.0.0/29, 192.0.0.8/32, and 192.0.0.170/31 at both save and dial time, while the regression test explicitly keeps globally reachable 192.0.0.9 (PCP) and 192.0.0.10 (TURN) allowed.

backend/internal/server/provider_upstream_ssrf.go
```diff
@@ -0,0 +1,489 @@
...
+	cidrs := []string{
+		"0.0.0.0/8",
+		"100.64.0.0/10",
+		"192.0.0.0/29",
+		"192.0.0.8/32",
+		"192.0.0.170/31",
+		"192.0.2.0/24",
+		"192.88.99.2/32",
```
astaxie: [P2] Block the whole deprecated 6to4 relay-anycast prefix

RFC 7526 deprecates `192.88.99.0/24`, says it MUST NOT be reassigned without a future IETF Standards Action, and identifies `.1` as the classic 6to4 relay-anycast address. Blocking only `192.88.99.2/32` leaves `http://192.88.99.1/v1` accepted at both save time and dial time; I reproduced both paths on this head. A network retaining a local route for the deprecated relay space therefore remains reachable despite the stated special-purpose-address policy. Please deny the full `/24` (or explicitly justify any safe exceptions) and add save/dial regression cases for `.1` as well as `.2`.

deepjerry-ai: Addressed in 320e4b2. The classifier now denies the full deprecated `192.88.99.0/24`, and `TestProviderUpstreamRejectsCuratedNonProviderRanges` covers both `192.88.99.1` and `.2` through the save-time and dial-time decisions.

I also tightened the policy wording in code and the PR description: this is a curated trust-boundary classifier for loopback, metadata, internal/private, and other unsafe/non-provider ranges, not a claim to mirror every IANA special-purpose allocation. Deployment-specific reachability remains an egress-policy concern. Full `go test ./...`, `go vet ./...`, repository gates, Compose rendering, and `git diff --check` pass.

backend/internal/server/provider_upstream_redirect.go
```diff
@@ -0,0 +1,28 @@
...
+		return fmt.Errorf("stopped after 10 redirects")
+	}
+	if err := validateProviderUpstreamBaseURL(req.URL, nil, false); err != nil {
+		return err
+	}
+	if len(via) == 0 || via[0] == nil || via[0].URL == nil {
+		return fmt.Errorf("redirect has no original request")
+	}
+	original := via[0].URL
+	if !strings.EqualFold(req.URL.Scheme, original.Scheme) || !strings.EqualFold(req.URL.Host, original.Host) {
```
astaxie, line 50: [P2] Permit same-host HTTP-to-HTTPS upgrades

This scheme equality check also rejects the normal safe upgrade from `http://api.example.com/...` to `https://api.example.com/...`. I reproduced that exact hop against `strictProviderUpstreamRedirect`; it returns `provider upstream redirect changed origin`. Because the PR still accepts `http` base URLs and installs this hook on discovery plus every inference/Codex client, a public Provider that canonicalizes HTTP to HTTPS becomes unusable despite the compatibility claim that legitimate public endpoints are unaffected. Keep rejecting cross-host redirects and HTTPS-to-HTTP downgrades, but allow an HTTP-to-HTTPS transition for the same hostname (and expected default ports), with an upgrade regression test.

backend/internal/server/provider_upstream_ssrf.go
```diff
@@ -0,0 +1,493 @@
...
+// Redirect re-validation passes nil for allowedPrivate and false for
+// allowLocalhost: a public URL must not be able to bounce requests into the
+// internal network or onto a loopback service, even when the operator
+// allowlists private ranges or runs a local provider.
+func validateProviderUpstreamBaseURL(endpoint *url.URL, allowedPrivate []*net.IPNet, allowLocalhost bool) error {
+	if endpoint == nil {
+		return NewHTTPError(http.StatusBadRequest, "provider_base_url_invalid", "Base URL is invalid")
+	}
+	scheme := strings.ToLower(strings.TrimSpace(endpoint.Scheme))
+	if scheme != "https" && scheme != "http" {
```
astaxie, line 42: [P1] Require TLS before attaching Provider credentials

The earlier suggestion to permit same-host HTTP-to-HTTPS upgrades was too permissive: this validator still accepts arbitrary public `http://` base URLs, and the discovery/inference clients attach Bearer tokens, `x-api-key`/`api-key`, custom authentication headers, or Gemini’s key query parameter to the initial request before any redirect hook runs. An on-path attacker can therefore read or alter the credentials and request on the first plaintext hop; a later HTTPS redirect cannot restore confidentiality or integrity. Require `https` for public upstreams at persistence and connection time. If plaintext is needed for local deployments, scope it to explicitly opted-in loopback or allowlisted private literal endpoints, and replace the public-HTTP acceptance case with create/update and client regressions.

deepjerry-ai: Fixed in bb3374c. Public Provider upstream URLs now require HTTPS at persistence time and again immediately before RoundTrip can send credentials. Plain HTTP remains available only for explicitly opted-in loopback endpoints or literal private IPs covered by TOKENHUB_PROVIDER_UPSTREAM_ALLOWED_CIDRS. The request-time policy wrapper is installed on non-streaming, streaming, Codex subscription, and catalog clients ahead of the existing guarded dial transport, so older/directly persisted records cannot bypass the rule. Regression coverage now includes provider create/update rejection, resource persistence, client wiring, and a credential-bearing public HTTP request whose underlying transport must not be called. The English, Chinese, and Japanese deployment docs describe the exception boundary. Full go test ./..., go vet ./..., repository tests/gates, and git diff --check pass.

backend/internal/server/provider_upstream_redirect.go
deepjerry-ai: Superseded by the later TLS-before-credentials finding and addressed in bb3374c. I did not permit public HTTP-to-HTTPS redirects: public HTTP base URLs are now rejected before the first credential-bearing request is sent, because a redirect cannot recover confidentiality or integrity after that plaintext hop. Public endpoints must start with HTTPS; HTTP remains limited to explicitly opted-in loopback or allowlisted private literal endpoints, and redirects continue to require the same scheme and origin.

astaxie (approved):
Re-reviewed the complete current diff at bb3374c. The latest commit requires TLS before any credential-bearing public upstream request while preserving explicit loopback/private-literal exceptions. All prior SSRF threads are verified and resolved. Full backend tests, go vet, repository tests/gates, synchronized translation/environment/source-line checks, Compose rendering, and diff check pass locally; required CI is green.

PR #172, opened 2026-08-09
test(server): cover method routing contracts

bailu-ZZ:
## Summary

Add characterization tests for the HTTP method-routing behavior that must remain stable before adopting method-aware `http.ServeMux` patterns.

## Related Issue

Related #145 

## Changes

- Cover generic JSON method errors and request IDs for health, admin, and OpenAI-compatible endpoints.
- Lock down the existing authentication-before-method-check behavior for protected admin endpoints.
- Cover Anthropic-specific method errors, GET/HEAD behavior, and CORS preflight handling.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor or maintenance
- [ ] Documentation

## Verification

- `git diff --check`

None. This pull request adds tests only and does not change production behavior.

## Checklist

### DIFF
--- a/backend/internal/server/method_routing_contract_test.go
+++ b/backend/internal/server/method_routing_contract_test.go
@@ -17,20 +17,22 @@ func TestMethodRoutingContracts(t *testing.T) {
 	}
 
 	tests := []struct {
-		name   string
-		method string
-		path   string
-		want   string
+		name      string
+		method    string
+		path      string
+		wantCode  string
+		wantAllow string
 	}{
-		{name: "health wrong method", method: http.MethodPost, path: "/livez", want: "method_not_allowed"},
-		{name: "admin login wrong method", method: http.MethodGet, path: "/api/admin/auth/login", want: "method_not_allowed"},
-		{name: "gateway wrong method", method: http.MethodGet, path: "/v1/chat/completions", want: "method_not_allowed"},
-		{name: "models rejects head", method: http.MethodHead, path: "/v1/models", want: "method_not_allowed"},
+		{name: "health wrong method", method: http.MethodPost, path: "/livez", wantCode: "method_not_allowed", wantAllow: ""},
+		{name: "admin login wrong method", method: http.MethodGet, path: "/api/admin/auth/login", wantCode: "method_not_allowed", wantAllow: ""},
+		{name: "gateway wrong method", method: http.MethodGet, path: "/v1/chat/completions", wantCode: "method_not_allowed", wantAllow: ""},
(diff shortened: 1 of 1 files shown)


backend/internal/server/method_routing_contract_test.go
```diff
@@ -0,0 +1,129 @@
...
+	tests := []struct {
+		name   string
+		method string
+		path   string
+		want   string
+	}{
+		{name: "health wrong method", method: http.MethodPost, path: "/livez", want: "method_not_allowed"},
+		{name: "admin login wrong method", method: http.MethodGet, path: "/api/admin/auth/login", want: "method_not_allowed"},
+		{name: "gateway wrong method", method: http.MethodGet, path: "/v1/chat/completions", want: "method_not_allowed"},
+		{name: "models rejects head", method: http.MethodHead, path: "/v1/models", want: "method_not_allowed"},
```
astaxie: **[P2] Exercise HEAD on a route that phase 1 will actually migrate**

The approved plan defers public /v1 endpoints, so this /v1/models case will not protect the first migration. If /livez (or another fixed GET admin/internal endpoint) is registered with a method-aware GET pattern and its handler guard is removed, ServeMux will accept HEAD and this suite still passes. Please put the HEAD contract on at least one phase-1 GET route. For that HEAD assertion, avoid decoding ResponseRecorder.Body: httptest.ResponseRecorder records handler writes for HEAD while a real net/http.Server suppresses them.

backend/internal/server/method_routing_contract_test.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			response := methodRoutingRequest(app, test.method, test.path, "")
+			assertJSONError(t, response, http.StatusMethodNotAllowed, test.want)
```
astaxie: **[P2] Assert the Allow header contract**

Issue #145 explicitly requires regression coverage for Allow, but every wrong-method case funnels through assertJSONError, which never checks it. This leaves one of the main observable changes from method-aware ServeMux unpinned, so the migration could expose the wrong allowed-method set (or omit it in a custom JSON 405 fallback) without a test failure. Please add endpoint-specific expected Allow values and assert them; if this PR is strictly characterizing the pre-migration baseline, make the current absence explicit and update that expectation with the migration.

astaxie (approved):
Current head f138ee9 reviewed. The test-only method-routing characterization matches the stated scope; GitHub CI is green and local backend tests, vet, repository gates, translation checks, and diff checks pass.

bailu-ZZ:
Good catch, I missed this in the baseline tests.

  I’ve added explicit checks for the missing Allow header and tightened the error response assertions. The routing tests are passing now.

PR #174, opened 2026-08-09
feat: add project-scoped content security guardrails

deepjerry-ai:
## Summary

Add project-scoped content security Guardrails that inspect supported model requests before Provider routing, apply `audit`, `mask`, or `block`, and give administrators actionable policy diagnostics in the API and Model Playground.

## Related Issue

N/A.

## Changes

- Add persistent Guardrail policies, detection items, and all-project/project bindings with admin CRUD and policy-test APIs.
- Add deterministic keyword/regex and sensitive-data detectors plus an optional OpenAI-compatible Qwen3Guard detector.
- Enforce policies for `/v1/chat/completions`, `/v1/responses`, `/v1/messages`, and Model Playground requests before Provider routing.
- Return HTTP 403 `guardrail_blocked` errors with safe `categories`, `reason_codes`, `policy_matches`, and `request_id` details; matched raw text is not included in the error details.
- Add the Content Security administration UI, policy samples, localized copy, and actionable Playground block feedback.
- Validate sensitive-data structure, including Chinese identity-card checksum/date, Luhn bank-card values, and real calendar birth dates.
- Read each policy and its preloaded detection items/bindings from one database snapshot so a concurrent update cannot produce a mixed policy version.
- Remove the independent 64 KiB Guardrails inspection ceiling. Long-context requests remain bounded by the configured request-body limit; model-detector chunking is a follow-up optimization.
- Document behavior, architecture, configuration, data boundaries, rollout, and rollback in English, Simplified Chinese, and Japanese.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Refactor or maintenance
- [x] Documentation

## Verification

- [x] `go test ./...`
- [x] `go vet ./...`
- [x] `npm run typecheck`
- [x] `npm run build`
- [x] `node --test tools/*.test.mjs` — 116 passed
- [x] `node tools/check-ui-translations.mjs`
- [x] `git diff --check`
- [x] Default, local PostgreSQL, and remote PostgreSQL Compose configurations parse successfully; the remote configuration was checked with non-secret placeholder values for its required deployment variables.

- **API compatibility:** Existing successful responses are unchanged. A matching block policy adds an HTTP 403 `guardrail_blocked` outcome on the supported model endpoints. Guardrails no longer add a dedicated 413 outcome; the existing configured request-body limit remains in effect and may independently reject oversized HTTP bodies.
- **Inspection boundary:** This version inspects ordinary user-visible request text. Structured tool arguments, arbitrary JSON values, code-specific parsing, and Provider responses are not inspected.
- **Security and data handling:** Deterministic detectors run inside TokenHub. Enabling `TOKENHUB_GUARDRAIL_MODEL_URL` sends inspected user-visible request text to the configured Qwen3Guard service, which must be deployed within an approved trust boundary with appropriate transport, logging, and retention controls. Leaving the URL empty prevents model calls and uses each policy's configured unavailable behavior. Structured block details and Guardrail findings do not include matched raw text. Existing request-payload audit storage remains governed by its configured retention and access controls.
- **Database migration:** Startup AutoMigrate creates tables for `guardrails.Policy`, `guardrails.DetectionItem`, and `guardrails.Binding`, including their foreign keys and uniqueness constraints. No destructive migration or backfill is required.
- **Configuration:** Adds `TOKENHUB_GUARDRAIL_MODEL_URL`, `TOKENHUB_GUARDRAIL_MODEL_API_KEY`, `TOKENHUB_GUARDRAIL_MODEL_NAME`, and `TOKENHUB_GUARDRAIL_MODEL_TIMEOUT_SECONDS`; backend and deployment examples, Compose wiring, startup forwarding, and deployment documentation are synchronized.
- **Rollout:** Deploy the application so AutoMigrate completes, initially leave the Qwen endpoint empty, create policies disabled or validate them with Test Policy, then enable project-scoped policies gradually before broader rollout.
- **Rollback:** Disable active policies and unset the Qwen endpoint first, then deploy the previous application version if needed. The additive Guardrail tables may remain unused; removing them is an optional, separate destructive database operation and is not required for application rollback.

## Checklist

- [x] `data/model-catalog.yaml` remains tracked and catalog changes were reviewed where applicable; this PR does not change the catalog.


backend/internal/server/gateway_http.go
```diff
@@ -35,21 +35,30 @@ func (s *Server) handleChatCompletions(w http.ResponseWriter, r *http.Request) {
 		writeError(w, r, NewHTTPError(400, "missing_model", "model is required"))
 		return
 	}
+	guardrailStarted := time.Now().UTC()
+	decision, err := s.evaluateOutboundGuardrails(r.Context(), project.ID, chatGuardrailTargets(&req))
```
astaxie: [P1] Enforce API-key admission before invoking the guard model

This evaluation happens before `startRoutedCallWithAudit`, while `StartCall` is where model access, RPM/TPM, daily/monthly quotas, and the API-key concurrency lease are enforced. With an applicable model detector, any authenticated key can therefore issue unlimited Qwen3Guard calls that do not consume its configured limits (and can do so while naming a model the key is not allowed to use). A leaked or low-limit key can exhaust or bill the dedicated guard service without ever reaching admission. Please validate/admit the call, or apply an equivalent rate/concurrency budget, before making the external detector request; blocked calls still need the admitted call to be finished so reservations/leases are released.

frontend/features/admin/views/security-policies.tsx
```diff
@@ -0,0 +1,496 @@
...
+                  <button aria-label={tx("删除策略")} className="icon-button subtle danger" onClick={() => setDeletePolicy(policy)} title={tx("删除策略")} type="button"><Trash2 size={15} /></button>
+                </span>
+              </div>
+            ))}
+          </div>
+        )}
+      </div>
+
+      {draft ? <GuardrailPolicyEditor api={api} data={data} draft={draft} saving={saving} onChange={setDraft} onClose={() => !saving && setDraft(null)} onSave={() => void savePolicy()} /> : null}
+      {deletePolicy ? <ConfirmDialog title="删除内容安全策略" message={tx("删除后，这条策略将立即停止生效。此操作无法撤销。")} confirmLabel="删除" loading={saving} onCancel={() => setDeletePolicy(null)} onConfirm={() => void removePolicy()} /> : null}
```
astaxie: [P2] Translate the delete dialog title and action

`title` and `confirmLabel` are literal Simplified Chinese, so this confirmation remains Chinese in the English and Japanese consoles even though both catalog entries already exist. Wrap both values with `tx(...)`.

frontend/features/admin/domain/formatting.tsx
```diff
@@ -283,9 +283,39 @@ export async function readAPIError(resp: Response) {
...
   if (code === "provider_resource_concurrency_exceeded") return "Provider 资源并发已满，请稍后再试。";
   if (code === "provider_resource_cooling_down") return "Provider 资源处于冷却中，请检查资源健康状态。";
+  if (code === "guardrail_blocked") {
+    const categories: string[] = Array.isArray(payload?.error?.details?.categories) ? payload.error.details.categories.filter((category: unknown): category is string => typeof category === "string") : [];
+    const labels: string[] = Array.from(new Set<string>(categories.map(guardrailCategoryLabel)));
+    const rawMatches: unknown[] = Array.isArray(payload?.error?.details?.policy_matches) ? payload.error.details.policy_matches : [];
+    const policyLabels = Array.from(new Set(rawMatches.map(guardrailPolicyMatchLabel).filter((label): label is string => Boolean(label)))).slice(0, 3);
+    const requestID = typeof payload?.request_id === "string" ? payload.request_id.trim() : "";
+    const details = [
+      labels.length > 0 ? `${tx("原因")}：${labels.map((label) => tx(label)).join("、")}` : "",
```
astaxie: [P2] Localize the complete diagnostic instead of joining fragments

This builds every locale with Chinese full-width separators (`：`, `、`, `；`) and fixes the clause order in code, so the English/Japanese Playground message is only partially localized. It also violates the repository rule against concatenating translated sentence fragments. Please put the complete diagnostic shapes, including separators and placeholder order, behind locale-aware templates/helpers.

frontend/features/admin/views/security-policies.tsx
```diff
@@ -0,0 +1,496 @@
...
+        {error ? <div className="guardrail-inline-message error"><AlertCircle size={16} /><span>{error}</span><button aria-label={tx("关闭")} className="icon-button subtle" onClick={() => setError("")} title={tx("关闭")} type="button"><X size={14} /></button></div> : null}
+        {notice ? <div className="guardrail-inline-message success"><ShieldCheck size={16} /><span>{notice}</span><button aria-label={tx("关闭")} className="icon-button subtle" onClick={() => setNotice("")} title={tx("关闭")} type="button"><X size={14} /></button></div> : null}
+
+        <div className="table-toolbar guardrail-toolbar">
+          <div className="search-box">
+            <Search size={16} />
+            <input onChange={(event) => setQuery(event.target.value)} placeholder={tx("搜索策略名称或说明")} type="search" value={query} />
+          </div>
+          <div className="table-toolbar-actions">
+            <span className="table-result-count">{filteredPolicies.length} {tx("条策略")}</span>
```
astaxie: [P2] Format the new counters with the selected application locale

This raw count plus translated-unit pattern is repeated for detection-item counts, item numbers, duration, and `scopeSummary` below. It bypasses `Intl`/`languageLocale()` and composes UI text from translated fragments, contrary to the repository i18n contract. Please use `countWithUnit` or complete locale-aware formatters for these new values, and an `Intl.NumberFormat(languageLocale())`-based duration formatter.

frontend/features/admin/views/security-policies.tsx
```diff
@@ -0,0 +1,496 @@
...
+            {filteredPolicies.map((policy) => (
+              <div className="guardrail-policy-row" key={policy.id}>
+                <button className="guardrail-policy-main" onClick={() => setDraft(editPolicyDraft(policy))} type="button">
+                  <span><strong>{policy.name}</strong><small>{policy.description || tx("暂无说明")}</small></span>
+                  <StatusPill status={policy.status} />
+                </button>
+                <span>{policy.detection_items.length} {tx("项")}</span>
+                <span>{scopeSummary(policy, data)}</span>
+                <span>{tx("请求发送前")}</span>
+                <span>{formatTime(policy.updated_at ?? "")}</span>
```
astaxie, line 227: [P2] Render the timestamp in the selected application language

`formatTime` calls parameterless `toLocaleString()`, so this new column follows the browser/OS locale rather than the language selected in TokenHub. For example, selecting Japanese in an English browser still renders an English timestamp. Please format this value with `Intl.DateTimeFormat(languageLocale(), ...)`, either here or through a locale-aware shared helper.

backend/internal/guardrails/engine.go
```diff
@@ -0,0 +1,520 @@
...
+
+func sensitiveDataDefinitions(dataType string) []sensitiveDataDefinition {
+	switch dataType {
+	case "credential":
+		return sensitiveDefinitions(
+			`\b(?:AKIA|ASIA)[A-Z0-9]{16}\b`,
+			`(?i)\b(?:sk-[a-z0-9_-]{16,}|gh[pousr]_[a-z0-9]{20,}|glpat-[a-z0-9_-]{20,}|xox[baprs]-[a-z0-9-]{20,})\b`,
+			`\bAIza[0-9A-Za-z_-]{35}\b`,
+			`\beyJ[A-Za-z0-9_-]{10,}\.[A-Za-z0-9_-]{10,}\.[A-Za-z0-9_-]{10,}\b`,
+			`-----BEGIN [A-Z ]*PRIVATE KEY-----`,
```
astaxie: [P1] Mask the private-key material, not only the PEM header

For a credential detector configured with the `mask` action, this match span covers only `-----BEGIN ... PRIVATE KEY-----`. The replacement therefore leaves the base64 key body and the `END` marker untouched and forwards the actual private key to the provider. Please match and redact the complete PEM block, and add a regression test that asserts no key-body bytes survive masking.

backend/internal/server/gateway_http.go
```diff
@@ -518,27 +540,31 @@ func (s *Server) handleAdminPlaygroundChat(w http.ResponseWriter, r *http.Reques
...
+		httpErr := AsHTTPError(guardrailErr)
+		s.finishRoutedCall(r, GatewayCallCompletion{
+			Kind: CompletionKindPlayground, Call: call, StatusCode: httpErr.Status,
+			ErrorCode: httpErr.Code, ErrorMessage: httpErr.Message, RequestPayload: requestAuditPayload,
+			ResponsePayload: auditErrorPayload(guardrailErr, requestID),
+		})
+		s.recordAdminAudit(r, user, "chat_failed", "playground", req.Model, "", map[string]any{
+			"model": req.Model, "attempts": []PlaygroundRouteAttempt{}, "error": httpErr.Code,
+		})
+		writeError(w, r, guardrailErr)
```
astaxie, line 589: [P1] Return the request ID used for the blocked Playground audit record

The blocked call is persisted by `finishRoutedCall` under `call.RequestID` (the `pg_*` value), but this path calls `writeError` without first putting that ID in the response header. `errorResponseHeaders` then generates a different `req_*` value, so the response body and Playground UI show an ID that cannot locate the stored request log. Set `x-request-id` to `requestID` before writing the error; the streaming Playground path has the same issue.

backend/internal/server/gateway_http.go
deepjerry-ai: Fixed in de937d7. API-key admission now runs through StartCall before guard-model evaluation for Chat Completions, Responses, and Anthropic Messages. Guardrail failures finish the admitted call so token reservations, quotas, and concurrency leases are released. Added regression coverage proving a disallowed model never invokes the guard service and uninspected input is not persisted.

frontend/features/admin/views/security-policies.tsx
deepjerry-ai: Fixed in de937d7. The delete dialog title and confirm action now both go through tx(...), so the existing English and Japanese catalog entries are used.

frontend/features/admin/domain/formatting.tsx
deepjerry-ai: Fixed in de937d7. The blocked-request diagnostic now uses a locale-aware guardrailBlockedDiagnostic helper with complete Chinese, English, and Japanese message shapes, including locale-appropriate punctuation, separators, and clause order.

frontend/features/admin/views/security-policies.tsx
deepjerry-ai: Fixed in de937d7. Policy, detection-item, and project counts now use locale-aware formatters; unnamed detection items use a complete locale-specific label; duration uses an Intl.NumberFormat-backed milliseconds helper.

frontend/features/admin/views/security-policies.tsx
deepjerry-ai: Fixed in de937d7. formatTime now uses Intl.DateTimeFormat(languageLocale(), { dateStyle: "medium", timeStyle: "medium" }) and safely handles invalid dates.

backend/internal/guardrails/engine.go
deepjerry-ai: Fixed in de937d7. The credential detector now matches and masks the complete PEM private-key block through the END marker. Added a regression test asserting neither key-body bytes nor the end marker survive masking.

backend/internal/server/gateway_http.go
deepjerry-ai: Fixed in de937d7. Both synchronous and streaming Playground handlers now set x-request-id from the Playground call context before guardrail evaluation. Regression coverage asserts the response header, response body, and persisted audit record all use the same ID.

backend/internal/server/playground_stream.go
```diff
@@ -319,15 +336,18 @@ func validatePlaygroundRequest(req *ChatCompletionRequest) error {
 	return nil
 }
 
-func (s *Server) preparePlaygroundRoutedCall(ctx context.Context, user AdminUser, req ChatCompletionRequest, startedAt time.Time) (RoutedCall, error) {
-	call := CallContext{
+func (s *Server) newPlaygroundCallContext(user AdminUser, model string, startedAt time.Time) CallContext {
+	return CallContext{
 		RequestID:  NewID("pg"),
 		Project:    Project{ID: "admin_playground", Name: "Admin Playground", Status: StatusActive},
```
astaxie, line 392: [P1] Project-scoped guardrails never apply in Model Playground

Both Playground handlers evaluate against this synthetic `admin_playground` project, while `policyApplies` requires an exact project ID and policy CRUD rejects bindings to nonexistent projects. As a result, a blocking policy bound to a real project blocks that project API key but the same prompt in Model Playground skips the policy and reaches the provider; only `all_projects` policies can produce the advertised Playground diagnostics. I reproduced this on the current head with `prj_demo`. Please let Playground supply the project context, or explicitly limit and document Playground enforcement to all-project policies.

backend/internal/guardrails/engine.go
```diff
@@ -0,0 +1,520 @@
...
+	defer func() { decision.DurationMS = time.Since(started).Milliseconds() }()
+
+	if request.Checkpoint == "" {
+		request.Checkpoint = CheckpointBeforeProvider
+	}
+	if request.Protocol == "" {
+		request.Protocol = ProtocolAll
+	}
+	policies := applicablePolicies(request)
+	if len(policies) == 0 {
```
astaxie, line 97: [P2] Align the 64 KiB limit with the documented contract

This early return means oversized text is accepted whenever no active policy applies, including when every policy is scoped to another project. `TestEngineDoesNotLimitInputWithoutApplicablePolicies` explicitly locks in that behavior, but the PR body and all three administrator guides state that inspectable text above 64 KiB is rejected with HTTP 413 and `guardrail_input_too_large`. Please move the size check before this return if the boundary is intended to be unconditional, or narrow the public documentation and PR contract to say it only applies when at least one policy is evaluated.

astaxie:
The PR body does not preserve the required `.github/pull_request_template.md` sections. It omits `Related Issue`, `Changes`, `Type of Change`, `Compatibility, Security, and Operations`, and `Checklist`, although both `AGENTS.md` and `CONTRIBUTING.md` require every section to be completed and non-applicable checks explained. This change adds AutoMigrate-backed tables, four environment variables, and new `/v1` 403/413 outcomes, so please restore the template and explicitly document compatibility, security/data handling, database migration, configuration/deployment, rollout, and rollback impact.

backend/internal/server/playground_stream.go
deepjerry-ai: Fixed in 3c3e0a7. Model Playground now requires an explicit project_id selected from the current user's active accessible projects. Both synchronous and streaming handlers validate project existence, status, and access, then evaluate guardrails against that real project ID while retaining the synthetic Playground call context for billing/audit isolation. The internal project_id is removed before provider forwarding. Regression coverage verifies project-scoped blocking on both handlers, missing and unauthorized project failures, and that the field never reaches the provider.

backend/internal/guardrails/engine.go
deepjerry-ai: Fixed in 3c3e0a7. The 64 KiB evaluation-size check now runs before applicable-policy filtering, so oversized inspectable text is rejected even with no policies or only policies scoped to another project. Engine tests cover both cases, and HTTP regression coverage verifies 413 guardrail_input_too_large across Chat Completions, Responses, Anthropic Messages, and both Playground handlers.

deepjerry-ai: Follow-up correction in 8a434ff. While aligning the implementation with the documented 64 KiB contract, we found a more fundamental compatibility issue: making that ceiling unconditional causes TokenHub to reject legitimate long-context requests before provider routing, including workloads intended for 1M-token models. The 64 KiB inspection ceiling is also far below the gateway's existing 32 MiB default request-body allowance.

The revised fix removes the independent Guardrails evaluation-size limit instead of moving it around. The configured request-body limit remains the transport/capacity boundary, deterministic detectors continue to inspect large inputs, and large requests with no applicable policy are allowed normally. Regression coverage now exercises 1 MiB inputs both with an applicable deterministic policy and with no applicable policy.

Model-detector chunking remains a follow-up improvement; until then, detector failures continue to follow each policy's existing `on_unavailable` behavior. The separate 64 KiB audit-payload truncation remains in place because it only bounds persisted diagnostics and does not truncate or reject the request sent to the provider.

deepjerry-ai:
Addressed in `7c54ecc` and the PR body has been restored to the required template.

The updated **Compatibility, Security, and Operations** section now explicitly covers:

- the new HTTP 403 `guardrail_blocked` API outcome and safe structured diagnostics;
- the supported inspection boundary and the fact that enabling Qwen3Guard sends inspected user-visible request text to the configured service;
- additive AutoMigrate tables, constraints, and rollback behavior;
- all four environment variables, deployment sequencing, staged rollout, and rollback.

The 413 note changed after deeper review: the original Guardrails-specific 64 KiB ceiling would have rejected legitimate long-context traffic before Provider routing, including requests intended for large/1M-context models. That was the more important issue, so the independent Guardrails limit and its dedicated 413 outcome were removed. Requests remain bounded by TokenHub's configured HTTP request-body limit, while model-detector chunking remains a separate follow-up optimization.

This review also fixed two additional correctness/security issues:

- Guardrail policies and their preloaded detection items/bindings are now read from one database snapshot, preventing a concurrent update from producing a mixed policy version.
- Birth-date detection now validates real calendar dates, and Qwen prompt egress is documented in the English, Simplified Chinese, and Japanese administrator/deployment guides.

Local verification passed: backend tests and vet, frontend typecheck/build, 116 repository tool tests, translation/environment/source gates, `git diff --check`, and all three Compose configurations.

backend/internal/server/playground_stream.go
```diff
@@ -12,6 +12,28 @@ import (
...
+	if err := json.Unmarshal(data, &request); err != nil {
+		return err
+	}
+	var context struct {
+		ProjectID string `json:"project_id"`
+	}
+	if err := json.Unmarshal(data, &context); err != nil {
+		return err
+	}
+	delete(request.raw, "project_id")
```
astaxie: [P2] Strip every spelling accepted for the gateway-only project field

`encoding/json` matches tagged fields case-insensitively, so `PROJECT_ID` (or `Project_ID`) populates `ProjectID`, but this removes only the exact lowercase raw key. `ChatCompletionRequest.MarshalJSON` preserves unknown raw fields, so the alternate spelling is then forwarded to the Provider by both Playground paths. I reproduced this on this head: unmarshalling a payload containing `"PROJECT_ID":"prj_demo"` selects that project, and marshaling the embedded request still emits `PROJECT_ID`. Reject non-canonical/duplicate spellings or remove the matched key case-insensitively, and cover both buffered and streaming handlers.

backend/internal/guardrails/engine.go
```diff
@@ -0,0 +1,515 @@
...
+	}
+
+	modelItems := make([]policyItem, 0)
+	for _, policy := range policies {
+		for _, item := range policy.DetectionItems {
+			if item.DetectorType == DetectorModel {
+				modelItems = append(modelItems, policyItem{policy: policy, item: item})
+				continue
+			}
+			decision.Findings = append(decision.Findings, deterministicFindings(policy, item, request.Fragments)...)
```
astaxie: [P1] Bound deterministic matching work before accepting long contexts

This loop permits one policy to run 32 detection items times 64 expressions over the entire request, and the number of applicable policies is not bounded. With a valid no-match regex set (`a{N,512}b`), a 64 KiB fragment on an Apple M5 Pro was still evaluating after 35 seconds; a 1 MiB fragment exceeded 84 seconds before I interrupted it. The default endpoint limit is 32 MiB, and these scans do not observe context cancellation, so one admitted request can pin a core long after the client disconnects. Add an aggregate work budget or combined/precompiled matcher strategy with cancellation checks before treating the request-body ceiling as the only inspection bound.

PR #186, opened 2026-08-10
test(performance): add reproducible gateway benchmarks

### LINKED ISSUE #175: test(performance): add gateway performance budgets
## What to build

Build a reproducible TokenHub gateway benchmark suite that requires no real provider credentials. Use deterministic in-process upstreams to cover representative non-streaming, streaming, single-route, failover, SQLite, PostgreSQL, audit, and tracing scenarios. Report throughput, P50/P95/P99 latency, gateway overhead, and allocations. Store baselines with hardware and configuration metadata and define tolerant performance budgets for critical hot paths so regressions are visible before release.

This work is inspired by Bifrost's public benchmarking approach, but it must measure TokenHub's own governance, routing, and audit model rather than copy absolute RPS claims.

Reference: https://docs.getbifrost.ai/benchmarking/getting-started

## Acceptance criteria

- [ ] benchmarks use deterministic in-process fake upstreams and no external providers.
- [ ] cover Chat Completions, Responses, streaming, single-candidate routing, and failover.
- [ ] measure the impact of SQLite/PostgreSQL, payload auditing, metrics, and tracing independently.
- [ ] report throughput, percentiles, gateway overhead, allocations, memory, hardware, Go version, and configuration.
- [ ] provide regression checks with documented tolerances that avoid flaky failures on shared CI.
- [ ] document local runs, comparisons, and baseline updates in English, Simplified Chinese, and Japanese.
- [ ] backend tests, vet, repository gates, and git diff --check pass.

## Blocked by

None - can start immediately

astaxie:
## Summary

Add a reproducible performance benchmark suite for TokenHub that uses deterministic local upstreams instead of real provider credentials. The suite supports fair black-box gateway comparisons, in-process allocation benchmarks, scenario-compatible regression budgets, and machine-described baselines.

## Related Issue

Closes #175

## Changes

- Add `tokenhub-benchmark` commands for a deterministic mock upstream, a self-contained TokenHub benchmark gateway, fixed-rate/fixed-concurrency load, JSON/Markdown reports, and budget checks.
- Add in-process Chat, Responses, streaming, failover, SQLite, opt-in PostgreSQL, audit, metrics, and tracing benchmarks with allocation reporting.
- Add alternating TokenHub/Bifrost comparison scripts, four tracked local baselines, tolerant budget checks, and English, Simplified Chinese, and Japanese documentation.
- Record offered, completed, and generator-dropped requests separately; measure throughput over actual wall-clock drain time and separate streaming TTFT overhead from total-response overhead.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor or maintenance
- [x] Documentation

## Verification

- `go test ./...`
- `go vet ./...`
- `go test -race ./internal/perfbench ./cmd/tokenhub-benchmark -count=1`
- `TOKENHUB_BENCHMARK_BENCHTIME=50ms TOKENHUB_BENCHMARK_COUNT=1 ./benchmarks/run-internal.sh`
- `./benchmarks/check-suite.sh output/benchmarks/current`
- `bash -n benchmarks/*.sh`
- `git diff --check`
- PostgreSQL benchmarks were not run because `TOKENHUB_BENCHMARK_POSTGRES_URL` was not configured; the opt-in cases use rollback-isolated transactions.

No production API, persistence schema, authentication, deployment, or runtime behavior changes. Benchmark credentials are supplied through environment variables and are removed from result artifacts. Tracked baselines contain only scenario, runtime, hardware, and aggregate performance metadata.

## Checklist


benchmarks/run-internal.sh
```diff
@@ -0,0 +1,16 @@
...
+RESULT_PATH="${TOKENHUB_BENCHMARK_GO_RESULT:-$ROOT_DIR/output/benchmarks/go-benchmarks.txt}"
+
+mkdir -p "$(dirname "$RESULT_PATH")"
+cd "$ROOT_DIR/backend"
+go test ./internal/server \
+  -run '^$' \
+  -bench '^BenchmarkGateway' \
+  -benchmem \
+  -benchtime "$BENCHTIME" \
+  -count "$COUNT" | tee "$RESULT_PATH"
```
astaxie, line 24: [P1] Add regression budgets for the internal benchmark matrix

This script only saves raw Go benchmark output. The only budget checker in this PR consumes the four black-box JSON baselines, so regressions in SQLite/PostgreSQL, audit payloads, metrics, tracing, `B/op`, and `allocs/op` never affect an exit status. Issue #175 explicitly asks for tolerant budgets on critical hot paths so these regressions are visible before release. Please turn these results into comparable metrics with thresholds (or add an equivalent `benchstat`-based gate) rather than only teeing them to a file.

backend/internal/server/gateway_performance_test.go
```diff
@@ -0,0 +1,198 @@
...
+	})
+}
+
+func BenchmarkGatewayGovernanceCosts(b *testing.B) {
+	cases := []struct {
+		name    string
+		options gatewayBenchmarkOptions
+	}{
+		{name: "AuditSmall"},
+		{name: "AuditLargePayload", options: gatewayBenchmarkOptions{requestPadding: 32 * 1024}},
```
astaxie: [P1] Isolate the payload-auditing cost

Increasing `requestPadding` changes JSON decoding, request construction, upstream transfer/decoding, and persistence at the same time as the audit payload. The delta between `AuditSmall` and this case therefore cannot be attributed independently to payload auditing, which is an acceptance criterion in #175. Please add an audit-off/no-op control using the same 32 KiB request, or benchmark the audit capture/persistence path directly with small and large payloads.

backend/internal/perfbench/budget.go
```diff
@@ -0,0 +1,96 @@
...
+		a    any
+		b    any
+	}{
+		{"protocol", a.Protocol, b.Protocol}, {"stream", a.Stream, b.Stream}, {"mode", a.Mode, b.Mode},
+		{"rate", a.Rate, b.Rate}, {"concurrency", a.Concurrency, b.Concurrency}, {"model", a.Model, b.Model},
+		{"request bytes", a.RequestBytes, b.RequestBytes}, {"expected upstream latency", a.ExpectedUpstreamLatency, b.ExpectedUpstreamLatency},
+		{"response bytes", a.ResponseBytes, b.ResponseBytes}, {"stream chunks", a.StreamChunks, b.StreamChunks}, {"chunk interval", a.ChunkInterval, b.ChunkInterval},
+		{"expected upstream TTFT", a.ExpectedUpstreamTTFT, b.ExpectedUpstreamTTFT}, {"deployment profile", a.DeploymentProfile, b.DeploymentProfile},
+		{"duration", a.Duration, b.Duration}, {"warmup", a.Warmup, b.Warmup}, {"timeout", a.Timeout, b.Timeout}, {"max in flight", a.MaxInFlight, b.MaxInFlight},
+	}
```
astaxie, line 67: [P2] Reject incompatible runtime profiles

The compatibility fingerprint ignores all recorded metadata. A result produced on Linux/amd64 with a different CPU or Go version is therefore accepted against the tracked Apple M5/darwin baseline, and the resulting throughput/latency difference is reported as a code regression. Please include the relevant runtime/hardware fields (or an explicit runner-profile identifier) in compatibility checks so the budget fails closed instead of comparing unlike machines.

benchmarks/baselines/tokenhub-local-smoke.json
```diff
@@ -0,0 +1,76 @@
...
+    "warmup": "500ms",
+    "timeout": "30s",
+    "expected_upstream_latency": "5ms",
+    "expected_upstream_ttft": "5ms",
+    "chunk_interval": "1ms"
+  },
+  "metadata": {
+    "generated_at": "2026-08-10T01:27:47.399332Z",
+    "git_commit": "25c481a387365846620375080b83795af6227add",
+    "git_dirty": true,
```
astaxie: [P2] Regenerate the tracked baselines from a clean revision

All four baseline files record base commit `25c481a` with `git_dirty: true`, but the benchmark command itself does not exist at that commit. Checking out the recorded revision cannot reproduce these numbers or identify the code that was measured. Please regenerate the baselines from the clean PR head (or explicitly stamp the actual source revision) before tracking them.

backend/internal/perfbench/analyze_test.go
```diff
@@ -0,0 +1,90 @@
...
+	if result.Summary.SuccessRatePercent < 33.3 || result.Summary.SuccessRatePercent > 33.4 {
+		t.Fatalf("success rate does not include offered drops: %v", result.Summary.SuccessRatePercent)
+	}
+}
+
+func TestAnalyzeSeparatesStreamingTTFTAndTotalUpstreamTime(t *testing.T) {
+	t.Parallel()
+
+	result := perfbench.Analyze(perfbench.Config{
+		Stream:                   true,
```
astaxie: [P1] Format this file before merge

`gofmt` removes one alignment space on this line. The required Backend check currently stops here and the aggregate CI job fails, so the PR cannot pass the documented backend gate until `gofmt -w internal/perfbench/analyze_test.go` is applied.

astaxie (commented):
Re-reviewed current head `01f34ff`. All five prior findings are fixed: the internal Go benchmark matrix now has median `ns/op` / `B/op` / `allocs/op` budget enforcement; payload-audit persistence has a same-request on/off control plus direct rendering benchmarks; black-box and Go budget checks reject incompatible runtime/hardware profiles; tracked baselines identify clean reproducible source revisions; and the formatting/lint failure is corrected. I found no new blocking issue in the current complete diff.

Validation in an isolated worktree passed `go test ./...`, `go vet ./...`, focused race tests, all 116 repository tests, documentation/UI translation gates, environment and source-line gates, shell syntax, `git diff --check`, and `TOKENHUB_BENCHMARK_BENCHTIME=50ms TOKENHUB_BENCHMARK_COUNT=1 ./benchmarks/run-internal.sh` with all 36 budget checks passing. GitHub CI is fully green and the head is conflict-free. The opt-in PostgreSQL performance benchmark was not run locally because `TOKENHUB_BENCHMARK_POSTGRES_URL` was not configured; GitHub Backend PostgreSQL is green.

From code and CI review, this head is merge-ready. GitHub does not allow astaxie to approve a pull request authored by astaxie, so this is submitted as COMMENTED rather than APPROVE.

docs/performance-benchmarking.md
```diff
@@ -0,0 +1,148 @@
...
+  --label tokenhub-stream-c32 \
+  --base-url http://127.0.0.1:8080 \
+  --model benchmark-model \
+  --protocol chat \
+  --stream \
+  --mode concurrency \
+  --concurrency 32 \
+  --warmup 5s \
+  --duration 30s \
+  --upstream-latency 5ms \
```
astaxie: [P2] Make the streaming example use the documented total latency

With the default mocker shown above, total upstream time is 5 ms initial latency plus 8 x 1 ms chunk intervals, or 13 ms. This command records 5 ms for both total latency and TTFT, so it inflates full-response gateway overhead by about 8 ms and produces a scenario incompatible with the tracked stream baseline. Please use --upstream-latency 13ms together with --upstream-ttft 5ms, as the paragraph below already requires.

backend/internal/perfbench/runner.go
```diff
@@ -0,0 +1,356 @@
...
+				requests.Add(1)
+				go func() {
+					defer requests.Done()
+					defer func() { <-semaphore }()
+					results <- r.request(ctx)
+				}()
+			default:
+				results <- Observation{Error: "load_generator_saturated", Dropped: true}
+			}
+			timer.Reset(interval)
```
astaxie: [P1] Keep fixed-rate scheduling anchored to the requested rate

This command advertises fixed-rate load, but resetting a one-shot timer only after each dispatch makes loop overhead part of every interval and silently loses missed offers. Against the local zero-latency mocker, rate=100000 for 200ms should offer about 20,000 requests; this loop offered only 11,385 while reporting zero generator drops. Analyze then treats only recorded observations as offered work, so a latency run can execute far below its configured load and still pass the budget. Please schedule against an absolute next deadline, or otherwise account for elapsed ticks, and record missed scheduled offers as generator drops so underload cannot pass silently.

backend/internal/perfbench/analyze.go
```diff
@@ -0,0 +1,253 @@
...
+type LoadMode string
+
+const (
+	ModeRate        LoadMode = "rate"
+	ModeConcurrency LoadMode = "concurrency"
+)
+
+type Config struct {
+	Label                   string        `json:"label"`
+	BaseURL                 string        `json:"base_url"`
```
astaxie: [P2] Redact the target URL from exported results

BaseURL is serialized verbatim, while runner.go clears only APIKey and HTTPClient before producing the result. Since validation accepts URL userinfo, a value such as https://user:password@internal-gateway.example writes the username, password, and private hostname into every JSON result or tracked baseline. This contradicts the documentation promise that results exclude hostnames and usernames, and the comparison fingerprint does not use BaseURL. Please omit it from exported JSON or serialize only a redacted, non-identifying target, with a regression test for URL userinfo.

backend/internal/perfbench/runner.go
```diff
@@ -0,0 +1,403 @@
...
+				go func() {
+					defer requests.Done()
+					defer func() { <-semaphore }()
+					results <- r.request(ctx)
+				}()
+			default:
+				results <- Observation{Error: "load_generator_saturated", Dropped: true}
+			}
+			accountedOffers += dueOffers
+			if accountedOffers >= targetOffers {
```
astaxie, line 207: [P1] Keep the fixed-rate phase open for the configured duration

This exits as soon as the last scheduled offer is launched, so the measured elapsed time can be much shorter than Config.Duration. Offers start at t=0; against the local zero-latency mocker, rate=1 with duration=1s completed one request in about 0.6 ms and reported 1,570 achieved RPS instead of roughly 1 RPS. Fixed-rate throughput and any budget based on it are therefore invalid, especially at low rates. Please keep the phase open until deadline before finishing the drain, or otherwise ensure elapsed covers the full configured offer window.

backend/internal/perfbench/runner.go
```diff
@@ -0,0 +1,403 @@
...
+	targetOffers := scheduledOfferCount(deadline.Sub(started), r.config.Rate)
+	accountedOffers := 0
+	timer := time.NewTimer(max(0, time.Until(started)))
+	defer timer.Stop()
+	deadlineTimer := time.NewTimer(max(0, time.Until(deadline)))
+	defer deadlineTimer.Stop()
+loop:
+	for ctx.Err() == nil {
+		select {
+		case <-ctx.Done():
```
astaxie, line 178: [P2] Propagate mid-run context cancellation

When the context is canceled here, scheduling stops without accounting the remaining offers, but runPhase and Run still return a partial Result with nil error and the original configured duration. A caller can then persist or compare an interrupted run as though it completed normally. Please drain in-flight work as needed and return ctx.Err(), with a regression test that cancels during both rate and concurrency phases.

backend/internal/perfbench/runner.go
```diff
@@ -0,0 +1,407 @@
...
+	}
+	if config.Timeout <= 0 {
+		config.Timeout = 30 * time.Second
+	}
+	if config.RequestBytes <= 0 {
+		config.RequestBytes = 256
+	}
+	if config.MaxInFlight <= 0 {
+		config.MaxInFlight = 1000
+		if config.Rate > 0 && config.Rate <= int(time.Second) {
```
astaxie: [P2] Keep the implicit max-in-flight allocation bounded

When MaxInFlight is zero, this scales the default to Rate * 2, while validation accepts rates up to one billion requests per second. runRate then allocates the results channel with capacity min(MaxInFlight, Rate), so an accepted configuration such as --rate 100000000 --max-in-flight 0 can reserve memory for 100 million Observation values and be OOM-killed before sending a request. Please keep the documented default bounded at 1000, or validate a safe explicit upper bound, so overload is reported through generator drops instead of an allocation failure.

backend/internal/perfbench/runner.go
```diff
@@ -0,0 +1,404 @@
...
+	}
+	go func() {
+		workers.Wait()
+		close(results)
+	}()
+	return collectObservations(results)
+}
+
+func (r *loadRunner) runRate(ctx context.Context, started, deadline time.Time) []Observation {
+	results := make(chan Observation, min(r.config.MaxInFlight, max(1, r.config.Rate)))
```
astaxie: [P2] Bound the result buffer for explicit max-in-flight values

The implicit default is bounded now, but explicit values remain unchecked. An accepted configuration such as `--mode rate --rate 1000000000 --concurrency 0 --max-in-flight 1000000000` reaches `make(chan Observation, 1_000_000_000)`, reserving tens of gigabytes before sending a request and likely being OOM-killed. This is distinct from the previous implicit-default path. Please cap this results buffer independently, or reject unsafe explicit `MaxInFlight` values, so overload is reported through generator drops rather than an allocation failure.

PR #190, opened 2026-08-10
fix(provider): make Claude attribution cache-friendly

### LINKED ISSUE #135: feat(gateway): filter Claude Code attribution blocks per provider resource
## Background

Claude Code adds attribution information to the first `system` text block in Anthropic Messages requests, for example:

```json
{
  "type": "text",
  "text": "x-anthropic-billing-header: cc_version=2.1.220.134; cc_entrypoint=sdk-cli;"
}
```

Some versions have also used a dynamic `cch` format:

```text
x-anthropic-billing-header: cc_version=2.1.177.01c; cc_entrypoint=sdk-cli; cch=<dynamic value>;
```

Claude Code has stated that the Anthropic API removes this attribution block before processing, so it does not affect first-party Prompt Cache. When using a third-party LLM gateway, clients can omit the block with `CLAUDE_CODE_ATTRIBUTION_HEADER=0`.

In centralized gateway environments, however, clients often cannot be controlled or upgraded uniformly. If a third-party provider matches KV/prefix cache entries using the complete prompt prefix, the dynamic fingerprint in the initial attribution block changes the entire subsequent prefix and forces otherwise reusable system prompts and tool definitions to be prefetched again.

Related upstream Claude Code reports:

- https://github.com/anthropics/claude-code/issues/68900
- https://github.com/anthropics/claude-code/issues/50085

## Current TokenHub behavior

Verified against the latest TokenHub `main` at commit `0e99e77de091543d051b2b521022189d7fa0cb4c`:

1. OpenAI-compatible routes call `anthropicSystemText`, joining every text block in the Anthropic `system` array into one OpenAI `system` message separated by newlines. They do not identify or filter `x-anthropic-billing-header`:
   - https://github.com/astaxie/TokenHub/blob/0e99e77de091543d051b2b521022189d7fa0cb4c/backend/internal/server/anthropic_messages.go#L220-L229
   - https://github.com/astaxie/TokenHub/blob/0e99e77de091543d051b2b521022189d7fa0cb4c/backend/internal/server/anthropic_messages.go#L310-L327
2. Native Anthropic routes build the upstream payload from the original request and also preserve this `system` block:
   - https://github.com/astaxie/TokenHub/blob/0e99e77de091543d051b2b521022189d7fa0cb4c/backend/internal/server/anthropic_messages.go#L763-L772
   - https://github.com/astaxie/TokenHub/blob/0e99e77de091543d051b2b521022189d7fa0cb4c/backend/internal/server/codex_compat_anthropic_native.go#L8-L50
3. The repository contains no `x-anthropic-billing-header`, `billing-header`, or corresponding provider setting.

Consequently, when TokenHub connects to a third-party provider such as OpenAI-compatible or SGLang, this attribution text becomes part of the prompt actually sent to the upstream model.

## Local verification

Environment:

- Claude Code `2.1.220`
- Isolated local TokenHub environment
- Anthropic `/v1/messages` → TokenHub → OpenAI-compatible provider

TokenHub audit captured nine real requests, all with the attribution block as the first `system` block. Observed `cc_version` suffixes included:

```text
2.1.220.25c
2.1.220.134
2.1.220.90f
2.1.220.150
2.1.220.681
```

Five consecutive requests in one multi-turn call all used `.134`, showing that in `2.1.220` the value does not simply randomize once per API request. Different initial user inputs/sessions still produce different fingerprints, however, breaking shared-prefix reuse across sessions at third-party providers.

Inspection of the local `2.1.220` executable also showed:

- The attribution block is at the beginning of the system prompt.
- The `cc_version` suffix is a three-digit hexadecimal fingerprint derived from the first non-meta user text.
- First-party/Vertex paths may also append `cch=00000`; older versions have used a dynamic five-digit `cch`.
- `CLAUDE_CODE_ATTRIBUTION_HEADER=0` is the client-side disable switch.

Therefore, matching should use the attribution block's stable prefix rather than a fixed version or one `cch` format.

## Impact

This does not affect TokenHub authentication or basic forwarding, but it affects third-party inference backends with automatic prefix caching:

- System-prompt and tool-definition prefixes cannot be reused reliably across sessions.
- Prefill computation and input-token cost increase.
- TTFT increases with long system prompts, many tools, or high concurrency.
- Upstream cache-hit statistics and capacity estimates become distorted.

Impact depends on the provider's cache implementation. It is incorrect to state that every request must have a zero cache-hit rate: an in-session prefix may still hit when its fingerprint remains stable, and first-party Anthropic removes the block server-side.

## Proposed solution

Add a Claude Code attribution-block policy at **provider-resource** scope, configurable through the Admin API and UI:

```text
preserve  Keep the block unchanged
strip     Remove the block only from requests sent to this resource
```

Suggested defaults:

- Native Anthropic provider: `preserve`
- Other providers: retain `preserve` for compatibility and require administrators to enable `strip` explicitly per resource. A later release could recommend `strip` for OpenAI-compatible providers.

Processing rules:

1. Match only the complete first `type=text` block in the `system` array whose text begins with `x-anthropic-billing-header:`.
2. Do not alter ordinary user messages, historical tool results, or later system blocks, avoiding accidental deletion of business content.
3. Do not depend on whether `cch` exists, its width, the `cc_version` suffix length, or field order.
4. Preserve the order, text, and `cache_control` of every other system block.
5. Apply the policy while building the upstream payload for each route attempt rather than mutating the global request early. If one request fails over from a third-party resource to an Anthropic resource, the attempts can apply `strip` and `preserve` independently.
6. TokenHub audit continues to store the original client request. Route Attempts may record whether the policy was applied but need not record or expose the dynamic fingerprint.

If client version statistics need to be retained, a future `normalize` policy could keep the stable base version and entrypoint while removing the dynamic fingerprint. Because Claude Code's field format changes, full-block `strip` is safer for the first phase.

## Suggested acceptance criteria

- Two requests differing only in their dynamic fingerprint produce identical upstream system-prompt prefixes in `strip` mode.
- In `preserve` mode, upstream requests match current behavior.
- When one model has resources with different policies, each route attempt applies the selected resource's policy.
- If a third-party resource fails and the request switches to an Anthropic resource, the Anthropic request still preserves the attribution block.
- Streaming, buffered, Tool Use, and multi-turn Messages requests behave consistently.
- Ordinary text that happens to contain `x-anthropic-billing-header:` or `cch=` is not modified.
- Tests cover dynamic `cch`, absent `cch`, dynamic `cc_version` suffixes, and different field order.

## Temporary workaround

Clients can currently set:

```bash
export CLAUDE_CODE_ATTRIBUTION_HEADER=0
```

This requires configuration on every Claude Code client and does not replace centralized provider-resource governance at the gateway.

legendtkl:
## Summary

Improve prompt-prefix cache reuse for Claude Code requests routed through non-official Providers while preserving compatibility for official Anthropic and existing Provider configurations.

## Related Issue

Fixes #135

## Changes

- Add Provider and Provider Resource policies for preserving or stripping the Claude Code attribution system block.
- Default new official Anthropic Providers to `preserve`, known non-official Providers to `strip`, and unknown custom Anthropic endpoints to `preserve`.
- Apply the effective policy independently for each route attempt without mutating audit payloads or failover requests.
- Preserve legacy behavior for existing and migrated Providers that do not define the policy.
- Add admin console controls, English/Chinese/Japanese documentation, and focused backend/frontend coverage.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor or maintenance
- [x] Documentation

## Verification

- `node --test tools/*.test.mjs` (103 tests passed)
- `git diff --check`
- Browser regression verified the non-official `strip` default and official Anthropic `preserve` default.
- Independent read-only review reported no findings.
- SDK smoke tests were not run because they require configured external upstream credentials.

The API change is additive. Existing Providers without the setting continue to preserve the attribution block, and migrated legacy Providers retain that behavior. Provider Resources inherit the Provider setting unless explicitly overridden. No persistence schema, environment variable, credential-handling, deployment, or rollout changes are required. Audit payloads retain the original client request.

## Checklist


astaxie (commented):
Reviewed the complete current diff at b12a630 against issue #135. I found no additional code-level blocker in the attribution policy itself; focused backend tests/vet, frontend typecheck/build, the attribution gate, and diff check passed in an isolated worktree. This head still cannot merge because GitHub reports merge conflicts and has no CI checks. Resolve against current main and let the full repository CI run before requesting approval.

PR #191, opened 2026-08-10
feat(providers): add custom upstream request headers

### LINKED ISSUE #165: feat(providers): configure custom upstream request headers
## Background

Some upstream services require fixed request headers to identify the client, select a tenant, or enable compatibility behavior, such as a custom `User-Agent`, `X-Client-Name`, or another vendor-defined header. Administrators currently cannot configure these headers reliably for providers in the TokenHub console and must rely on an external reverse proxy or server code changes.

## Current state

- The backend `Provider` and `ProviderResource` data models already include a `headers` field, and the Admin API can accept and persist it.
- OpenAI-compatible, native Anthropic, and some image request paths apply provider headers; a provider resource header overrides a provider header with the same name.
- Provider/resource forms, frontend types, and submitted payloads do not expose `headers`, so ordinary administrators cannot view or maintain them.
- Support for custom request headers is inconsistent across provider adapters, model discovery, connection tests, and health checks.
- The backend lacks unified reserved-header validation and sensitive-value display rules. Allowing authentication or transport headers to be overridden directly introduces security and compatibility risks.

## Goal

Allow administrators to configure fixed custom request headers sent upstream at both provider and provider-resource levels, with consistent merge and security rules for runtime requests, connection tests, and model discovery. Custom `User-Agent` is the primary use case, along with other safely permitted extension headers.

### Console

- Add a key-value editor for custom request headers to provider connection or advanced settings, supporting multiple rows that can be added, edited, or deleted.
- Use the same editor for provider resources, showing headers inherited from the provider and the effective resource overrides.
- Allow values to be marked sensitive. After saving, sensitive values show only a masked state; blank retains the existing value and explicit deletion clears it.
- Validate header names, duplicates, empty values, case-equivalent entries, and illegal newlines, showing clear errors before saving.
- Provide a connection test or request preview that displays only effective header names and redacted values, excluding API keys and other system credentials.

Example:

```text
User-Agent: TokenHub-Custom-Client/1.0
X-Client-Name: internal-ai-platform
```

### Scope and precedence

- Provider request headers are defaults for the channel.
- A provider resource header overrides a provider default with the same case-insensitive name.
- System-managed authentication and HTTP transport headers have the highest precedence and cannot be overridden by custom configuration.
- Ordinary client headers are not forwarded upstream automatically; this feature applies only fixed values explicitly saved by an administrator.
- On routing failover, recompute headers for the selected provider resource on every attempt and never reuse values from the previous route.

### Security boundary

- Forbid hop-by-hop or transport headers including `Host`, `Content-Length`, `Transfer-Encoding`, `Connection`, and `Proxy-Authorization`.
- Forbid overriding adapter-managed authentication and protocol-critical headers such as `Authorization`, `api-key`, `x-api-key`, and `Content-Type`. Future custom authentication should use a separate controlled credential capability.
- Permit `User-Agent` and ordinary `X-*` extension headers by default, while still validating name, length, total size, and CR/LF injection.
- Encrypt and mask sensitive custom-header values using the same policy as provider credentials. Never expose them in ordinary API responses, request logs, errors, or audit details.
- Audit header names, scope, and add/update/delete actions without recording plaintext values.

### Request-path coverage

- For API-key providers supporting custom headers, consistently cover `/v1/chat/completions`, `/v1/responses`, `/v1/embeddings`, `/v1/messages`, image generation/editing, and buffered and streaming requests.
- Provider connection tests, health checks, model-catalog discovery, and model synchronization must use the same effective headers so upstreams requiring them can be configured successfully.
- Explicitly define the allowed scope for adapters such as Azure and Codex Subscription that have dedicated authentication or client identity requirements. Unified implementation must not alter existing protocol headers or identity emulation.
- New adapters should apply validated headers through a shared helper rather than implementing different rules in each request path.

## Compatibility

- Preserve current upstream request behavior when no custom headers are configured.
- Existing `headers` data saved through the Admin API must continue to work and load safely into the new UI for editing.
- Define a migration or read-time validation policy for arbitrary existing header data. Noncompliant entries should prompt administrator action and must not be sent upstream silently.

## Acceptance criteria

- [ ] Administrators can add, edit, and delete custom request headers on provider and provider-resource pages.
- [ ] A custom `User-Agent` reaches the upstream in buffered and streaming requests.
- [ ] Provider resources inherit provider headers and override matching names case-insensitively.
- [ ] Routing and failover use the correct headers for the actual resource without cross-resource leakage.
- [ ] Tests cover Chat Completions, Responses, Embeddings, Anthropic Messages, image requests, and applicable connection-test/model-discovery paths.
- [ ] Reserved headers, invalid names, CR/LF, overlong values, and excessive total size are rejected.
- [ ] Custom headers cannot override system authentication, protocol-critical, or hop-by-hop headers.
- [ ] Sensitive values are encrypted at rest and masked in API responses, with no plaintext leakage in logs, errors, or audit.
- [ ] Providers without configured headers remain compatible with current behavior.
- [ ] English, Simplified Chinese, and Japanese documentation explains configuration, precedence, coverage, and security restrictions.

astaxie:
## Summary

Adds a secure, consistent way for administrators to configure fixed upstream request headers at both Provider and Provider Resource scope. This removes the need for external proxies when upstreams require identifiers such as `User-Agent` or tenant headers, while preserving TokenHub-managed authentication and protocol headers.

## Related Issue

Closes #165

## Changes

- Add Provider and Provider Resource header editors with inheritance, case-insensitive Resource overrides, validation, masked previews, and English/Japanese translations for the canonical Simplified Chinese UI copy.
- Validate and apply custom headers consistently across OpenAI-compatible, Anthropic, Gemini, image, streaming, model-discovery, connection-test, health-check, routing, and failover paths.
- Encrypt sensitive header values, mask management responses and previews, redact upstream errors and streams, and exclude values from audit snapshots.
- Reject authentication, protocol, forwarding, cookie, hop-by-hop, malformed, oversized, and unsupported Azure/Codex header configurations; surface legacy invalid data without applying it upstream.
- Add migration support through per-header `SecretRef` values and bump the canonical bundle schema to `2.0.0`.
- Document configuration, precedence, adapter support, and security limits in English, Simplified Chinese, and Japanese.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor or maintenance
- [x] Documentation

## Verification

- `node --test tools/*.test.mjs` (121 tests passed)
- `node tools/check-doc-translations.mjs --base main --head HEAD`
- `git diff --check main...HEAD`
- SDK smoke tests were not run because no compatible external backend and credentials were configured; the in-process backend tests cover the changed request paths.

The OpenAI-compatible `/v1` API contract remains unchanged when no custom headers are configured. Management payloads add optional `sensitive_headers` and `header_validation_errors` fields. The migration bundle schema advances from `1.1.0` to `2.0.0` so older readers reject bundles containing security-bearing per-header secret references instead of silently dropping them.

Sensitive values use the existing AES-GCM secret storage, are masked in management surfaces, and are redacted from upstream errors, streamed events, and audit details. Authentication, API key, cookie, forwarding, protocol-critical, hop-by-hop, and transport headers cannot be overridden. Azure OpenAI and Codex Subscription adapters reject new custom headers; legacy invalid values are reported and disabled until removed. No environment, deployment, or database migration operation is required; rollback is the normal application rollback to the prior version.

## Checklist


backend/internal/migration/bundle/version.go
```diff
@@ -9,7 +9,7 @@ import (
 // SchemaVersion is the current version of the CanonicalMigrationBundle
 // schema shipped by this package. Follow semver: bump the major on any
 // breaking change to the on-disk JSON layout.
-const SchemaVersion = "1.1.0"
+const SchemaVersion = "1.2.0"
```
astaxie: [P1] Please make this a major schema bump (and update the schema ID), or otherwise make old readers reject security-bearing unknown fields. Released 1.1 readers accept every same-major version, then `json.Unmarshal` silently drops the new `header_secrets` fields and validation re-marshals that already-truncated struct. An old CLI can therefore accept a 1.2 bundle and migrate providers/resources without their required sensitive headers.

backend/internal/server/provider_error_classification.go
```diff
@@ -113,14 +114,37 @@ func newProviderMisconfigured(message string) error {
...
+		values = append(values, value)
+	}
+	for _, name := range provider.SensitiveHeaders {
+		if value, ok := providerHeaderValue(provider.Headers, name); ok && value != "" && value != providerHeaderMask {
+			values = append(values, value)
+		}
+	}
+	sort.Slice(values, func(i, j int) bool { return len(values[i]) > len(values[j]) })
+	for _, value := range values {
+		message = strings.ReplaceAll(message, value, providerHeaderMask)
```
astaxie: [P1] Please redact encoded representations as well as the literal value. Upstream JSON escapes string contents, so a configured secret such as `secret"token` appears in the raw body as `secret\"token` and this replacement does not match; the value then reaches client and audit errors (streaming paths reuse this helper). The same applies to JSON's escaped `<`, `>`, and `&`. Consider redacting decoded JSON string values or matching JSON-escaped forms, with a regression test.

backend/internal/migration/sink/tokenhub/store_sink.go
```diff
@@ -829,7 +852,8 @@ func sameProvider(existing server.Provider, desired server.Provider) bool {
 	if desired.Priority != 0 && existing.Priority != desired.Priority {
 		return false
 	}
-	return reflect.DeepEqual(normalizeStringMap(existing.Headers), normalizeStringMap(desired.Headers)) &&
+	return reflect.DeepEqual(normalizeHeaderMap(existing.Headers, existing.SensitiveHeaders), normalizeHeaderMap(desired.Headers, desired.SensitiveHeaders)) &&
```
astaxie, line 864: [P1] This comparison prevents sensitive-header rotation. Both the masked existing value and the newly resolved desired plaintext are normalized to `<sensitive>`, so changing a resolver value from `old-secret` to `new-secret` still returns equal and `applyProvider` skips the update; `sameProviderResource` has the same behavior. I reproduced this with two applies against `MemoryStore`: the second apply reported `ActionSkip` and retained `old-secret`. Please provide an explicit way to force/identify secret updates rather than treating all sensitive values as equal.

astaxie (commented):
Security and migration review: three confirmed issues are noted inline.

astaxie (commented):
Follow-up review: three confirmed edge cases are noted inline.

backend/internal/server/provider_error_classification.go
```diff
@@ -113,14 +115,72 @@ func newProviderMisconfigured(message string) error {
...
+	return []byte(message)
+}
+
+func providerSecretRepresentations(value string) []string {
+	representations := []string{value}
+	encoded, err := json.Marshal(value)
+	if err != nil || len(encoded) < 2 {
+		return representations
+	}
+	escaped := string(encoded[1 : len(encoded)-1])
```
astaxie, line 260: [P1] This still misses valid JSON escape spellings. For a secret `secret`, an upstream JSON error value encoded as `\u0073ecret` contains neither the literal value nor Go's preferred `json.Marshal` representation, so this helper returns it unchanged while every JSON decoder reconstructs the full secret; mixed Unicode escapes and surrogate pairs provide the same bypass. Please redact decoded JSON string values before re-marshalling (with literal fallback for non-JSON bodies), and add a regression test using an arbitrary escape such as `\u0073`.

backend/internal/migration/sink/tokenhub/store_sink.go
```diff
@@ -513,17 +513,28 @@ func (s *StoreSink) Plan(b *bundle.CanonicalMigrationBundle) (*MigrationReport,
...
+	if err := server.ValidateProviderHeaderConfigForWrite(&spec); err != nil {
+		return Change{}, fmt.Errorf("validate provider headers %s: %w", item.ExternalRef.ID, err)
+	}
 
 	for _, existing := range s.store.ListProviders() {
 		if existing.Name == spec.Name && existing.Type == spec.Type {
-			if sameProvider(existing, spec) {
+			// Header secrets are write-only on the target and their resolver values
+			// may rotate without changing the bundle, so every apply reasserts them.
+			if len(item.HeaderSecrets) == 0 && sameProvider(existing, spec) {
```
astaxie, line 543: [P1] Forcing a full update here also overwrites target-owned/runtime fields. I reproduced: apply a bundle with `HeaderSecrets`, mark its Provider and Resource unhealthy via the health APIs, then apply the unchanged bundle again; both become healthy because `UpdateProvider`/`UpdateProviderResource` write `spec.Healthy`. Resource updates can likewise clear optional BaseURL/region/rate/token/concurrency values that the bundle omitted even though `sameProviderResource` treats those zero values as preserve. The HTTP sink has the same full-update behavior. Please merge existing target-owned fields before refreshing, or add a header-secret-specific update path.

backend/internal/server/provider_headers.go
```diff
@@ -0,0 +1,497 @@
...
+		}
+		canonicalName := http.CanonicalHeaderKey(name)
+		if seen[canonicalName] {
+			return nil, NewHTTPError(http.StatusBadRequest, "provider_header_duplicate", "Custom request header names must be unique ignoring case")
+		}
+		seen[canonicalName] = true
+		if value == "" {
+			return nil, NewHTTPError(http.StatusBadRequest, "provider_header_value_required", "Custom request header value cannot be empty")
+		}
+		if strings.ContainsAny(value, "\r\n") {
```
astaxie: [P2] Please reject every value that Go's HTTP transport cannot send, not only CR/LF. For example a custom header value containing NUL (`a<NUL>b`) is accepted and persisted here, but `http.Transport` rejects the resulting request with `invalid header field value`, leaving the Provider/Resource unusable; DEL and other control bytes behave similarly. Use a header-field-value validation equivalent to the transport and mirror it in frontend validation/tests.

backend/internal/server/provider_error_classification.go
```diff
@@ -113,14 +116,142 @@ func newProviderMisconfigured(message string) error {
...
+	for _, name := range provider.SensitiveHeaders {
+		if value, ok := providerHeaderValue(provider.Headers, name); ok && value != "" && value != providerHeaderMask {
+			values = append(values, value)
+		}
+	}
+	if len(values) == 0 {
+		return data
+	}
+	sort.Slice(values, func(i, j int) bool { return len(values[i]) > len(values[j]) })
+	if redacted, changed := redactProviderJSONSecrets(data, values); changed {
```
astaxie: [P1] Redact decoder-equivalent secrets even when the bounded upstream body is truncated. checkProviderResponseForProvider reads only providerErrorBodyPrefix bytes before this call; if a larger JSON error is cut before its closing delimiters, redactProviderJSONSecrets rejects it and the fallback cannot recognize arbitrary valid escapes. I reproduced this with a prefix containing "\\u0073ecret" followed by padding to the limit: the returned error still contains "\\u0073ecret", which a JSON decoder reconstructs as the configured secret. Please make the bounded/truncated path escape-aware (or tokenize JSON strings incrementally) and add this exact regression case.

astaxie (commented):
Re-reviewed the complete current diff at c558833. The six prior findings are fixed and the full backend, frontend, repository, translation, environment, source-line, and diff checks pass in an isolated worktree. One new security blocker remains in the bounded upstream-error redaction path, so this head is not merge-ready.

backend/internal/server/provider_error_classification.go
```diff
@@ -113,14 +118,311 @@ func newProviderMisconfigured(message string) error {
...
+				end += 6
+			} else if 0xDC00 <= first && first <= 0xDFFF {
+				return string(decoded), spans, len(data)
+			}
+			encoded := make([]byte, utf8.RuneLen(runeValue))
+			utf8.EncodeRune(encoded, runeValue)
+			appendDecoded(encoded, start, end)
+			index = end
+		default:
+			return string(decoded), spans, len(data)
```
astaxie: [P1] Please do not let one malformed string terminate scanning of the entire body. I reproduced this with secret `secret` and a pending SSE error whose JSON contains a `bad` string with an invalid backslash-q escape before an `error` string that encodes its initial `s` as Unicode U+0073. Full JSON decoding fails, this branch returns `next == len(data)`, and the outer scanner never visits the later error string; the fallback also cannot match that Unicode spelling, so the decoder-equivalent secret is forwarded unchanged. The invalid-hex and invalid-surrogate branches have the same fail-open behavior. Resume scanning after the failed candidate and add a malformed-string-before-escaped-secret regression test.

astaxie (commented):
Latest follow-up review: one confirmed malformed-input redaction bypass is noted inline.

backend/internal/server/provider_error_classification.go
```diff
@@ -113,14 +118,329 @@ func newProviderMisconfigured(message string) error {
...
+// each JSON string that can be read from data. Unlike redactProviderJSONSecrets,
+// it does not require the enclosing object, array, or final string to be closed.
+func redactProviderJSONPrefixSecrets(data []byte, secrets []string) ([]byte, bool) {
+	replacements := make([]providerJSONReplacement, 0)
+	for offset := 0; offset < len(data); {
+		if data[offset] != '"' {
+			offset++
+			continue
+		}
+		decoded, spans, next := decodeProviderJSONStringPrefix(data, offset)
```
astaxie: [P1] Please avoid scanning every quote in raw pending SSE as if it belonged to one JSON document. Valid SSE metadata can desynchronize these candidates: I reproduced with secret `secret`, an `id:` line whose value is a single quote, followed by a `data:` error whose value encodes the initial `s` as Unicode U+0073. The metadata quote pairs with the JSON key quote, subsequent candidates remain shifted, and the escaped secret is forwarded unchanged when the stream fails before the blank terminator. An invalid escape followed by the secret in the same malformed string is skipped similarly. Redact the assembler's parsed `Data` fields and rewrite the frame, or fail closed when string recovery becomes ambiguous; add metadata-quote and same-string regression tests.

backend/internal/server/provider_error_classification.go
```diff
@@ -113,14 +118,329 @@ func newProviderMisconfigured(message string) error {
...
+// decodeProviderJSONStringPrefix decodes complete characters from the JSON
+// string beginning at quote. Each decoded byte retains the source range that
+// produced it, allowing a decoded match to be replaced in the original prefix.
+func decodeProviderJSONStringPrefix(data []byte, quote int) (string, []providerJSONByteSpan, int) {
+	decoded := make([]byte, 0)
+	spans := make([]providerJSONByteSpan, 0)
+	appendDecoded := func(value []byte, start, end int) {
+		decoded = append(decoded, value...)
+		for range value {
+			spans = append(spans, providerJSONByteSpan{start: start, end: end})
```
astaxie: [P1] This per-byte span table creates severe memory amplification on malformed streamed errors. A local one-shot benchmark with a 1 MiB unterminated quoted value allocated 94,338,208 bytes; the SSE decoder permits an event up to 8 MiB specifically to bound memory, so one event can drive allocations into the hundreds of MiB and concurrent upstream failures can exhaust the gateway. Please track source ranges only for actual matches, use a compact/streaming mapping, or impose a much smaller redaction scan bound, with a near-limit allocation regression or benchmark.

astaxie (commented):
Follow-up review: two confirmed issues remain in the malformed/pending error redaction path.

astaxie (commented):
Re-reviewed the complete current diff at 62b0e1f. The fail-closed malformed/truncated JSON path and parsed-SSE-data rewrite address the four pending redaction findings without the previous per-byte allocation amplification; those threads are resolved. Required CI is green. GitHub still reports merge conflicts with current main, so this head cannot be approved until it is rebased/resolved and CI reruns.

PR #192, opened 2026-08-11
feat(codex): renew tokens and configure image capability

Mr0bean:
## Summary

Renew OpenAI/Codex account OAuth access tokens proactively, provide an operator-controlled renewal action, and let administrators test and configure the subscription-backed `codex-gpt-image-2` capability directly from the Codex Provider model list.

## Related Issue

N/A.

## Changes

- Start a one-minute credential renewal scheduler that renews active OAuth account tokens within the existing five-minute refresh window.
- Serialize renewal with the existing cluster lease and mark invalidated refresh tokens as requiring reauthorization instead of repeatedly retrying.
- Add the `续租 Token` action to the Provider advanced subscription account card and show a clear reauthorization message.
- Display `codex-gpt-image-2 ← gpt-image-2` in the OpenAI Codex Provider model list with a checkbox-driven setup flow.
- Require an administrator to select a real active Codex account and acknowledge that the low-quality image probe consumes a small amount of subscription quota.
- Create or re-enable the Provider route only after the probe returns a non-empty, decodable image; categorize unsupported, reauthorization, rate-limit, timeout, and transient failures without exposing credentials.
- Serialize image capability tests per Provider so concurrent requests cannot create duplicate routes or duplicate probes.
- Disable all matching routes when the checkbox is cleared while preserving the account capability result.
- Backfill one missing route for active accounts already confirmed as supported, without re-enabling explicitly disabled routes or repeatedly recreating a route deleted after migration.
- Keep advanced route priority, weight, project scope, resource, and resource-group controls available in routing.
- Continue low-frequency capability retries only through real image traffic on an existing route; never generate background probe images.
- Add backend, race, frontend domain, routing, model-publication, OAuth renewal, and regression coverage.
- Document OAuth renewal and the image capability setup flow in English, Simplified Chinese, and Japanese.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Refactor or maintenance
- [x] Documentation

## Verification

- `go test ./...`: passed against a clean snapshot of the committed tree; `internal/server` completed in 137.802 seconds.
- `go test -race ./internal/server -run TestAdminCodexImageCapabilityTestsOnceAndManagesRoute -count=1`: passed.
- `go test ./internal/server -run Test(AdminCodexImageCapability|CodexImageRouteBackfill) -count=1`: passed.
- `go vet ./...`: passed.
- `npm test`: passed, 38 tests.
- `npm run typecheck`: passed.
- `npm run build`: passed.
- `node --test tools/*.test.mjs`: passed, 122 tests.
- Documentation translation, UI translation, environment contract, source line, and `git diff --check` gates: passed.
- Local admin-page browser smoke test: `/providers` loaded without console errors. The local database contained no real Codex Provider records, so no synthetic account or quota-consuming image probe was created.
- [GitHub Actions run 31822668200](https://github.com/astaxie/TokenHub/actions/runs/31822668200): Backend, Backend PostgreSQL, Frontend, Deployment, Repository gates, and aggregate CI all passed.
- Production OAuth validation: changed account 7455 to a near-expiry time, then verified automatic renewal persisted a new expiry and credential blob; service health was `healthy` and `/tokenhub/providers` returned HTTP 200.

The OpenAI-compatible `/v1` image API contract is unchanged. The new administrator endpoint accepts only an explicit enable/disable request for an OpenAI Codex subscription resource and returns capability and route identifiers without credential data. Enabling image capability deliberately sends one real low-quality image request and therefore consumes a small amount of subscription quota; no background probe is scheduled. Route creation is idempotent and cluster-serialized. Upgrade backfill is one-time and preserves explicit operator opt-outs. OAuth credentials remain encrypted, invalidated refresh tokens enter an explicit reauthorization state, and the renewal scheduler uses the existing cluster lease to prevent concurrent token rotation across replicas. No database schema or environment variable changes are introduced.

## Checklist


backend/internal/server/provider_credential_refresh_service_test.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+}
+
+func TestProviderCredentialRefreshServiceSkipsHealthyOrUnsupportedResources(t *testing.T) {
+	store := NewMemoryStore()
+	provider := store.AddProvider(Provider{Name: "OAuth Resources", Type: ProviderOpenAICodex, Status: StatusActive, Healthy: true})
+	for _, resource := range []ProviderResource{
+		{ProviderID: provider.ID, Name: "No Refresh Token", ResourceType: ProviderResourceOpenAISubscription, Status: StatusActive, Healthy: true, Credentials: &ProviderResourceCredentials{AccessToken: "access-no-refresh", ExpiresAt: time.Now().UTC().Add(time.Minute).Format(time.RFC3339)}},
+		{ProviderID: provider.ID, Name: "Disabled", ResourceType: ProviderResourceOpenAISubscription, Status: StatusDisabled, Healthy: true, Credentials: &ProviderResourceCredentials{AccessToken: "access-disabled", RefreshToken: "refresh-disabled", ExpiresAt: time.Now().UTC().Add(time.Minute).Format(time.RFC3339)}},
+		{ProviderID: provider.ID, Name: "API Key", ResourceType: ProviderResourceAPIKey, Status: StatusActive, Healthy: true, APIKey: "api-key"},
```
astaxie: [P1] Fix the failing test setup. This adds an API-key resource to a Codex Subscription Provider, but the store correctly rejects that adapter/resource combination with "Codex Subscription Providers only accept subscription account resources". As a result TestProviderCredentialRefreshServiceSkipsHealthyOrUnsupportedResources fails both in GitHub Backend and locally. Put the unsupported resource under a compatible Provider (or test the skip without violating the store contract) so the new behavior is actually exercised.

backend/internal/server/provider_credential_refresh_service.go
```diff
@@ -0,0 +1,82 @@
...
+		if !isOpenAIAccountResource(resource.ResourceType) || resource.Status != StatusActive || resource.CredentialSummary["has_refresh_token"] != "true" || resource.CredentialSummary[openAIAccountReauthorizationRequiredOption] == "true" {
+			continue
+		}
+		if _, err := s.store.RefreshProviderResourceCredentials(ctx, resource.ID, false); err != nil {
+			if AsHTTPError(err).Code == "provider_resource_reauthorization_required" {
+				if _, updateErr := s.store.UpdateProviderResourceOptions(resource.ID, map[string]string{openAIAccountReauthorizationRequiredOption: "true"}); updateErr != nil {
+					log.Printf("[tokenhub] could not mark provider resource %s for reauthorization: %v", resource.ID, updateErr)
+				}
+			}
+			log.Printf("[tokenhub] OAuth token renewal failed for provider resource %s: %v", resource.ID, err)
```
astaxie: [P1] Do not write the full OAuth refresh error to runtime logs. requestOpenAIAccountOAuthToken includes sanitizeOAuthErrorDetail(body) in this error, but when an OAuth endpoint returns a nested error object the sanitizer retains the original JSON prefix; that can include token/session material. This scheduler turns it into an unattended persistent log entry on every renewal attempt. Log only the stable HTTP error code/status here, and add a regression test proving upstream response bodies and credentials never reach the logger.

astaxie (commented):
Reviewed the complete current diff. Frontend and repository CI pass, but Backend/CI fail reproducibly on the new refresh-service test. The logging path below also needs to avoid persisting OAuth response details. No approval until both blockers are fixed and CI is green.

astaxie (commented):
Re-reviewed current head 8d7da7c. The new test fixture now exercises the intended unsupported-resource skip without violating the Provider resource contract, so that thread is resolved and required CI is green. The separate OAuth logging finding is unchanged: the scheduler still logs the full refresh error, which can contain sanitized-but-retained upstream response detail. This head also conflicts with current main; no approval until the logging path is fixed, conflicts are resolved, and CI reruns.

backend/internal/server/provider_credential_refresh_service.go
```diff
@@ -0,0 +1,82 @@
...
+	schedulerStop context.CancelFunc
+	schedulerDone chan struct{}
+}
+
+func newProviderCredentialRefreshService(store Store) *ProviderCredentialRefreshService {
+	return &ProviderCredentialRefreshService{store: store}
+}
+
+func (s *ProviderCredentialRefreshService) RunDue(ctx context.Context) {
+	for _, resource := range s.store.ListProviderResources() {
```
astaxie, line 30: [P1] Do not let one account block the entire renewal scan

RunDue refreshes every resource serially, while each OAuth request may block for 120 seconds. Because resources are not ordered by expiry, three slow/unreachable accounts ahead of another due account can consume more than the whole five-minute lead window; the later token expires before this loop reaches it, and missed ticker intervals are not queued while RunDue is blocked. Please use bounded parallelism and/or a much shorter per-resource deadline so one account cannot prevent unrelated accounts from being renewed on time.

backend/internal/server/provider_credential_refresh_service.go
```diff
@@ -0,0 +1,82 @@
...
+		if ctx.Err() != nil {
+			return
+		}
+		if !isOpenAIAccountResource(resource.ResourceType) || resource.Status != StatusActive || resource.CredentialSummary["has_refresh_token"] != "true" || resource.CredentialSummary[openAIAccountReauthorizationRequiredOption] == "true" {
+			continue
+		}
+		if _, err := s.store.RefreshProviderResourceCredentials(ctx, resource.ID, false); err != nil {
+			if AsHTTPError(err).Code == "provider_resource_reauthorization_required" {
```
astaxie: [P1] Persist reauthorization state in the core refresh path

Only the scheduler catches provider_resource_reauthorization_required and sets this option. The same RefreshProviderResourceCredentials function is also called by the manual refresh endpoint and request/quota paths; those callers return the reauthorization error without persisting the flag, so after a UI reload the warning disappears and they can keep retrying the invalidated refresh token. I reproduced this by forcing a manual refresh against an invalidated-token response: the error code is correct but CredentialSummary still lacks oauth_reauthorization_required. Please persist the state inside the leased refresh operation so every caller gets the same durable behavior (and avoids the post-lease race here).

backend/internal/server/provider_account_oauth.go
```diff
@@ -421,6 +424,11 @@ func refreshOpenAIAccountOAuthCredentials(ctx context.Context, current ProviderR
 	return creds, nil
 }
 
+func isOpenAIAccountOAuthReauthorizationRequired(err error) bool {
+	httpErr := AsHTTPError(err)
+	return httpErr.Code == "oauth_token_failed" && strings.Contains(strings.ToLower(httpErr.Message), "refresh_token_invalidated")
```
astaxie: [P1] Parse the invalidation code before truncating the response

This classification searches the already-sanitized error message, but sanitizeOAuthErrorDetail truncates nested OAuth JSON to 240 characters. I reproduced a valid nested response with a 260-character message followed by code=refresh_token_invalidated; this check returns false, so the scheduler retries the invalidated token forever instead of marking it for reauthorization. Please parse the structured OAuth error/code before formatting or truncating it, and test a nested response where the code is not in the first 240 bytes.

astaxie (commented):
Re-reviewed the complete current diff at `20eb7be6`. The only commit after astaxie's last review is a merge of current `main`; required GitHub CI and the focused credential-refresh tests pass. The four existing unresolved threads still apply to unchanged production paths: renewal remains serial behind a 120-second request timeout, the scheduler logs the full OAuth error, reauthorization state is persisted only by the scheduler rather than the core refresh operation, and invalidation classification still searches the already-truncated message. I reproduced the last issue with a nested `refresh_token_invalidated` code after a 260-character message: the diagnostic test fails because `sanitizeOAuthErrorDetail` removes the code. Please address those threads before requesting another approval.

One additional current-diff issue:

- **[P2] Use a complete locale-aware success template.** `frontend/features/admin/resources/provider-model-config.tsx:136` builds `${item.name} ${tx("Token 已续租")}` by concatenating provider content with a translated fragment. This violates the repository rule for dynamic sentences and cannot adapt word order per locale. Add one complete Chinese key such as `"{name} Token 已续租"`, provide English/Japanese entries, and format its `{name}` placeholder with the existing template pattern.

astaxie (commented):
Re-reviewed the complete current diff at `ae2963cf` after the code-only update. The four prior OAuth renewal findings are fixed and I resolved those threads: renewal is bounded-concurrent, scheduler logs contain only a stable code, reauthorization is persisted in the core refresh path, and invalidation codes are parsed before message truncation. The corresponding focused tests pass.

This head still cannot be approved:

1. **[P1] The required backend suite has four current-head failures.** I reproduced `go test ./internal/server -run 'TestConcurrentProviderResourceInsertWithSameIDIsIdempotent|TestProviderEditPreservesConcurrentReauthorizationMarker|TestCodexRouteFilteringUsesPerAccountModels|TestCodexUnsupportedModelFailsOverAndUpdatesAccountModels' -count=1`. The two new cross-store SQLite tests fail with `database is locked` at `image_capability_concurrency_test.go:598` and `:1060`, so the new transaction/upsert paths do not yet satisfy their own concurrency contract. The two existing per-account Codex routing tests pass on base `1abd8c0` but fail here after `AddProviderResource` unconditionally strips server-managed model-catalog options at `store_providers.go:330`; restore a trusted server-owned setup/update path (or adjust the tests and production path consistently) without allowing client-forged managed state.
2. **[P1] GitHub Backend/aggregate CI is red.** `golangci-lint` reports six unchecked `Callback().Query().Remove(callbackName)` returns in `image_capability_concurrency_test.go` (lines 332, 486, 794, 863, 979, 1045). Check or explicitly discard those errors so the required check passes.
3. **[P2] Correct the PR template's configuration impact.** The body says there are no environment-variable changes and leaves Deployment/configuration unchecked, but this diff removes `TOKENHUB_IMAGE_CAPABILITY_RETRY_SECONDS` from backend config, examples/startup, and all three deployment guides. State that old settings are now ignored, describe rollout/rollback impact, and mark the configuration change accurately.

Frontend, PostgreSQL, deployment, repository gates, targeted OAuth/image tests, and `git diff --check` pass. Please fix the backend regressions and lint, update the PR description, then rerun the full required checks.

astaxie (commented):
Re-reviewed the complete current diff at ecd82c8b after the code-only update. The previous backend/lint regressions are no longer present on this head: all current required GitHub checks are green, and the focused Codex image-capability, route-backfill, image-generation, and credential-refresh tests pass. I found no new implementation blocker in the current diff.

This head still cannot be approved because it conflicts with current main in backend/internal/server/admin_providers_http.go (confirmed with git merge-tree). Please resolve the conflict and rerun the required checks; the resulting head will need review before approval.

frontend/features/admin/views/provider-codex-image-capability.tsx
```diff
@@ -0,0 +1,150 @@
...
+      setBusy("");
+    }
+  }
+
+  return (
+    <>
+      <label className={`model-option provider-codex-image-option ${modelState}`}>
+        <input
+          checked={enabled}
+          disabled={Boolean(busy) || providerResources.length === 0}
```
astaxie, line 87: [P2] Keep disable available after the final account is deleted

The route is provider-level, but disabling is coupled to providerResources[0] and this control becomes disabled when providerResources is empty. DeleteProviderResource leaves a matching provider-level route active (it only clears explicit resource pins), so deleting the last Codex account produces a checked enabled_without_account state that cannot be cleared here, and the resource-scoped API has no remaining ID to call. Please make disable provider-scoped/independent of an extant account, or disable matching image routes when the final account is deleted, and add enable -> delete-last-account -> disable coverage.

frontend/features/admin/views/provider-codex-image-capability.tsx
```diff
@@ -0,0 +1,150 @@
...
+            </label>
+            {busy === "enable" ? (
+              <div className="provider-codex-image-testing" role="status">
+                <LoaderCircle aria-hidden="true" className="spin" size={20} />
+                <div><strong>{tx("正在测试生图能力")}</strong><span>{tx("正在等待真实 Codex 上游返回图片，请勿关闭弹窗。")}</span></div>
+              </div>
+            ) : null}
+            {error ? <div className="provider-codex-image-error" role="alert"><AlertCircle aria-hidden="true" size={18} /><span>{error}</span></div> : null}
+            {selectedResource?.options?.image_generation_capability ? (
+              <p className="provider-credential-note">{tx("该账号上次测试结果")}：{tx(formatImageGenerationCapabilityTag(selectedResource.options.image_generation_capability))}</p>
```
astaxie: [P2] Use a complete translated status template

This builds a user-visible dynamic sentence by concatenating tx("该账号上次测试结果"), a hard-coded Chinese colon, and a translated result fragment. AGENTS.md requires complete locale-aware templates instead of concatenated translated fragments because punctuation and word order vary by locale. Add a full Chinese template with a result placeholder, provide English/Japanese entries, and format the placeholder through the existing translation-template helper.

astaxie (commented):
Re-reviewed the complete current diff at 6659710c after the main merge. The conflict resolution correctly preserves main's explicit method routing and registers the image-capability POST path in both the route table and nested compatibility handler. All required GitHub checks are green; the full backend suite, go vet, frontend typecheck/build, 122 repository tests, translation/environment/source-line gates, diff check, and focused race tests pass locally.

This head still needs the two inline findings below addressed before approval: the image route can become impossible to disable from this setup flow after the final account is deleted, and one dynamic status sentence violates the repository's locale-aware template rule.

backend/internal/server/admin_codex_image_capability.go
```diff
@@ -0,0 +1,289 @@
...
+		}
+		if imageErr != nil {
+			if AsHTTPError(imageErr).Code == "codex_image_forbidden" {
+				_, updateErr := s.updateCodexImageCapability(current.ID, codexImageCapabilityUnsupported)
+				if updateErr != nil {
+					return updateErr
+				}
+				result.Capability = codexImageCapabilityUnsupported
+			}
+			return imageErr
```
astaxie: [P1] Do not return raw Codex upstream error details

`imageErr` can come from `classifyCodexHTTPError`, which copies the upstream `message` (or a non-JSON body) into `HTTPError.Message`; `writeError` then returns it verbatim from this admin endpoint. A failed probe can therefore expose account or session details contained in the upstream response, contrary to the credential-safety requirement. Map probe failures to stable public messages and codes at this boundary, and add a response containing a sentinel secret to prove it is never returned.

backend/internal/server/provider_resource_credentials.go
```diff
@@ -226,6 +232,23 @@ func (s *GormStore) RefreshProviderResourceCredentials(ctx context.Context, reso
 
 		refreshed, err := refreshOpenAIAccountOAuthCredentials(leaseCtx, creds)
 		if err != nil {
+			if httpErr := AsHTTPError(err); httpErr != nil && httpErr.Code == "provider_resource_reauthorization_required" {
+				s.mu.Lock()
+				var current ProviderResource
+				loadErr := s.db.WithContext(leaseCtx).First(&current, "id = ?", resourceID).Error
```
astaxie: [P2] Do not apply a stale failure to replacement credentials

The OAuth request runs after `s.mu` is released, so an administrator can complete reauthorization and save a new credential blob while this refresh of the old token is in flight. If the old request then returns `provider_resource_reauthorization_required`, this reloads the new row and unconditionally marks those replacement credentials invalid. Compare a credential version or token fingerprint captured before the request, or make credential replacement participate in the same lease, before persisting this marker; add a blocked-refresh → reauthorize → invalidation regression test.

backend/internal/server/admin_codex_image_capability.go
```diff
@@ -0,0 +1,289 @@
...
+		}
+		if len(response.Data) == 0 || strings.TrimSpace(response.Data[0].B64JSON) == "" {
+			return NewHTTPError(http.StatusBadGateway, "image_result_missing", "Codex image capability test completed without an image result")
+		}
+		if _, err := decodeGeneratedImage(response.Data[0].B64JSON); err != nil {
+			return NewHTTPError(http.StatusBadGateway, "image_result_invalid", err.Error())
+		}
+		route, err := s.ensureCodexImageRoute(currentProvider.ID)
```
astaxie, line 120: [P2] Serialize final-account deletion with capability enable

The deletion cleanup only disables routes that already exist. While this 120-second probe is in flight, the final account can be deleted because deletion does not take the provider `codex-image-capability` lease. If deletion commits before this line, it sees no route; the probe then creates an active provider-level route and the following capability update fails because the resource is gone, leaving the orphan route that the latest fix intended to prevent. Revalidate and create atomically with resource deletion or serialize deletion under the same provider lease, and add a blocked-probe → delete-final-account → release-probe regression test.

backend/internal/server/provider_resource_credentials.go
```diff
@@ -226,6 +232,23 @@ func (s *GormStore) RefreshProviderResourceCredentials(ctx context.Context, reso
 
 		refreshed, err := refreshOpenAIAccountOAuthCredentials(leaseCtx, creds)
 		if err != nil {
+			if httpErr := AsHTTPError(err); httpErr != nil && httpErr.Code == "provider_resource_reauthorization_required" {
```
astaxie, line 343: [P1] Short-circuit credentials already marked for reauthorization

Persisting this marker only stops `RunDue`; this core refresh path never checks it. Routed requests, quota and image calls, and manual renewals still enter `refreshOpenAIAccountOAuthCredentials` whenever the unchanged expiry is due, so a known-invalid refresh token can be submitted again on every request. Return `provider_resource_reauthorization_required` before network I/O while the marker is set, until replacement credentials clear it, and test that a second refresh call issues no second token request.

backend/internal/server/admin_codex_image_capability.go
```diff
@@ -0,0 +1,289 @@
...
+	}
+	if requireActive && (provider.Status != StatusActive || !provider.Healthy || resource.Status != StatusActive || !resource.Healthy) {
+		return ProviderResource{}, Provider{}, NewHTTPError(http.StatusConflict, "provider_resource_unavailable", "Enable the Codex subscription account before testing image generation")
+	}
+	return resource, provider, nil
+}
+
+func (s *Server) updateCodexImageCapability(resourceID string, capability string) (ProviderResource, error) {
+	options := map[string]string{
+		codexImageCapabilityOption:          capability,
```
astaxie, line 180: [P1] Preserve server-managed account state across ordinary edits

These capability fields are stored in `ProviderResource.Options`, but the normal resource PATCH replaces that map with the client payload. The frontend edit payload does not round-trip this capability, its check timestamp, the route-backfill marker, `oauth_reauthorization_required`, or `has_refresh_token`; saving an otherwise unchanged account can therefore remove the image model from publication while its route remains active, hide reauthorization state, and stop scheduled renewal. Preserve protected server-owned keys when applying account edits and add enable or reauthorization → edit account → state retained coverage.

backend/internal/server/admin_codex_image_capability.go
```diff
@@ -0,0 +1,314 @@
...
+					return updateErr
+				}
+				result.Capability = codexImageCapabilityUnsupported
+			}
+			return publicCodexImageCapabilityProbeError(imageErr)
+		}
+		if len(response.Data) == 0 || strings.TrimSpace(response.Data[0].B64JSON) == "" {
+			return NewHTTPError(http.StatusBadGateway, "image_result_missing", "Codex image capability test completed without an image result")
+		}
+		if _, err := decodeGeneratedImage(response.Data[0].B64JSON); err != nil {
```
astaxie, line 117: [P2] Validate that the decoded result is actually an image

`decodeGeneratedImage` only checks Base64 decoding, non-emptiness, and size, so Base64-encoded text such as `bm90IGFuIGltYWdl` passes this probe and enables the route. Normal image-job persistence later rejects the same bytes because `saveImageAsset` accepts only PNG, JPEG, or WebP, leaving a capability marked supported whose real requests fail. Apply the same content-type or image decoding validation before route creation and add a successful JSON response containing non-image Base64 as a regression case.

backend/internal/server/admin_provider_resources_routing.go
```diff
@@ -158,7 +163,21 @@ func (s *Server) serveAdminProviderResourcePatch(w http.ResponseWriter, r *http.
...
-	if err := s.store.DeleteProviderResource(resourceID); err != nil {
+	deleteResource := func() error { return s.store.DeleteProviderResource(resourceID) }
+	resource, found := s.store.GetProviderResource(resourceID)
+	if !found {
+		writeError(w, r, NewHTTPError(http.StatusNotFound, "provider_resource_not_found", "Provider resource not found"))
+		return
+	}
+	var err error
+	if resource.ResourceType == ProviderResourceOpenAISubscription {
+		err = s.store.RunClusterOperation(r.Context(), "codex-image-capability:"+resource.ProviderID, func(context.Context) error {
```
astaxie, line 187: [P2] Serialize whole-Provider deletion with this probe lease

This protects deletion through the resource endpoint, but the normal `DELETE /api/admin/providers/{provider_id}` path still calls `DeleteProvider` without the `codex-image-capability:<provider>` lease. If a slow successful probe is in flight, Provider deletion removes the Provider, resources, and routes; the probe can then create a new active orphan route because `CreateRoute` does not validate or foreign-key the Provider, before capability persistence fails. Make Codex Provider deletion participate in the same lease (or atomically revalidate before route creation) and add a blocked-probe → delete-Provider → release-probe regression test.

backend/internal/server/store_providers.go
```diff
@@ -435,7 +435,11 @@ func (s *GormStore) UpdateProviderResource(id string, patch ProviderResource) (P
 		resource.SensitiveHeaders = sensitive
 	}
 	if patch.Options != nil {
-		resource.Options = patch.Options
+		if isOpenAIAccountResource(resource.ResourceType) {
+			resource.Options = preserveOpenAIAccountProtectedOptions(resource.Options, patch)
```
astaxie, line 471: [P2] Preserve protected options atomically across replicas

This merge preserves keys from the row snapshot loaded at the start of `UpdateProviderResource`, but the mutex is process-local and the method later `Save`s the whole row. In a multi-replica deployment, another instance can persist `image_generation_capability` or `oauth_reauthorization_required` after this snapshot but before the save; this PATCH then overwrites that new server-owned state with the stale value, reintroducing route/publication drift or repeated invalid-token refreshes. Use row locking, optimistic versioning, shared serialization, or column-level updates so preservation is atomic, and add a two-store concurrent PATCH → capability/reauth update regression test.

backend/internal/server/admin_codex_image_capability.go
Mr0bean: Fixed in e114be5. decodeGeneratedImage now accepts decoded payloads only when http.DetectContentType identifies PNG, JPEG, or WebP. TestAdminCodexImageCapabilityRejectsNonImageResult covers a successful JSON response containing Base64 text and verifies that it neither records support nor creates a route. The focused race test and full backend suite pass.

backend/internal/server/admin_provider_resources_routing.go
Mr0bean: Fixed in e114be5. Whole Codex Provider deletion now takes the same codex-image-capability provider lease as capability probing. TestAdminCodexImageCapabilitySerializesProviderDeletionWithProbe blocks a probe, starts Provider deletion, verifies deletion waits, then confirms no orphan route remains. The focused race test and full backend suite pass.

backend/internal/server/store_providers.go
Mr0bean: Fixed in e114be5. Full resource edits and option-map updates now share a per-resource cluster mutation lease across replicas. OAuth refresh keeps network input and output outside that lease, then acquires it to reload, compare, merge, and persist the latest credential and reauthorization state. TestProviderResourceMutationPreservesConcurrentProtectedOptions uses two stores against one SQLite database and verifies that a concurrent account PATCH cannot overwrite capability or reauthorization keys. Focused race, full backend, and go vet checks pass.

backend/internal/server/store_providers.go
```diff
@@ -487,11 +542,22 @@ func (s *GormStore) ensureProviderResourceNameUnique(resourceID, name string) er
 }
 
 func (s *GormStore) UpdateProviderResourceOptions(id string, options map[string]string) (ProviderResource, error) {
+	var updated ProviderResource
+	err := s.withClusterLease(context.Background(), providerResourceMutationLeaseName(id), func(leaseCtx context.Context) error {
```
astaxie, line 571: [P1] Include deletion in the resource mutation lease

The new lease serializes full edits, option updates, and refresh persistence, but `DeleteProviderResource` and `DeleteProvider` do not acquire it. Across replicas, an updater can load this row, deletion can commit, and the updater can then call GORM `Save`; when its UPDATE affects zero rows, GORM falls back to an upsert and recreates the deleted subscription account (or an orphan account after Provider deletion) with its credentials. Make both deletion paths participate in the per-resource mutation lease, or replace these `Save` calls with guarded updates that cannot insert, and add a two-store load → delete → release regression test that asserts the row stays absent.

backend/internal/server/store_providers.go
```diff
@@ -453,13 +475,46 @@ func (s *GormStore) UpdateProviderResource(id string, patch ProviderResource) (P
...
 	}
-	if err := s.db.Save(&resource).Error; err != nil {
+	if err := db.Transaction(func(tx *gorm.DB) error {
+		if err := tx.Save(&resource).Error; err != nil {
+			return err
+		}
+		if !imageBindingChanged || beforeImageCapability != codexImageCapabilitySupported {
+			return nil
+		}
+		var otherAccounts int64
```
astaxie: [P2] Respect route scope when disabling stale image routes

Counting every other subscription account is not sufficient for scoped routes. If an image route is pinned to account A (or resource group G1), replacing A credentials clears its tested capability, but the route remains active merely because unrelated account B exists under the Provider, even when B is outside that route scope. The active route then continues selecting the untested replacement credentials despite the setup contract. Evaluate each matching route against its `provider_resource_id` or `resource_group` and disable routes whose scope no longer contains an eligible tested account; add pinned-account and different-group regression cases.

backend/internal/server/provider_resource_credentials.go
```diff
@@ -84,9 +119,66 @@ func (s *GormStore) mergeOpenAIAccountCredentials(resource *ProviderResource, pa
...
+
+func mergeRefreshedProviderResourceCredentials(
+	original ProviderResourceCredentials,
+	current ProviderResourceCredentials,
+	refreshed ProviderResourceCredentials,
+) ProviderResourceCredentials {
+	current.AuthType = firstNonEmpty(refreshed.AuthType, current.AuthType)
+	current.AccessToken = firstNonEmpty(refreshed.AccessToken, current.AccessToken)
+	current.RefreshToken = firstNonEmpty(refreshed.RefreshToken, current.RefreshToken)
+	current.IDToken = firstNonEmpty(refreshed.IDToken, current.IDToken)
```
astaxie, line 170: [P1] Merge refreshed identity claims when they were not concurrently edited

This copies the rotated access token and ID token but never copies the refreshed `AccountID`, `UserID`, email, organization, or plan type. When the stored values are already non-empty, `mergeOpenAIAccountCredentials` keeps those old values even though the new ID token was decoded into different claims, so later Codex calls can pair the new bearer token with a stale `ChatGPT-Account-ID` and fail authentication. Apply the same compare-with-`original` rule used for scopes to each derived identity field: preserve a genuinely concurrent edit, otherwise persist the refreshed claim; add an old-claims → refreshed-ID-token-with-new-claims test.

backend/internal/server/image_generation.go
```diff
@@ -793,7 +798,12 @@ func decodeGeneratedImage(encoded string) ([]byte, error) {
 	if len(decoded) > maxGeneratedImageBytes {
 		return nil, fmt.Errorf("image result exceeds %d bytes", maxGeneratedImageBytes)
 	}
-	return decoded, nil
+	switch http.DetectContentType(decoded) {
```
astaxie: [P2] Decode the image instead of trusting its signature

`http.DetectContentType` only sniffs leading bytes; a truncated payload containing a PNG/JPEG/WebP signature is classified as an image and still enables the route even though no image decoder can read it. That does not satisfy the PR requirement that the probe return a decodable image, and the same corrupt bytes can later be stored and returned to clients. Validate the complete payload with a format decoder (or equivalent structural validation) after Base64 decoding, and add a correct-magic-but-truncated regression case.

backend/internal/server/admin_providers_http.go
```diff
@@ -1256,6 +1272,16 @@ func (s *Server) validateRoutePolicy(route ModelRoute) error {
...
 	upstreamModel := strings.TrimSpace(route.ProviderModel)
+	if strings.TrimSpace(route.ModelName) == codexImageModelName {
+		provider, ok := s.providerByID(providerID)
+		if !ok || provider.Type != ProviderOpenAICodex {
+			return NewHTTPError(http.StatusBadRequest, "codex_image_provider_required", "The Codex subscription image model must use an OpenAI Codex Provider")
+		}
+		if upstreamModel != codexImageUpstreamModel {
+			return NewHTTPError(http.StatusBadRequest, "codex_image_upstream_model_invalid", "The Codex subscription image route must use gpt-image-2 as its upstream model")
+		}
+		return nil
```
astaxie, line 1287: [P2] Require the capability probe before activating this route

This exception lets the generic routing POST/PATCH path create or re-enable an active `codex-gpt-image-2` route without any account having passed the quota-acknowledged image probe. `TestAdminAllowsCodexImageVirtualModelRouteWithoutProviderInventory` even demonstrates creation with no account at all; once an account exists, direct image requests can use unknown credentials through `filterAndPrioritizeCodexImageRoutes`. That bypasses the PR contract that a real selected account must return a valid image before the Provider route is created or re-enabled. Reject active Codex image routes unless their effective resource scope contains a supported account (while still allowing advanced edits to an already valid route), and cover both create and disabled-to-active PATCH paths.

backend/internal/server/provider_resource_credentials.go
```diff
@@ -226,33 +325,65 @@ func (s *GormStore) RefreshProviderResourceCredentials(ctx context.Context, reso
...
+				current.Options = map[string]string{}
+			}
+			delete(current.Options, openAIAccountReauthorizationRequiredOption)
+			current.Credentials = &refreshed
+			s.mergeOpenAIAccountCredentials(&current, &ProviderResource{Credentials: &refreshed})
+			if strings.TrimSpace(current.APIKey) != "" {
+				current.APIKey = s.encryptSecret(current.APIKey)
+			}
+			current.UpdatedAt = time.Now().UTC()
+			if err := s.db.WithContext(mutationCtx).Save(&current).Error; err != nil {
```
astaxie: [P1] Serialize status and health mutations with refresh persistence

The new mutation lease does not cover `BulkOperateProviderResources` or `SetProviderResourceHealth`, although this path still saves the entire row. Across replicas, refresh can load an active account, an administrator can disable it through the account card's bulk endpoint, and this later `Save` writes the stale active/healthy fields back, silently re-enabling traffic. Runtime health/cooldown updates have the same lost-update window; the invalidation `Save` above is affected too. Make every ProviderResource mutation participate in this lease, or persist only the credential/options columns with guarded updates, and add a two-store refresh -> disable/health update -> release regression test.

astaxie (commented):
Re-reviewed the complete current diff at e114be57. In addition to the four existing unresolved threads, I found two further correctness gaps in the image setup and cross-replica mutation paths.

backend/internal/server/store_providers.go
Mr0bean: Fixed in 051130f. Provider resource edits, option updates, and credential refresh persistence now use guarded column updates instead of GORM Save. A zero-row update returns provider_resource_not_found and cannot fall back to an insert. TestProviderResourceGuardedUpdateDoesNotRecreateDeletedResource uses two stores and covers both resource deletion and whole-Provider deletion, asserting that the account remains absent.

backend/internal/server/provider_resource_credentials.go
Mr0bean: Fixed in 051130f. Credential refresh and reauthorization persistence now update only api_key, credential_blob, options, and updated_at with a guarded update, so concurrent status, health, failure, and cooldown changes are not overwritten. TestProviderCredentialRefreshPreservesConcurrentAvailabilityState uses two stores and covers both bulk disable and health updates during the final refresh persistence window.

backend/internal/server/store_providers.go
Mr0bean: Fixed in 051130f. Stale Codex image routes are now evaluated against their effective provider_resource_id and resource_group scope, and only routes without another active, healthy, supported account in that scope are disabled. TestAdminCodexImageCapabilityReplacementScopesRouteInvalidation covers replaced and unaffected pinned resources, different groups, and the Provider-wide route.

backend/internal/server/provider_resource_credentials.go
Mr0bean: Fixed in 051130f. Refreshed AccountID, UserID, email, organization, and plan claims are now persisted when the stored value still matches the pre-refresh snapshot, while genuine concurrent edits are preserved. TestProviderCredentialRefreshUpdatesDerivedIdentityClaims covers old-to-new claims, and TestProviderCredentialRefreshSurvivesConcurrentMetadataEdit covers the concurrent-edit case.

backend/internal/server/image_generation.go
Mr0bean: Fixed in 051130f. Generated image bytes are now fully decoded as PNG, JPEG, or WebP before the capability can be accepted; WebP decoding uses golang.org/x/image/webp. TestAdminCodexImageCapabilityRejectsTruncatedImageResult supplies a valid PNG signature with truncated data and verifies that support is not recorded and no route is created.

backend/internal/server/admin_providers_http.go
Mr0bean: Fixed in 051130f. Generic Codex image route POST and PATCH validation now requires an active, healthy, supported subscription account inside the route effective pinned-resource, group, or Provider scope before an active route is accepted. Disabled preconfiguration remains allowed. TestAdminCodexImageVirtualModelRouteRequiresSupportedProviderInventory covers active creation, disabled creation, rejected disabled-to-active PATCH, and successful activation after capability support is present.

astaxie (approved):
Re-reviewed the complete current diff at `051130fe`. The latest commit addresses the six current-head findings: guarded resource updates cannot recreate deleted rows; Codex image route invalidation respects pinned resource and group scope; refreshed identity claims persist without overwriting concurrent operator edits; image capability probes fully decode PNG/JPEG/WebP results; active routes require supported inventory; and refresh persistence preserves concurrent availability state. The full backend suite, `go vet ./...`, frontend typecheck and production build, all 122 repository tool tests, translation/environment/source-line gates, and `git diff --check` pass. Required CI is green, all review threads are resolved, and GitHub reports a clean merge state.

Issue #160, opened 2026-08-08
feat(agent): add an A2A agent gateway and runtime governance


astaxie:
## Background

As enterprises package model calls into reusable agents, inconsistent agent sources and protocols, duplicate client integrations, fragmented authorization, untraceable long-running tasks, runaway sub-agent loops, and difficult cost attribution emerge. Proxying LLM requests alone cannot govern discovery, invocation, and task lifecycles between agents.

## Goal

Build an A2A agent gateway with unified registration, discovery, invocation, streaming, task management, access control, runtime budgets, and end-to-end audit. Callers connect only to TokenHub; the gateway handles protocol adaptation, identity, and policy enforcement and links an agent's internal model, MCP, and sub-agent calls into one trace.

### Agent registration and discovery

- Provide an agent registry, Admin API, and console that can import name, description, version, capabilities, skills, input/output modes, and security declarations from an upstream Agent Card.
- Allow administrators to review and choose externally exposed Card fields and skills. Support resynchronization, diff preview, approval, and rollback for upstream changes.
- Expose a stable proxy URL and standard Agent Card discovery endpoint for each agent while hiding its real upstream address and internal credentials.
- Support configuration-file and database registration sources with defined rules for name conflicts, hot reload, and multi-instance consistency.
- Connect native A2A agents and common hosted/framework agents through adapters. The first release must support generic A2A upstreams; other types must not compromise the core protocol model.

### Protocol and invocation

- Support A2A JSON-RPC `message/send`, `message/stream`, and task lifecycle methods including `tasks/get`, `tasks/list`, `tasks/cancel`, and `tasks/resubscribe`.
- Support streaming state and artifact events with correct client cancellation, reconnection, timeout, and backpressure behavior.
- Support A2A 0.3 and 1.0, allowing each agent to expose a fixed protocol version and returning explicit errors for unsupported versions.
- Provide an OpenAI-compatible invocation bridge so existing SDKs can call agents through an explicit namespace. Native A2A capabilities and task semantics remain authoritative on A2A endpoints.
- Support skill-based routing while preserving context, task, artifact, and structured metadata. Do not treat an agent as a plain text model and lose protocol state.

### Authorization and identity

- Scope agent discovery and invocation permissions by organization, team, project, API key, end user, and agent, with access groups. Combine multiple scopes by least privilege.
- Exclude unauthorized agents from lists and Card discovery; direct invocation must return the same non-disclosing denial.
[...]

PR #196, opened 2026-08-11
feat(a2a): add 1.0 agent gateway

legendtkl:
## Summary

Adds a disabled-by-default A2A 1.0 Agent Gateway to TokenHub. The gateway lets authenticated callers discover and invoke reviewed upstream Agents through TokenHub while preserving centralized access control, task isolation, routing, budgets, auditability, and deployment controls. This implements the agreed scope of #160 using A2A 1.0 exclusively; A2A 0.3 is intentionally unsupported.

## Related Issue

Relates to #160.

## Changes

- Add A2A 1.0 JSON-RPC and SSE gateway support for `SendMessage`, `SendStreamingMessage`, `GetTask`, `ListTasks`, `CancelTask`, and `SubscribeToTask`, plus authenticated Agent Card discovery.
- Add the `agent/<slug>` OpenAI Responses bridge for non-streaming and streaming calls.
- Add database- and catalog-managed Agent registration, encrypted instance credentials, revisions and rollback, weighted priority routing, task affinity, passive health cooldowns, and atomic cross-replica concurrency leases.
- Add default-deny Agent access bindings for global, team, project, API key, end user, calling Agent, and access-group scopes, including skill filtering and deny precedence.
- Add sanitized gateway task identities, API-key and end-user task isolation, delegation tokens, loop/depth protection, execution graphs, runtime/call/token/cost/concurrency budgets, and MCP/model usage accounting.
- Add SSRF and DNS-rebinding protections, redirect blocking, sensitive-header filtering, bounded upstream responses, and request-level execution finalization.
- Add the Agent Gateway admin page, execution observability, English/Simplified Chinese/Japanese UI copy and documentation, deployment configuration, release packaging, and SDK process E2E fixtures.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor or maintenance
- [x] Documentation

## Verification

- `node --test tools/*.test.mjs` — 116 tests passed
- `git diff --check`
- Real backend process plus independent A2A upstream process: all six A2A 1.0 methods, task and artifact SSE events, explicit 0.3 rejection, and streaming/non-streaming Responses bridge passed.
- PostgreSQL 16 two-replica E2E: shared task affinity and atomic `max_concurrency=1` admission passed.
- Browser E2E: Agent registration, access binding, revision creation, rollback, execution observability, responsive layout, and clean browser console passed.
- Native/container installer tests and all three rendered Docker Compose configurations passed; the container entrypoint test was run in its Linux target environment.

- Existing OpenAI-compatible endpoints remain compatible. New public surfaces are `/a2a/<slug>` and the `agent/<slug>` Responses model namespace.
- The gateway requires `A2A-Version: 1.0`; missing versions and A2A 0.3 are rejected intentionally.
- The feature defaults to disabled through `TOKENHUB_A2A_ENABLED=false`. Rollback is to disable the flag; registry, task, and execution data remain stored.
- Startup migration adds Agent registry, instance lease, access-control, task, execution graph, revision, and usage tables.
- Upstream credentials are encrypted with the TokenHub secret key and are never exposed through public cards or admin execution payloads.
- Production upstreams require HTTPS and public network destinations. `TOKENHUB_A2A_ALLOW_PRIVATE_UPSTREAMS=true` is restricted to controlled development environments.
- A reviewed `data/agent-catalog.yaml` is added to source, container, and native release artifacts.

## Checklist


astaxie (commented):
Reviewed the complete current diff at d96c682. This head is not mergeable yet:

- Required Backend/CI fails because parseSSEDataForTest is unused in backend/internal/server/agent_e2e_test.go:819.
- The runtime A2A dial guard in agent_gateway.go only rejects private/loopback/unspecified/link-local addresses. It still permits CGNAT and other special-use destinations (for example 100.100.100.200), after which agentCredentialTransport attaches static instance credentials. Reuse a complete special-use classifier at validation and dial time, and add credential-bearing dial regressions.
- Issue #160 requires fixed A2A 0.3/1.0 compatibility, but this head only advertises/accepts 1.0 and explicitly rejects 0.3. Either implement the acceptance criterion with compatibility tests or update the issue/scope before closing it.
- Direct JSON-RPC calls distinguish a missing Agent (TASK_NOT_FOUND) from an existing but unauthorized Agent (UNAUTHORIZED), creating the existence oracle that #160 prohibits. Return the same non-disclosing response as Card discovery.
- Only input text is evaluated by guardrails. Non-streaming output, streaming status messages, and artifact parts are persisted/returned without the output/artifact content-policy hook required by #160.
- This adds tracked data/agent-catalog.yaml, while AGENTS.md still says only data/model-catalog.yaml is tracked and other runtime data is ignored. Synchronize the repository guide.

No approval while these blockers and the failing required check remain.

legendtkl:
Addressed the review findings in `9b2a81d`:

- Removed the unused `parseSSEDataForTest` helper/import that broke the required Backend check.
- Replaced the narrow private-IP check with a shared conservative special-use classifier at both registration validation and runtime dial time. It covers CGNAT (`100.64.0.0/10`), documentation/benchmark/reserved/multicast ranges, and IPv6 special-use space; a credential-bearing regression verifies `100.100.100.200` is rejected before the connection is made.
- Unified missing, unauthenticated, and unauthorized direct JSON-RPC Agent admission responses to the same non-disclosing `TASK_NOT_FOUND` response, with exact-body regressions.
- Added output guardrails for non-streaming messages/tasks, task status messages, streaming status events, task artifacts, and streaming artifact events before return and persistence. Direct A2A errors expose `CONTENT_POLICY_VIOLATION`; the Responses bridge maps policy blocks to `403 guardrail_blocked` or a streaming `guardrail_blocked` error event instead of an upstream failure.
- Updated `AGENTS.md` to document both tracked catalog sources.

A2A 0.3 is intentionally not implemented, per the confirmed product scope that this gateway uses A2A 1.0 exclusively. I changed the PR relationship from `Closes #160` to `Relates to #160`, so this PR will not incorrectly close the broader issue while its 0.3 acceptance criterion remains unresolved.

Validation rerun: focused A2A E2E, `go test ./...`, `go vet ./...`, all repository gates (116 tests), full translation co-change check, frontend typecheck/build, and `git diff --check`. The mandatory independent read-only review found one error-classification issue, which was fixed and then re-reviewed with no remaining findings.

Please re-review the updated head when convenient.

legendtkl:
Follow-up: merged the latest `origin/main` into the PR head in `979388c` and resolved the four overlapping files while preserving both A2A and the newly merged durable Responses functionality. Agent Responses continue through the A2A bridge; ordinary `background=true` Responses continue through the durable queue; `background=true` with `agent/<slug>` now fails explicitly with `agent_background_not_supported` and has an E2E regression.

Post-merge validation passed: focused A2A E2E, `go test ./...`, `go vet ./...`, repository gates (122/122), full doc/UI/env/source-line checks, frontend `npm ci`, typecheck and production build, and `git diff --check`. A final independent read-only review of the conflict resolutions found no issues.

backend/internal/server/agent_gateway.go
```diff
@@ -0,0 +1,933 @@
...
+	case *a2a.TaskStatusUpdateEvent:
+		item.TaskID = a2a.TaskID(task.ID)
+		task.State = string(item.Status.State)
+		if item.Status.Message != nil {
+			item.Status.Message.TaskID = a2a.TaskID(task.ID)
+		}
+	case *a2a.TaskArtifactUpdateEvent:
+		item.TaskID = a2a.TaskID(task.ID)
+	}
+	if task.SnapshotJSON == "" {
```
astaxie: [P2] Fold streamed updates into the task snapshot

SnapshotJSON is only initialized when empty. Once an initial Task snapshot exists, TaskStatusUpdateEvent changes only task.State and TaskArtifactUpdateEvent changes no snapshot fields; ListTasks later unmarshals SnapshotJSON rather than replaying the event rows. I reproduced a SUBMITTED task receiving a COMPLETED status event and ListTasks still seeing SUBMITTED with no status message (streamed artifacts are likewise omitted). Please merge each streamed event into the persisted snapshot, or rebuild the listed task from its events.

backend/internal/server/gateway_http.go
```diff
@@ -156,6 +156,19 @@ func (s *Server) handleResponses(w http.ResponseWriter, r *http.Request) {
...
+			return
+		}
+		if req.Background {
+			writeError(w, r, NewHTTPError(http.StatusBadRequest, "agent_background_not_supported", "Background Responses are not supported for Agent models"))
+			return
+		}
+		s.handleAgentResponses(w, r, project, key, req, slug)
+		return
+	}
 	if req.Background {
```
astaxie, line 181: [P1] Preserve delegation attribution for background jobs

A caller authenticated with a thd_ token can submit a normal model with background=true. Admission increments model_calls, but responseJobRequestHeaders deliberately strips authorization and the durable envelope stores no trusted execution claims, so the worker never calls recordDelegatedModelUsage when it settles the job. The resulting tokens, cost, and AgentUsageRecord are missing from the root Agent execution. Please either reject background requests authenticated by delegation tokens or persist trusted delegation attribution and settle it on job completion.

backend/internal/server/agent_http.go
```diff
@@ -0,0 +1,891 @@
...
+			_ = s.store.FinishAgentExecution(claims.ExecutionID, "loop_detected")
+			return agentInvocation{}, ErrInvalidAPIKey
+		}
+		project, found := s.store.GetProject(claims.ProjectID)
+		if !found || project.Status != StatusActive {
+			return agentInvocation{}, ErrInvalidAPIKey
+		}
+		var key APIKey
+		for _, candidate := range s.store.ListAPIKeys() {
+			if candidate.ID == claims.APIKeyID && candidate.ProjectID == project.ID && candidate.Status == StatusActive {
```
astaxie: [P1] Reject expired API keys on direct delegated calls

This delegated A2A path accepts any active key but never checks ExpiresAt. I reproduced this by minting a thd_ token, expiring its underlying API key, and calling authenticateAgentInvocation: authentication still succeeds, while authenticateAgentDelegation correctly rejects the same key. A token minted immediately before key expiry can therefore keep invoking Agents for the remainder of its five-minute lifetime. Please apply the same ExpiresAt check here and add a regression test.

backend/internal/server/agent_gateway.go
```diff
@@ -0,0 +1,933 @@
...
+		}
+	case *a2a.Message:
+		if task.ID != "" && event.TaskID != "" {
+			event.TaskID = a2a.TaskID(task.ID)
+		}
+	}
+	if err != nil {
+		return nil, a2a.NewError(a2a.ErrInternalError, "Task state could not be persisted")
+	}
+	status = "completed"
```
astaxie: [P1] Keep the execution alive for non-terminal tasks

This unconditionally completes the root execution after SendMessage returns, even when the result is a Task in SUBMITTED or WORKING state. An upstream Agent that continues asynchronously then finds its freshly issued delegation token rejected with agent_execution_not_running, and later GetTask/SubscribeToTask calls create unrelated root executions instead of resuming AgentTask.ExecutionID. Please keep the original execution open until the task reaches a terminal state (and resume that execution for follow-up task operations) so long-running work retains one trace and budget.

backend/internal/server/agent_responses.go
```diff
@@ -0,0 +1,190 @@
...
+	"strings"
+	"time"
+
+	"github.com/a2aproject/a2a-go/v2/a2a"
+)
+
+func (s *Server) handleAgentResponses(w http.ResponseWriter, r *http.Request, project Project, key APIKey, request ResponsesRequest, slug string) {
+	agent, found, err := s.store.GetAgentBySlug(slug)
+	if err != nil || !found || agent.Status != StatusActive {
+		writeError(w, r, NewHTTPError(http.StatusNotFound, "agent_not_found", "Agent was not found"))
```
astaxie: [P2] Make Responses authorization failures non-enumerable

The Responses bridge returns 404 agent_not_found for an unknown slug, but line 30 returns 403 agent_access_denied for an existing Agent that this same API key cannot invoke. That restores an Agent-registry existence oracle on the direct agent/<slug> invocation path, despite the JSON-RPC path now deliberately making missing and unauthorized Agents indistinguishable. Please use the same status/code/body for both cases.

backend/internal/server/agent_gateway.go
legendtkl: Fixed in db021ef. Streamed status/message and artifact updates are now folded into SnapshotJSON before persistence; artifact chunks merge by artifact ID with append semantics. TestAgentStreamEventsUpdateListTaskSnapshot verifies ListTasks returns the completed status, sanitized task identity, and combined artifact content.

backend/internal/server/gateway_http.go
legendtkl: Fixed in db021ef. Background Responses authenticated by an Agent delegation token are now rejected before model-call budget admission with delegated_background_not_supported, so no durable envelope can lose attribution. The regression verifies model_calls remains zero and no ResponseJob is created.

backend/internal/server/agent_gateway.go
legendtkl: Fixed in db021ef. AgentTask now retains its execution step; non-terminal task states leave the execution and step running, and direct follow-up task operations resume the same execution, trace, deadline, chain, and budget. A terminal follow-up finishes the original execution. TestNonTerminalAgentTaskKeepsAndResumesExecution verifies one execution and a stable trace across SendMessage and GetTask.

backend/internal/server/agent_responses.go
legendtkl: Fixed in db021ef. The Responses bridge now returns the same 404 agent_not_found error envelope for a missing Agent, an unauthorized existing Agent, and disallowed end-user identity, matching the non-enumerable JSON-RPC/Card behavior. TestAgentResponsesAuthorizationIsNonEnumerable covers the missing-versus-denied case.

backend/internal/server/agent_http.go
legendtkl: Fixed in db021ef. Direct delegated Agent authentication now applies the same active-and-unexpired API key check as ordinary delegation authentication. TestDirectDelegatedAgentInvocationRejectsExpiredAPIKey covers a freshly minted token whose backing key expires before the nested call.

astaxie (commented):
Re-reviewed the complete current diff at 51b24f5d. The three author replies are verified: streamed task state and artifacts are folded into the persisted snapshot, delegated background Responses are rejected before budget admission and job creation, and direct delegation now rejects expired backing API keys. The full backend suite, go vet, frontend tests/typecheck/build, all 122 repository tests and gates, three Compose renders, native installer tests, git diff check, and all required GitHub checks pass. I found no remaining code blocker. Approval still requires the PR body to preserve the current template: add the required English title/body checklist item and record explicit pass results for the first four Verification commands. The Linux-only container entrypoint test and SDK process smoke were not rerun locally; the corresponding GitHub Deployment and backend checks are green.

astaxie (commented):
Re-reviewed the complete current diff at head `deff10fc36b148efd471fb35e1c7915f1ad3a89e`. The latest task-artifact fix is correct and its focused tests pass: Responses now prefer and join all text-bearing artifacts before falling back to status/history.

One additional current-diff blocker remains in `frontend/features/admin/views/agent-gateway.tsx:257-259,312,319`. These rows construct user-visible counts and metric summaries by placing translated fragments around dynamic values, and render USD as a literal `$` plus an unformatted/decimal-formatted number. `AGENTS.md` requires complete translated templates for dynamic sentences and `languageLocale()` plus `Intl` currency formatting. Please replace the count/summary fragments with complete locale templates and format USD with `Intl.NumberFormat(languageLocale(), { style: "currency", currency: "USD" })`, with focused locale coverage.

This head also currently has merge conflicts and no GitHub CI check suite, so it cannot be approved.

PR #201, opened 2026-08-12
fix(codex): converge OAuth request fingerprints

astaxie:
## Summary

Converge Codex OAuth device and session fingerprints at the account-resource level so shared subscription accounts present stable identifiers upstream. This adapts the behavior  to TokenHub's Provider Resource and Next.js architecture.

## Related Issue

N/A.

## Changes

- Add `off`, `device`, `session`, and `full` fingerprint modes, with `session` as the default and `off` as the passthrough rollback.
- Rewrite Codex protocol headers and `client_metadata` from one precomputed ID set, preserving unrelated metadata and keeping retries consistent.
- Add account-resource UI configuration, payload normalization, focused Go and frontend tests, and synchronized English, Simplified Chinese, and Japanese documentation.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor or maintenance
- [ ] Documentation

## Verification

- `gofmt -w` on changed Go files
- `go test ./...`
- `go vet ./...`
- `npm ci`
- `npm run typecheck`
- `npm run build`
- `git diff --check origin/main...HEAD`

`npm ci` also reported four existing high-severity audit findings; this PR does not change the frontend dependency lockfile.

No schema, public API, environment, or deployment changes. Existing and new Codex subscription resources default to `session` convergence; administrators can set `codex_fingerprint_mode=off` to restore passthrough behavior. Stable IDs are derived from the Provider Resource ID and do not include OAuth credentials.

## Checklist


backend/internal/server/codex_protocol.go
```diff
@@ -10,6 +10,7 @@ import (
 
 var codexRequestHeaderAllowlist = []string{
 	"session-id",
+	"session_id",
```
astaxie, line 13: `session_id` is now accepted as a Codex session header, but `codexSessionIdentifier` only checks `session-id`. In the default `session` mode, clients that send only this newly supported underscore form therefore produce an empty `clientSessionID`, so `resolveCodexFingerprintIDs` falls back to `threadID = sessionID`; distinct client sessions are collapsed onto the same account-level thread (effectively `full` mode). I reproduced this with two requests carrying different `session_id` values: both generated the same thread ID. Please include `headers.Get("session_id")` in the canonical session lookup and add a regression case.

backend/internal/server/codex_fingerprint.go
```diff
@@ -0,0 +1,194 @@
...
+		return
+	}
+	headers.Set("x-codex-installation-id", ids.installationID)
+	if ids.mode == codexFingerprintDevice {
+		rewriteCodexTurnMetadataHeader(headers, map[string]any{
+			"installation_id": ids.installationID,
+		})
+		return
+	}
+	headers.Set("x-codex-window-id", ids.windowID)
```
astaxie, line 112: `x-codex-parent-thread-id` is still copied verbatim by `applyCodexRequestHeaders`, even in `session`/`full` mode. This is a Codex thread identity (used for child-thread lineage), so leaving it untouched exposes the original client thread and makes the rewritten `thread-id`/turn metadata internally inconsistent; in `full` mode, child requests are not actually converged onto the account thread. Please derive/rewrite (or deliberately remove) this header from the same precomputed IDs and cover child-thread requests in the mode tests.

backend/internal/server/codex_fingerprint.go
```diff
@@ -0,0 +1,194 @@
...
+	if err != nil {
+		turnID = uuid.New()
+	}
+	ids.turnID = turnID.String()
+	ids.windowID = ids.threadID + ":0"
+	ids.turnStartedAtUnixMS = time.Now().UnixMilli()
+	return ids
+}
+
+func prepareCodexFingerprintRequest(provider Provider, incoming http.Header, request *ResponsesRequest) *codexFingerprintIDs {
```
astaxie, line 88: This policy is only applied by `OpenResponses`; `CompactWithHeaders` still forwards the original Codex identity headers directly via `applyCodexRequestHeaders` and never calls `resolveCodexFingerprintIDs`/`applyCodexFingerprintHeaders`. As a result, `/v1/responses/compact` bypasses every non-`off` account setting and exposes the client installation/session/thread IDs upstream, even though fingerprint convergence is configured at the account-resource level. Please precompute/apply the same mode-specific IDs in the compact retry path as well (while continuing to omit unsupported `client_metadata`).

backend/internal/server/codex_fingerprint.go
```diff
@@ -0,0 +1,197 @@
...
+		request.raw = make(map[string]json.RawMessage)
+	}
+	metadata := map[string]any{}
+	if raw, ok := request.raw["client_metadata"]; ok {
+		_ = json.Unmarshal(raw, &metadata)
+	}
+	if metadata == nil {
+		metadata = map[string]any{}
+	}
+	metadata["x-codex-installation-id"] = ids.installationID
```
astaxie, line 155: The follow-up removes `x-codex-parent-thread-id` from the HTTP headers, but the same identity is still preserved verbatim when it appears in `client_metadata`. In `session`/`full` mode this still leaks the pre-rewrite parent namespace and contradicts the otherwise converged `thread_id`. I reproduced it on `54b4ec3`: a body with `client_metadata.session_id` plus `client_metadata["x-codex-parent-thread-id"] = "client-parent"` is forwarded with `client-parent` unchanged. Please delete or consistently remap this client-metadata field for non-device modes and add it to the regression test.

backend/internal/server/codex_fingerprint.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+func applyCodexFingerprintClientMetadata(request *ResponsesRequest, ids *codexFingerprintIDs) {
+	if request == nil || ids == nil {
+		return
+	}
+	if request.raw == nil {
+		request.raw = make(map[string]json.RawMessage)
+	}
+	metadata := map[string]any{}
```
astaxie, line 148: The JSON rewrite does not actually preserve all unrelated metadata: decoding into `map[string]any` converts every number to `float64`. On `54b4ec3` I reproduced a valid `client_metadata.unrelated` value of `9007199254740993` being forwarded as `9007199254740992`; the same risk exists in both turn-metadata rewrite helpers. Please retain untouched fields as `json.RawMessage` or decode with `UseNumber`, and add a large-integer regression case.

backend/internal/server/codex_fingerprint.go
```diff
@@ -0,0 +1,197 @@
...
+	if ids.mode == codexFingerprintDevice {
+		rewriteEmbeddedCodexTurnMetadata(metadata, map[string]any{
+			"installation_id": ids.installationID,
+		})
+	} else {
+		metadata["session_id"] = ids.sessionID
+		metadata["thread_id"] = ids.threadID
+		metadata["turn_id"] = ids.turnID
+		metadata["x-codex-window-id"] = ids.windowID
+		rewriteEmbeddedCodexTurnMetadata(metadata, codexFingerprintTurnMetadata(ids))
```
astaxie: There is a second lineage bypass beyond the standalone field above: current Codex clients put `parent_thread_id`, `forked_from_thread_id`, and `parent_turn_id` in `client_metadata` / embedded `x-codex-turn-metadata` (see OpenAI Codex `responses_metadata.rs`). `mergeCodexFingerprintFields` preserves all of them, so non-device modes still mix rewritten current IDs with original parent/fork IDs; Compact has the same issue through the turn-metadata header. Please include these canonical lineage keys in the delete/remap policy and exercise Responses plus Compact child-thread metadata.

backend/internal/server/codex_fingerprint.go
```diff
@@ -0,0 +1,233 @@
...
+func prepareCodexFingerprintRequest(provider Provider, incoming http.Header, request *ResponsesRequest) *codexFingerprintIDs {
+	if request == nil {
+		return nil
+	}
+	clientSessionID, _ := codexSessionIdentifier(incoming, *request)
+	ids := resolveCodexFingerprintIDs(provider, clientSessionID)
+	if ids != nil && request.raw != nil {
+		request.raw = cloneRawJSON(request.raw, 1)
+	}
+	applyCodexFingerprintClientMetadata(request, ids)
```
astaxie, line 97: `prompt_cache_key` is left in the original client session namespace while this function rewrites the other session identifiers. With the default `session` mode and a body containing `"prompt_cache_key":"client-session"`, upstream receives the account-derived `session-id` and `client_metadata.session_id`, but still receives the original `prompt_cache_key`. TokenHub itself treats this field as session-scoped (`session_affinity_key.go`), and current Codex defaults it to its session ID, so shared clients remain distinguishable upstream and cache routing stays split by their original sessions. Please rewrite a present `prompt_cache_key` to the precomputed `ids.sessionID` in `session`/`full` modes (preserving it for `off`/`device`) and cover retry consistency.

PR #203, opened 2026-08-12
feat(provider): add Kronk local inference support

### LINKED ISSUE #170: feat(provider): add Kronk local inference support
## Background

[Kronk](https://github.com/ardanlabs/kronk) is a Go-based open-source local model inference engine that integrates llama.cpp and whisper.cpp and provides a standalone Model Server. Its HTTP API supports Chat Completions, Responses, Anthropic Messages, Embeddings, Reranking, audio transcription, and other capabilities.

TokenHub can currently connect to some local inference services through the generic OpenAI-compatible provider, but it does not recognize Kronk as a built-in provider. Administrators must determine the base URL, protocol type, authentication method, model names, and probe configuration themselves, and TokenHub cannot discover models directly from the Kronk service.

## Goal

Add a standalone Kronk Model Server as a built-in local inference provider so administrators can:

1. Select Kronk in TokenHub and configure its service address and optional access token.
2. Verify connectivity to Kronk and discover locally installed models.
3. Import discovered models into TokenHub and configure standard model mappings and routes.
4. Invoke Kronk through TokenHub's OpenAI-compatible API while retaining existing authentication, project isolation, quotas, audit, routing, and failover.

### Provider catalog and configuration

- Add `Kronk` to the built-in provider catalog and classify it as a local inference service.
- Default the base URL to `http://127.0.0.1:11435/v1` while allowing full customization.
- Support credential-free access when Kronk authentication is disabled.
- Support forwarding a Kronk application token as `Authorization: Bearer <token>`.
- Never forward the TokenHub client's Authorization header or expose the Kronk token in logs, audit records, or error responses.
- Provide a Kronk name, description, documentation link, and sensible defaults in the console, with English, Simplified Chinese, and Japanese copy.

### Model discovery and import

- Retrieve currently available local models from Kronk through `GET /v1/models`.
- Preserve complete model IDs returned by Kronk, including `/`, `:`, and quantization suffixes, without lossy normalization.
- Allow administrators to select and import models as TokenHub models/routes and map external standard model names to Kronk model IDs.
- Make repeated synchronization idempotent so it creates no duplicate models or routes.
- Do not silently delete models removed from Kronk; mark them unavailable in TokenHub or prompt an administrator to resolve them.
- A model-discovery failure must not overwrite existing routes or model configuration.

### Inference capabilities

The first release must support at least:

- `POST /v1/chat/completions`: buffered and SSE streaming responses.
- `POST /v1/responses`: buffered and SSE streaming responses.
- `POST /v1/embeddings`.
- Pass through tool calls, structured output, vision input, and reasoning content according to Kronk's actual capabilities. Unsupported parameters must produce explicit errors rather than silently changing semantics.
- Record prompt, completion, and total token usage returned by Kronk correctly. When upstream usage is absent, use TokenHub's existing missing-usage policy.
- Normalize Kronk's top-level `code`/`message` error format into TokenHub errors while preserving the correct HTTP status and retry classification.
- Promptly terminate requests on client cancellation, TokenHub timeout, or upstream disconnect without leaving a stream running.

Kronk's native `/v1/messages`, Reranking, Tokenize, and Audio Transcriptions capabilities can be added separately after TokenHub exposes the corresponding public capability and do not block the first release.

### Health checks

- Use Kronk's `/v1/liveness` and `/v1/readiness` endpoints to check service-process reachability.
- Do not rely on readiness alone for model-level availability, because it does not validate model files, inference libraries, devices, or GPU memory. Report resource state separately through model discovery or a lightweight inference probe.
- Distinguish authentication failure, network unreachability, service not ready, model not found, insufficient resources, and inference timeout for administrator diagnostics.
- Health checks must not call Kronk management endpoints or trigger model downloads.

## Implementation guidance

Kronk's core inference endpoints are OpenAI-compatible. Prefer reusing existing OpenAI-compatible request, SSE, usage, and error-handling capabilities, adding Kronk-specific configuration or adaptation only for:

- Built-in provider metadata and default address.
- Model discovery through `GET /v1/models`.
- Kronk health-check endpoints and error format.
- Kronk-specific model IDs and capability metadata.
- Unauthenticated and Bearer Token authentication modes.

If a `kronk` provider type is added, declare its capabilities explicitly in the Adapter Registry and reuse the OpenAI-compatible adapter core instead of copying the entire HTTP invocation implementation.

## Deployment and security

- TokenHub connects only to an external Kronk Model Server; it does not embed the Kronk SDK or llama.cpp in the TokenHub backend process.
- This issue does not require Kronk to start by default in TokenHub Compose. Kronk deployment is responsible for CPU, Metal, CUDA, ROCm, Vulkan, and other runtime/image differences.
- Document that `127.0.0.1` inside a container refers to the TokenHub container itself. A reachable address must be configured when Kronk runs on the host or in another container.
- Kronk listens on plaintext local HTTP by default. Remote deployments should use a trusted private network or TLS reverse proxy and an appropriate Kronk authorization mode.
- TokenHub uses only inference and model-discovery endpoints. It does not proxy Kronk model downloads, catalogs, security management, debugging, pprof, or the management UI.

## Non-goals

- TokenHub will not install or upgrade Kronk or the native llama.cpp/whisper.cpp libraries.
- TokenHub will not download, delete, load, or unload GGUF models.
- This work will not implement GPU scheduling, GPU-memory planning, model pools, or process lifecycle management.
- TokenHub will not expose Kronk management APIs, MCP services, or debug ports to TokenHub API consumers.
- Existing TokenHub `/v1` public API compatibility will not change.

## Acceptance criteria

- [ ] Kronk can be selected in the provider catalog, with the default base URL and documentation link populated automatically.
- [ ] Both credential-free and Bearer Token-protected Kronk services are supported.
- [ ] Local models can be discovered from `GET /v1/models` and imported idempotently with complete model IDs preserved.
- [ ] Standard model mappings and routes can be created for imported models.
- [ ] Buffered and streaming Chat Completions requests can invoke Kronk through TokenHub.
- [ ] Buffered and streaming Responses requests can invoke Kronk through TokenHub.
- [ ] Embeddings requests work and return a compatible format.
- [ ] Kronk usage, error status, and retry semantics are recorded and normalized correctly.
- [ ] Health checks distinguish service health from model availability.
- [ ] The Kronk token never appears in logs, audit payloads, exported data, or client responses.
- [ ] An in-process fake Kronk HTTP server covers model discovery, authentication, streaming, error normalization, disconnect, and timeout tests.
- [ ] English, Simplified Chinese, and Japanese administrator and deployment documentation is updated.
- [ ] `go test ./...`, `go vet ./...`, frontend typecheck/build, repository gates, and `git diff --check` all pass.

astaxie:
## Summary

Add Kronk as a built-in local inference provider so administrators can discover installed models, import exact model IDs, configure standard model routes, and use Kronk through TokenHub's existing OpenAI-compatible gateway controls.

## Related Issue

Closes #170.

## Changes

- Add a dedicated `kronk` adapter and provider catalog entry with the default `http://127.0.0.1:11435/v1` endpoint, optional Bearer application token, and explicit saved-token clearing.
- Reuse the OpenAI-compatible core for Chat Completions, Responses, Embeddings, SSE forwarding, usage accounting, cancellation, and timeout handling.
- Add layered liveness, readiness, and model availability checks with structured diagnostics, plus Kronk-specific error normalization, resource-exhaustion classification, and credential redaction.
- Add live model discovery and idempotent synchronization that preserves exact Kronk model IDs, disables missing discovered inventory without deleting routes, and leaves configuration unchanged after discovery failures.
- Add administrator UI support and synchronized English, Simplified Chinese, and Japanese administrator/deployment documentation.
- Add in-process fake Kronk coverage for authentication, discovery/import, health, inference, streaming, usage, errors, cancellation, timeout, interrupted streams, and token clearing.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor or maintenance
- [x] Documentation

## Verification

- `go test ./...` — passed
- `go vet ./...` — passed
- `npm run typecheck` — passed
- `npm run build` — passed
- `node --test tools/*.test.mjs` — passed (122 tests)
- `node tools/check-doc-translations.mjs --base origin/main --head HEAD` — passed
- `node tools/check-ui-translations.mjs` — passed
- `node tools/check-env-contract.mjs` — passed
- `node tools/check-source-lines.mjs` — passed; reported one existing baseline that can be tightened
- `docker compose --env-file deploy/.env.example -f deploy/docker-compose.yml config` — passed
- `git diff --check` — passed
- SDK smoke tests were not run because no compatible live Kronk-backed TokenHub stack was configured; equivalent adapter behavior is covered by in-process fake upstream tests.

The existing public `/v1` contracts are unchanged. Kronk is external to TokenHub and is not added to Compose. Application tokens are optional, are used only for Kronk Bearer authentication, and are redacted from normalized upstream errors. Container deployments must use a Kronk address reachable from the TokenHub container; remote connections should use a trusted private network or TLS reverse proxy. Rollback is the normal application rollback; imported inventory and routes remain in persistence and can be disabled or removed by an administrator.

## Checklist

- [x] Environment variable changes are synchronized across examples, Compose, `start.sh`, and deployment documentation where applicable. No environment variables were added or changed.
- [x] `data/model-catalog.yaml` remains tracked and catalog changes were reviewed where applicable. The model catalog was not changed; the reviewed provider catalog was updated.


backend/internal/server/provider_kronk_test.go
```diff
@@ -0,0 +1,423 @@
...
+	}
+	if providerErrorDisposition(err) != ProviderErrorTransientSame {
+		t.Fatalf("unexpected retry classification: %s", providerErrorDisposition(err))
+	}
+}
+
+func TestKronkNonStreamingTimeoutIsClassifiedAndStopsUpstream(t *testing.T) {
+	upstreamCanceled := make(chan struct{})
+	upstream := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		<-r.Context().Done()
```
astaxie: [P1] Make the timeout test release its server handler

This handler waits forever if r.Context() is not observed as canceled. When the assertion at line 271 fails, t.Fatal runs deferred upstream.Close(), which then waits on this still-blocked connection, so the test never returns. This is the current Backend CI failure (the package hits its 10-minute timeout), and I reproduced it locally with -count=20. Use a bounded select/test-controlled release plus CloseClientConnections (or otherwise ensure cleanup unblocks the handler) so a failed cancellation assertion reports promptly instead of hanging the entire suite.

frontend/features/admin/views/provider-editor.tsx
```diff
@@ -273,16 +274,16 @@ export function ProviderUpsertModal({
         cancelled = true;
       };
     }
-    if (catalogID === "custom") {
+    if (catalogID === "custom" || catalogID === "kronk") {
```
astaxie, line 278: [P1] Let Kronk discovery bypass the standard-model filter

This loads arbitrary local Kronk IDs, but the models memo below only bypasses standardModels matching for catalogID === "custom". Kronk therefore falls through to canonical-name matching, so IDs such as org/model:Q4_K_M.gguf (the exact shape covered by the backend test) disappear unless they coincidentally exist in TokenHub's standard catalog. The normal wizard then shows nothing selectable and cannot import the discovered inventory, contrary to issue #170. Treat kronk like custom in that filter and add a UI regression.

backend/internal/server/provider_kronk.go
```diff
@@ -0,0 +1,261 @@
...
+			return ProviderCatalogEntry{}, NewHTTPError(http.StatusBadRequest, "provider_type_invalid", "Provider is not a Kronk provider")
+		}
+		if req.Name == "" {
+			req.Name = provider.Name
+		}
+		if req.BaseURL == "" {
+			req.BaseURL = provider.BaseURL
+		}
+		if req.APIKey == "" {
+			req.APIKey = provider.APIKey
```
astaxie, line 210: [P1] Do not inherit a stored token for an overridden destination

When provider_id is supplied, a non-empty request BaseURL is kept, but an empty request APIKey is filled from the stored provider here (and stored headers are inherited below). An administrator can therefore POST discovery for a saved Kronk provider with base_url set to another reachable host and cause TokenHub to send Authorization: Bearer <saved token> and stored headers to that host. Please either reject BaseURL overrides when provider_id is used, or inherit credentials/headers only when discovery uses the stored destination; add a cross-origin regression test.

frontend/features/admin/views/provider-editor.tsx
```diff
@@ -494,8 +496,7 @@ export function ProviderUpsertModal({
     () => (effectiveDetail?.models ?? []).filter((model) => {
       if (quickAPIFlow) return true;
       if (modelCategory !== "all" && modelCategoryForCatalog(model) !== modelCategory) return false;
```
astaxie: [P1] Bypass the category gate for Kronk discovery

The new unlisted-model exception runs only after this category check. In the normal create flow the built-in Kronk entry is exposed under the custom category, but discovered IDs such as meta/llama-3:Q4_K_M.gguf or qwen/... are inferred as llama/qwen and return here before providerCatalogAcceptsUnlistedModels can accept them. The wizard therefore still hides common Kronk inventory and cannot select/import it. Please bypass this category gate for Kronk too (or derive the active categories from the live discovery result), and add a test covering the complete filter order rather than only the helper.

PR #210, opened 2026-08-13
fix(bootstrap): preserve admin team assignments on upgrade

### LINKED ISSUE #205: Bug when assigning users after upgrading from v0.4.0 to v0.5.0
State before the upgrade: admin was added to the newly created Team H, and the original default team, Default Project Space, was disabled.

astaxie:
## Summary

Fix startup failures when upgrading a preserved v0.4.0 database after the bootstrap administrator has been moved to a custom team and the default team has been disabled.

The v0.5.0 bootstrap attempted to create the fixed admin user on every start. New active-team validation ran before duplicate-user detection, so startup failed with “Only an active team can be assigned to a user” instead of recognizing the existing administrator.

## Related Issue

Fixes https://github.com/astaxie/TokenHub/issues/205

## Changes

- Skip bootstrap-admin creation when the persisted administrator identity already exists.
- Preserve customized administrator team assignments and disabled default-team state.
- Add a v0.4.0-shaped database upgrade regression test that runs current schema migration and startup bootstrap.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor or maintenance
- [ ] Documentation

## Verification

- go test ./...
- go vet ./...
- node --test tools/*.test.mjs
- node tools/check-doc-translations.mjs
- node tools/check-ui-translations.mjs
- node tools/check-env-contract.mjs
- node tools/check-source-lines.mjs
- git diff --check
- PostgreSQL integration suite not run locally because TEST_POSTGRES_URL is not configured; the persisted v0.4.0 upgrade path is covered with the SQLite regression fixture.

No public API, persistence schema, authentication policy, environment, deployment, or rollback changes. Existing data is preserved; rollback is the normal code rollback.

## Checklist

### DIFF
--- a/backend/internal/server/seed_test.go
+++ b/backend/internal/server/seed_test.go
@@ -2,6 +2,7 @@ package server
 
 import (
 	"context"
+	"database/sql"
 	"os"
 	"path/filepath"
 	"testing"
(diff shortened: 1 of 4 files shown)


backend/internal/server/seed_test.go
```diff
@@ -49,6 +50,81 @@ func TestRunStartupBootstrapReloadsCatalogOnEveryStart(t *testing.T) {
 	}
 }
 
+func TestRunStartupBootstrapUpgradesV040DatabaseWithDisabledDefaultTeam(t *testing.T) {
+	databasePath := filepath.Join(t.TempDir(), "v0.4.0.db")
```
astaxie, line 54: The reported upgrade failure is on PostgreSQL, but this regression only constructs and reopens a SQLite database. PostgreSQL takes a distinct migration/locking path here, so this test would still pass if the preserved v0.4.0 PostgreSQL upgrade continued to fail. It also stops after bootstrap instead of exercising the issue title’s post-upgrade user assignment. Could we add a `TEST_POSTGRES_URL`-backed v0.4-shaped upgrade case and verify that assigning a user to the active custom team succeeds while the existing admin/default-team state remains unchanged?

PR #214, opened 2026-08-14
fix(providers): support configurable synthetic DNS ranges

### LINKED ISSUE #211: SSRF guard rejects mihomo/Clash fake-ip range (198.18.0.0/15), breaking all upstream requests behind transparent proxies
## Summary

When TokenHub runs behind a transparent proxy that uses fake-ip DNS (the default mode of mihomo/Clash TUN, used by millions of deployments), **every** upstream provider request fails locally within ~5ms with `Failed to request upstream models` / connection-test failures. Disabling the transparent proxy immediately restores normal behavior.

## Root cause

`backend/internal/server/provider_upstream_ssrf.go` hardcodes `198.18.0.0/15` in `disallowedProviderUpstreamCIDRs` (the RFC 2544 benchmarking range). This list is applied **not only to literal-IP base URLs but also to hostname resolution results at dial time** (`dialGuardedUpstream` filters resolved candidates through `isDisallowedProviderUpstreamIP`).

With fake-ip DNS, **every** upstream hostname resolves to a 198.18.x.x placeholder:

```
api.deepseek.com  →  198.18.0.217   (fake-ip minted by mihomo)
```

The dial guard drops the only candidate as "disallowed" and `client.Do` fails instantly:

```
POST /api/admin/provider-catalog/custom
→ 502 {"error":{"code":"provider_models_request_failed","message":"Failed to request upstream models"}}  (~5ms, no packet ever leaves the host)
```

## Why this is a false positive

198.18.0.0/16 is mihomo's **default** `fake-ip-range`. A fake IP is not the real connection target: the transparent proxy intercepts the connection and maps it back to the original domain, which is the administrator-configured `base_url` host — exactly the value the save-time validation already checked. Blocking literal `198.18.x` base URLs makes sense; blocking **resolved** 198.18.x addresses misclassifies a DNS-level indirection as a dial target and provides no SSRF protection in this topology.

## Impact

- All guarded upstream paths fail: model discovery, connection test, health check, chat routing, streaming, failover (`http.go`, `provider_stream_timeout.go`, `provider_catalog.go` all wrap clients with the guard).
- Any deployment behind a fake-ip transparent proxy (Clash/mihomo TUN is the de-facto standard gateway on many home/edge boxes) is completely unable to call providers. node/curl from the same container work fine, which makes this very confusing to diagnose.
- `TOKENHUB_PROVIDER_UPSTREAM_ALLOWED_CIDRS` cannot help: 198.18.0.0/15 is in the non-bypassable always-deny list.

## Suggested fixes (any of)

1. Add an opt-in env, e.g. `TOKENHUB_PROVIDER_UPSTREAM_ALLOW_FAKEIP=true`, that exempts 198.18.0.0/15 (or a configurable range) from the **resolved-address** check while keeping the literal-IP denial.
2. Split the policy: deny 198.18.0.0/15 for literal IPs, allow it for hostname resolution results.
3. Support `HTTPS_PROXY` on the guarded transport (currently `transport.Proxy = nil` explicitly) so operators can route upstream calls through the proxy's HTTP port instead of relying on DNS interception.

Happy to contribute a PR for whichever direction you prefer.

## Workaround (verified)

Self-build with the `198.18.0.0/15` entry removed from `disallowedProviderUpstreamCIDRs`; all upstream paths (CN providers direct, blocked providers via proxy nodes) work correctly, and the rest of the guard (RFC1918, loopback, link-local, CGNAT, 240.0.0.0/4) stays intact.

## Environment

- TokenHub: 0.5.0 and main@616ac0b (both affected)
- Host: LazyCat MicroServer box, mihomo (Clash Meta) TUN mode, `enhanced-mode: fake-ip`, `fake-ip-range: 198.18.0.0/16`
- TokenHub deployed as a Docker container; container DNS chain: 127.0.0.11 → local DNS → mihomo:1053

deepjerry-ai:
## Summary

Allow TokenHub deployments behind fake-IP/synthetic-DNS transparent proxies to reach upstream providers without weakening the default SSRF policy. The exception is opt-in, configurable in Gateway Settings, and applies only to addresses resolved from administrator-configured hostnames.

## Related Issue

Closes #211

## Changes

- Add an opt-in synthetic DNS setting with configurable IPv4/IPv6 CIDRs and a documented `198.18.0.0/15` preset.
- Keep literal-IP provider URLs and protected address ranges blocked, while allowing configured synthetic ranges only after hostname resolution.
- Rotate upstream transport generations when the setting changes so new requests use the new policy without interrupting in-flight requests.
- Validate every settings mutation path, refresh settings safely across replicas, and propagate bootstrap backfill failures.
- Add regression coverage for validation, concurrent updates, cache refreshes, transport rotation, and SSRF boundaries.
- Add localized admin UI guidance and synchronized English, Simplified Chinese, and Japanese deployment documentation.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor or maintenance
- [x] Documentation

## Verification

- [x] `go test ./...`
- [x] `go test -race ./internal/server` with focused synthetic-DNS, settings mutation, transport rotation, and guarded dial tests
- [x] `go vet ./...`
- [x] `npm run typecheck`
- [x] `npm run build`
- [x] Documentation translation, UI translation, environment contract, source line, and `git diff --check` gates
- [x] Manual browser walkthrough of the settings form, validation state, and successful save

- No compatibility change to the OpenAI-compatible `/v1` API.
- The exception is disabled by default. Existing installations retain the current SSRF behavior until an administrator enables it.
- Literal IP URLs, loopback, link-local, metadata, multicast, NAT64, and other protected ranges remain non-bypassable.
- Enabling custom CIDRs deliberately trusts the transparent proxy to intercept those resolved addresses; the UI warns administrators about this security boundary.
- Local setting changes apply immediately. Other replicas converge from the shared store within approximately five seconds.
- Policy changes close idle connections and direct new requests to a new transport generation; in-flight requests are allowed to finish.
- Rollback by disabling the setting or reverting this change.

## Checklist

### DIFF
--- a/backend/internal/server/provider_upstream_synthetic_dns.go
+++ b/backend/internal/server/provider_upstream_synthetic_dns.go
@@ -1,0 +1,349 @@
+package server
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"net"
+	"net/http"
+	"strings"
+	"sync"
+	"time"
+)
+
+const (
+	gatewaySettingsID                 = "cfg_gateway"
+	syntheticDNSEnabledField          = "provider_synthetic_dns_enabled"
+	syntheticDNSCIDRsField            = "provider_synthetic_dns_cidrs"
+	defaultSyntheticDNSCIDRs          = "198.18.0.0/15"
+	syntheticDNSPolicyRefreshInterval = 5 * time.Second
+	syntheticDNSPolicyRefreshTimeout  = 2 * time.Second
+	maxProviderSyntheticDNSCIDRCount  = 16
+)
+
+// providerSyntheticDNSPolicy is the deliberately narrow exception for proxy
+// clients that replace real DNS answers with synthetic addresses. It applies
+// only after a hostname lookup; literal-IP provider URLs never consult it.
+// The short cache avoids a database read for every new TCP connection while
+// still converging settings across replicas without a restart.
+type providerSyntheticDNSPolicy struct {
+	store Store
+
+	mu          sync.Mutex
+	loadedAt    time.Time
+	blocks      []*net.IPNet
+	refreshDone chan struct{}
+	revision    uint64
+
+	transportsMu sync.Mutex
+	transports   []*providerUpstreamTransportPool
+}
+
+func newProviderSyntheticDNSPolicy(store Store) *providerSyntheticDNSPolicy {
+	policy := &providerSyntheticDNSPolicy{store: store}
+	policy.refresh(context.Background())
+	return policy
+}
+
+func (policy *providerSyntheticDNSPolicy) allowsResolvedIP(ip net.IP) bool {
+	return policy.allowsResolvedIPContext(context.Background(), ip)
+}
+
+func (policy *providerSyntheticDNSPolicy) allowsResolvedIPContext(ctx context.Context, ip net.IP) bool {
+	if policy == nil || ip == nil || isNonBypassableProviderSyntheticDNSIP(ip) {
+		return false
+	}
+	for _, block := range policy.configuredBlocks(ctx) {
+		if block.Contains(ip) {
+			return true
+		}
+	}
+	return false
+}
+
+type providerSyntheticDNSContextStore interface {
+	ListResourcesContext(context.Context, string) ([]AdminResource, error)
+}
+
+func (policy *providerSyntheticDNSPolicy) configuredBlocks(ctx context.Context) []*net.IPNet {
+	if policy == nil || policy.store == nil {
+		return nil
+	}
+	policy.mu.Lock()
+	if !policy.loadedAt.IsZero() && time.Since(policy.loadedAt) < syntheticDNSPolicyRefreshInterval {
+		blocks := policy.blocks
+		policy.mu.Unlock()
+		return blocks
+	}
+	if policy.refreshDone != nil {
+		done := policy.refreshDone
+		policy.mu.Unlock()
+		select {
+		case <-done:
+		case <-ctx.Done():
+		}
+		policy.mu.Lock()
+		blocks := policy.blocks
+		policy.mu.Unlock()
+		return blocks
+	}
+	policy.refreshDone = make(chan struct{})
+	done := policy.refreshDone
+	revision := policy.revision
+	policy.mu.Unlock()
+
+	refreshCtx, cancel := context.WithTimeout(ctx, syntheticDNSPolicyRefreshTimeout)
+	defer cancel()
+	settings, err := policy.listSettings(refreshCtx)
+	var blocks []*net.IPNet
+	if err == nil {
+		blocks = providerSyntheticDNSBlocksFromSettings(settings)
+	}
+
+	policy.mu.Lock()
+	changed := false
+	staleRefresh := policy.revision != revision
+	if err == nil && !staleRefresh {
+		changed = !providerSyntheticDNSBlocksEqual(policy.blocks, blocks)
+		policy.blocks = blocks
+	} else if err != nil && !staleRefresh {
+		log.Printf("[tokenhub] failed to refresh Provider synthetic DNS settings: %v", err)
+	}
+	if !staleRefresh {
+		policy.loadedAt = time.Now()
+	}
+	policy.refreshDone = nil
+	close(done)
+	current := policy.blocks
+	policy.mu.Unlock()
+	if changed {
+		policy.rotateTransports()
+	}
+	return current
+}
+
+func (policy *providerSyntheticDNSPolicy) refresh(ctx context.Context) {
+	if policy == nil {
+		return
+	}
+	_ = policy.configuredBlocks(ctx)
+}
+
+func (policy *providerSyntheticDNSPolicy) listSettings(ctx context.Context) ([]AdminResource, error) {
+	if store, ok := policy.store.(providerSyntheticDNSContextStore); ok {
+		return store.ListResourcesContext(ctx, "settings")
+	}
+	return policy.store.ListResources("settings"), nil
+}
+
+func providerSyntheticDNSBlocksFromSettings(settings []AdminResource) []*net.IPNet {
+	for _, setting := range settings {
+		if setting.ID != gatewaySettingsID || setting.Status != StatusActive || !truthyField(setting.Fields, syntheticDNSEnabledField) {
+			continue
+		}
+		blocks, err := parseProviderSyntheticDNSCIDRs(stringField(setting.Fields, syntheticDNSCIDRsField))
+		if err == nil {
+			return blocks
+		}
+		return nil
+	}
+	return nil
+}
+
+func (policy *providerSyntheticDNSPolicy) invalidate() {
+	if policy == nil {
+		return
+	}
+	policy.mu.Lock()
+	policy.revision++
+	policy.loadedAt = time.Time{}
+	policy.mu.Unlock()
+	policy.rotateTransports()
+}
+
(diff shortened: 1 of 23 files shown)


backend/internal/server/provider_upstream_synthetic_dns.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+	if !truthyField(resource.Fields, syntheticDNSEnabledField) {
+		return nil
+	}
+	_, err := parseProviderSyntheticDNSCIDRs(stringField(resource.Fields, syntheticDNSCIDRsField))
+	return err
+}
+
+func ensureProviderSyntheticDNSSettings(store Store) error {
+	for _, setting := range store.ListResources("settings") {
```
astaxie: The promised bootstrap error propagation still misses read failures here. Production `GormStore.ListResources` discards `Find(...).Error`, so a database error is indistinguishable from an empty settings list and this function returns nil without attempting or reporting the backfill. The added failure test covers only `UpdateResource`. Please use an error-returning/context-aware settings read for bootstrap and add a read-failure regression so startup cannot silently succeed with unknown policy state.

backend/internal/server/provider_upstream_transport_pool.go
```diff
@@ -0,0 +1,74 @@
...
+	pool.current = replacement
+	pool.mu.Unlock()
+	if retired != nil {
+		retired.CloseIdleConnections()
+	}
+}
+
+func rotatingProviderUpstreamTransport(allowedPrivate []*net.IPNet, policy *providerSyntheticDNSPolicy, configure func(*http.Transport)) http.RoundTripper {
+	factory := func() *http.Transport {
+		transport := ssrfGuardedProviderTransport(allowedPrivate, policy)
```
astaxie: Each transport generation closes over the same mutable `*providerSyntheticDNSPolicy`, so it does not actually preserve the policy snapshot under which an in-flight request started. If a request has entered `RoundTrip` but is still resolving/dialing when the setting is disabled, the retired transport calls `allowsResolvedIPContext` against the new blocks and rejects that request. The current test waits until the slow request is already connected, so it misses this case. Please snapshot the resolved CIDR policy into each generated transport (while refreshing/rotating before selecting a generation) and add a test that changes the setting while the old request is paused before dial completion.

backend/internal/server/provider_upstream_synthetic_dns.go
```diff
@@ -0,0 +1,349 @@
...
+		if providerSyntheticDNSCIDRContainsNonBypassableAddress(block) {
+			return nil, NewHTTPError(http.StatusBadRequest, "provider_synthetic_dns_cidrs_not_allowed", fmt.Sprintf("Synthetic DNS CIDR %q overlaps a protected address range", entry))
+		}
+		blocks = append(blocks, block)
+	}
+	return blocks, nil
+}
+
+// These ranges remain denied even when a synthetic-DNS exception is enabled.
+// Private/ULA, benchmarking, and documentation ranges are intentionally not
```
astaxie: This exception also overrides the existing RFC1918/ULA hostname guard: `10.0.0.0/8` and `fc00::/18` are explicitly accepted by the new tests, and `isNonBypassableProviderSyntheticDNSIP` no longer treats them as protected. Once an admin session enables either range, any provider hostname that resolves or rebinds to a real internal address in that range becomes dialable, which is broader than the Issue 211 Fake-IP carve-out and weakens the SSRF boundary the guard was designed to preserve against a compromised admin session. Please keep RFC1918 and ULA non-bypassable (and limit the exception to synthetic special-use pools such as `198.18.0.0/15`), or introduce a separately explicit unsafe trust mode instead of accepting private networks as ordinary synthetic ranges.

backend/internal/server/provider_upstream_synthetic_dns.go
deepjerry-ai: Fixed in 91ffc48. Bootstrap now reads settings through ListResourcesChecked, which surfaces database errors while preserving any context already attached to the GormStore (including the startup lease context). ensureProviderSyntheticDNSSettings wraps and propagates read failures. Added TestBootstrapReportsProviderSyntheticDNSBackfillReadFailure and TestListResourcesCheckedPreservesStoreContext.

backend/internal/server/provider_upstream_transport_pool.go
deepjerry-ai: Fixed in 91ffc48. Each transport generation now captures an immutable synthetic-DNS policy snapshot instead of closing over the mutable policy. Policy refresh and transport rotation happen before selecting a generation, while already selected requests keep their original snapshot. Added pre-dial and in-flight regression coverage in provider_upstream_transport_pool_test.go.

backend/internal/server/provider_upstream_synthetic_dns.go
deepjerry-ai: Fixed in 91ffc48. RFC1918 and ULA ranges remain blocked in ordinary synthetic-DNS mode. Trusting them now requires the separate provider_synthetic_dns_allow_private_ranges unsafe switch, which defaults to false, is backfilled explicitly, is preserved by demo seed data, and has dedicated UI/documentation warnings and regression tests.

astaxie (commented):
Re-reviewed the complete current diff at 91ffc48c after the author's commit and inline replies. The three previous findings are fixed: bootstrap settings reads now propagate database errors while preserving the store context, each transport generation captures an immutable policy snapshot, and RFC1918/ULA trust requires a separate explicit high-risk switch. The focused race-enabled synthetic-DNS, transport, backfill, and context tests pass, and all current required GitHub checks are green. I resolved the three verified threads.

This head still cannot be approved because it conflicts with current main in backend/internal/server/admin_providers_http.go (also reproduced with git merge-tree). Please resolve the conflict and rerun the required checks; the resulting head will need review before approval.

backend/internal/server/seed.go
```diff
@@ -265,13 +269,19 @@ func seedDefaultOrgResources(store Store) {
...
+			"default_timeout":             "120s",
+			"audit_retention":             "180d",
+			"api_key_prefix":              DefaultAPIKeyPrefix,
+			"api_key_random_length":       DefaultAPIKeyRandomLength,
+			syntheticDNSEnabledField:      false,
+			syntheticDNSCIDRsField:        defaultSyntheticDNSCIDRs,
+			syntheticDNSAllowPrivateField: false,
 		},
 	})
+	if err := ensureProviderSyntheticDNSSettings(store); err != nil {
```
astaxie, line 284: [P2] The promised bootstrap error propagation still misses the fresh/missing-row path. `seedResourceIfMissing` writes `cfg_gateway` through `CreateResource`, whose Gorm implementation discards the `CreateResourceChecked` error; if that insert fails, `ensureProviderSyntheticDNSSettings` then sees no matching row and returns nil. Bootstrap can therefore report success without persisting either the gateway row or these policy defaults, and the admin UI has no create action to repair it. Please create `cfg_gateway` through the checked API (or make the ensure helper create a missing row) and add a missing-row create-failure regression test.

astaxie (approved):
Re-reviewed the complete current diff at `55facf1a`. The new commit closes the remaining bootstrap gap by creating the missing `cfg_gateway` row through the checked store API and propagating both read and create failures, with regression coverage. The synthetic-DNS exception remains opt-in and hostname-only; literal IPs and protected loopback, link-local, metadata, multicast, and NAT64 ranges remain blocked, and policy changes rotate idle transports safely. The full backend suite, `go vet ./...`, frontend typecheck and production build, all 122 repository tool tests, translation/environment/source-line gates, and `git diff --check` pass. Required CI is green and the merge state is clean.

PR #218, opened 2026-08-15
feat(catalog): fetch upstream catalog on refresh

### LINKED ISSUE #217: feat(catalog): fetch latest model catalog from upstream on refresh
The bundled `data/model-catalog.yaml` is hand-maintained and currently pinned to a `PublicProviderConf/dist/jiekou.json` snapshot from 2026-07-08, while upstream has moved on (now 2026-08-14, with new models like gemini-3.6-flash, grok-4.6, qwen3.8-max, etc.).

The admin "refresh" action (`GET /api/admin/provider-catalog?refresh=true`) only reloads the bundled file from disk — it never fetches upstream. So the catalog goes stale between releases, and clicking refresh doesn't actually pull newer model data.

Suggestion: make refresh fetch the latest catalog from upstream `PublicProviderConf` (falling back to the bundled file when offline), so users can update model data on demand without waiting for a new release. The existing `validateProviderCatalogRefresh` guard could validate the fetched payload the same way it validates the local file.

Thanks for the great project!

astaxie:
## Summary

Make explicit administrator Provider-catalog refreshes download the current PublicProviderConf catalog instead of only reloading the bundled file. The refresh remains available offline by falling back to the configured local catalog whenever the upstream response cannot be fetched or validated.

## Related Issue

Closes #217.

## Changes

- Fetch `PublicProviderConf/dev/dist/all.json` for explicit `refresh=true` requests with a 15-second timeout and a 16 MiB response limit.
- Reuse the existing provider/model count, required-provider, and retention validation before atomically replacing the stored snapshot.
- Fall back to the configured local Provider catalog on upstream network, HTTP, size, parse, or validation failures.
- Preserve local-only startup initialization and expose whether the accepted snapshot came from the upstream or local catalog.
- Add deterministic HTTP-server tests for successful large upstream payloads, offline fallback, and incomplete-catalog fallback.
- Document the behavior consistently in English, Simplified Chinese, and Japanese.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor or maintenance
- [x] Documentation

## Verification

- `go test ./...`
- `go vet ./...`
- `go test -race ./internal/server -run 'TestProviderCatalogService(RefreshesFromUpstream|RefreshFallsBackToLocalCatalog|RefreshRejectsIncompleteUpstreamCatalog)$'`
- `git diff --check`
- Frontend build and SDK smoke tests were not run because this change does not modify frontend code or compatible API request handling and does not require a live Provider backend.

The OpenAI-compatible `/v1` API and database schema are unchanged. Startup remains local-only and therefore does not gain a network dependency. Explicit administrator refreshes make an unauthenticated GET request to a fixed GitHub Raw URL without forwarding credentials; the response is time-bounded, size-bounded, parsed, and validated before persistence. Offline or rejected responses use the configured local fallback, and the last known-good database snapshot remains intact if both sources fail. No environment or deployment configuration changes are required. Rollback is a code revert.

## Checklist


backend/internal/server/provider_catalog_refresh.go
```diff
@@ -0,0 +1,82 @@
...
+		entries, upstreamErr = prepareProviderCatalogRefresh(entries, previous)
+	}
+	if upstreamErr == nil {
+		if err := s.store.SaveProviderCatalogSnapshot(entries, providerCatalogUpstreamSource, time.Now().UTC()); err != nil {
+			return nil, providerCatalogUpstreamSource, err
+		}
+		return cloneCatalogEntries(entries, false), providerCatalogUpstreamSource, nil
+	}
+
+	entries, localErr := loadLocalProviderCatalog(s.catalogFile)
```
astaxie, line 45: [P1] Stop when the refresh context is canceled

`loadUpstreamProviderCatalog(ctx)` also fails when the caller cancels or when `RunClusterOperation` cancels `operationCtx` after losing its lease. This branch treats either cancellation as an ordinary upstream outage and continues to `SaveProviderCatalogSnapshot`, which is not context-aware. A replica can therefore persist the local fallback after losing the cluster lease, racing with and potentially overwriting a newer snapshot written by the new lease holder; a canceled admin request can likewise downgrade an upstream snapshot to the bundled catalog. Please return `context.Cause(ctx)` before entering the fallback path, and re-check it before either snapshot save.

PR #230, opened 2026-08-15
fix(gateway): omit default background field in Responses requests

### LINKED ISSUE #228: fix(gateway): Responses API always sends background:false, breaking strict-whitelist upstreams
## Bug: Responses API always sends `background:false`, breaking strict-whitelist upstreams

When TokenHub proxies a `/v1/responses` request through an OpenAI-compatible provider, it **always injects `background:false` into the upstream request body**, even when the client never sent a `background` field.

Some OpenAI-compatible upstreams (e.g. certain sub2api / third-party gateways) apply a **strict parameter whitelist** on the Responses endpoint and reject any parameter they don't implement. Because they don't special-case "`false` equals the default", they return:

```
HTTP 400  {"error":{"message":"Unsupported parameter: background","type":"invalid_request_error"}}
```

This happens for **any** value (`true` / `false` / `null`) — they reject the *presence* of the parameter.

### Root cause

`ResponsesRequest.MarshalJSON` in `backend/internal/server/types.go`:

```go
setRawJSONField(raw, "background", r.Background, true)   // line ~1064
```

The `present` argument is hard-coded to `true`, so `background` is written to the raw map **even when `r.Background` is the zero value `false`** and the client never provided it.

Note the sibling `r.raw == nil` branch does NOT have this problem:

```go
if r.raw == nil {
    return json.Marshal(requestAlias(r))   // struct tag `background,omitempty` correctly omits false
}
```

So only the raw-passthrough path (the normal one for `/v1/responses`, since `UnmarshalJSON` always populates `raw`) is affected.

### Why this matters

- `background` defaults to `false` in OpenAI's API. Explicitly sending `false` is semantically identical to omitting it.
- Codex's official client **never sends the field at all** (confirmed by packet capture).
- Best practice: a field equal to its default should be omitted on the wire. Sending it needlessly breaks compatibility with strict-whitelist upstreams.

### Suggested fix

Only write `background` when it is actually `true` (the non-default value):

```go
setRawJSONField(raw, "background", r.Background, r.Background)
```

This preserves `background:true` for clients that genuinely request background mode, while omitting the field otherwise — matching both the struct-tag branch and the official client behavior.

Happy to open a PR with this change plus a regression test.

demystify0137:
## What

`ResponsesRequest.MarshalJSON` forced `background:false` onto the wire:

```go
setRawJSONField(raw, "background", r.Background, true)   // present hard-coded to true
```

This injected `background:false` into the upstream request body even when the client never sent a `background` field.

## Why it's a problem

Some OpenAI-compatible upstreams (sub2api-style gateways) apply a strict parameter whitelist on `/v1/responses` and reject any parameter they don't implement:

```
HTTP 400  {"error":{"message":"Unsupported parameter: background","type":"invalid_request_error"}}
```

— for `true`, `false`, and `null` alike. They reject the *presence* of the parameter, not its value.

## The fix

`background` defaults to `false` in OpenAI's API, so sending `false` explicitly is semantically identical to omitting it. The `raw == nil` struct-tag branch already omits it via `background,omitempty`. This one-line change makes the raw-passthrough branch consistent — write `background` only when it is `true`:

```go
setRawJSONField(raw, "background", r.Background, r.Background)
```

This matches the official Codex client, which never sends the field.

## Tests

Added `responses_background_test.go` with two regression cases:
- default background (client omits the field) → field omitted from the marshalled body
- explicit `background:true` → preserved

```
--- PASS: TestResponsesRequestMarshalOmitsDefaultBackground
--- PASS: TestResponsesRequestMarshalKeepsExplicitBackgroundTrue
```

`go test ./internal/server/` shows one pre-existing failure, `TestAdminProviderCatalogRoutesPreserveRefreshQuery`, which fails identically on unmodified `main` (verified) and is unrelated to this change.

Fixes #228


astaxie (commented):
Reviewed the complete current diff at `d85745c7`. I did not find a code blocker: the change omits the default `background:false` field while preserving `background:true`, matching the Responses default and the existing struct-tag path. The focused marshal regressions, `go vet ./...`, all 122 repository tool tests, documentation/UI/environment/source-line gates, and `git diff --check` pass locally.

This head cannot be approved because GitHub reports no CI checks for it. Please ensure the required checks run and pass on this exact head; it will then need current-head approval.

astaxie (commented):
Required Backend CI is failing on the current head. The concrete failure is TestAdminProviderCatalogRoutesPreserveRefreshQuery: GET /api/admin/provider-catalog/fresh-provider?refresh=true expected 200 but received 404. I reproduced it 5/5 times on d85745c7 and 3/3 times on the current base f3ddaafc, so this appears to be an existing base-branch regression rather than a regression introduced by this two-file change. It still blocks approval under the required-check policy. Please update the branch after the base regression is fixed and ensure Backend and the aggregate CI check pass on this exact head.

astaxie (commented):
Reviewed the complete current diff after the main-branch merge. The focused Responses marshal tests pass locally, but this head is not approvable: the Backend check fails in TestTeamLeaderScopedResourceFilteringUsesConstantQueries/project_members, the aggregate CI check fails, and Frontend browser smoke is cancelled. I also found the remaining compatibility gap inline.

backend/internal/server/types.go
```diff
@@ -1061,7 +1061,7 @@ func (r ResponsesRequest) MarshalJSON() ([]byte, error) {
 	setRawJSONField(raw, "model", r.Model, r.Model != "")
 	setRawJSONField(raw, "input", r.Input, r.Input != nil)
 	setRawJSONField(raw, "stream", r.Stream, true)
-	setRawJSONField(raw, "background", r.Background, true)
+	setRawJSONField(raw, "background", r.Background, r.Background)
```
astaxie, line 1064: [P1] Delete explicit false and null values from the raw payload

UnmarshalJSON preserves the original background member in r.raw. When Background is false, setRawJSONField returns without changing that map, so explicit background:false and background:null are still forwarded upstream. Issue #228 says these upstreams reject the field for any value, and the current regressions only cover an omitted field and true. Please use the delete-capable helper (or explicitly delete the member when false) and add false/null regression cases.

PR #235, opened 2026-08-16
fix(routing): validate model provider protocol compatibility

### LINKED ISSUE #233: feat(routing): validate model↔provider protocol compatibility at route creation
### The incident

We added `gpt-5.6-luna` (an upstream that only speaks OpenAI `chat`/`responses`) to a provider entry of type `anthropic`. The admin UI accepted the route without any warning; every request then failed at runtime with an opaque upstream error:

```
400 provider_invalid_request
{"error":{"type":"invalid_request_error","message":"model gpt-5.6-luna does not support the claude_native interface [code=30011]"}}
```

Nothing in the route-creation flow indicated the mismatch — the catalog metadata for the model even declares its supported endpoints, but that information is never consulted.

### Why it happens

- A provider entry's upstream protocol is determined solely by `provider.Type` via the adapter registry (`gateway_http.go:882` → `adapterRegistry.Resolve(route.Provider.Type)`). One provider entry = one protocol; the same upstream must be registered twice (e.g. `openai_compatible` + `anthropic`) to serve both protocols.
- `validateRouteAdapter` (`admin_providers_http.go:1209`) only checks that the provider exists and its adapter is registered. It never compares the model's supported endpoints against the provider's protocol.
- Model catalog entries carry exactly the needed data (`metadata.endpoints: "responses,chat/completions,anthropic"`, see `provider_catalog.go:452`), but it is unused at validation time.

### Suggestion

At route create/update time, when the model's `metadata.endpoints` (or provider-model capabilities) is known, warn or reject if the provider type's protocol is not in the list — e.g. routing a `chat,responses`-only model to an `anthropic`-type provider should produce something like `route_protocol_mismatch` (or at minimum a confirmation warning in the UI), instead of a runtime 400 from the upstream.

Longer term, supporting multiple protocols per provider entry (choosing the adapter per-route or per-model) would remove the need to register the same upstream twice — but validation alone would have prevented this incident.

Related: #224 (rendering `metadata.endpoints` in the model detail would make this visible before binding; validation closes the loop).

Mr0bean:
## Summary

Reject routes whose Provider protocol cannot serve a catalog model's declared endpoint protocol. This prevents an Anthropic Provider from being attached to a responses-only model, including when the route is submitted with a newly created model.

## Related Issue

Fixes #233.

## Changes

- Validate route protocols from the registered Provider adapter capabilities.
- Validate routes supplied during model creation before the model is persisted.
- Add compatibility and rejection regression coverage.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor or maintenance
- [ ] Documentation

## Verification

- [x] `go test ./internal/server -run 'TestAdmin(RouteRejectsKnownModelProviderProtocolMismatch|ModelCreateRejectsInitialRouteProtocolMismatch|RouteAllowsCatalogModelWithMatchingProviderProtocol)$'`
- [x] `go test ./...`
- [x] `go vet ./...`
- [x] `git diff --check`
- [x] Two consecutive independent review scans reported no actionable findings.

Rejects only routes for models that explicitly declare catalog endpoint metadata. Custom models without endpoint metadata retain their existing routing behavior. No configuration, deployment, or data migration is required.

## Checklist

### DIFF
--- a/backend/internal/server/admin_provider_catalog_test.go
+++ b/backend/internal/server/admin_provider_catalog_test.go
@@ -96,6 +96,57 @@ func TestAdminCreatesProviderModelAndRoute(t *testing.T) {
 	}
 }
 
+func TestAdminRouteRejectsKnownModelProviderProtocolMismatch(t *testing.T) {
+	store := NewMemoryStore()
+	if err := SeedDemoData(store); err != nil {
+		t.Fatal(err)
+	}
+	provider := store.AddProvider(Provider{ID: "prv_native_anthropic", Name: "Native Anthropic", Type: ProviderAnthropic, Status: StatusActive, Healthy: true})
+	store.AddProviderModel(ProviderModel{ProviderID: provider.ID, UpstreamModel: "gpt-5.6-luna", Status: StatusActive})
+	store.AddModel(Model{Name: "gpt-5.6-luna", Modality: "chat", Status: StatusActive, Metadata: map[string]string{"endpoints": "responses,chat/completions"}})
+
+	response := doJSON(t, New(store).Handler(), http.MethodPost, "/api/admin/routing-rules", map[string]any{
+		"model_name": "gpt-5.6-luna", "provider_id": provider.ID, "provider_model": "gpt-5.6-luna", "status": StatusActive,
+	}, "")
+	if response.Code != http.StatusBadRequest || !strings.Contains(response.Body, `"code":"route_protocol_mismatch"`) {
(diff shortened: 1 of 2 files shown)


astaxie (approved):
Reviewed the complete current diff at . I found no substantive blocker in validating known catalog endpoint metadata against registered adapter capabilities while retaining custom-model compatibility. The full backend suite, , all 122 repository tests and gates, , and all GitHub checks pass on this head.

PR #241, opened 2026-08-16
fix(catalog): report seed model write failures

### LINKED ISSUE #220: fix(catalog): AddModel silently swallows seeding errors
## Bug: catalog seeding swallows errors / can silently skip models

While investigating why a model present in `data/model-catalog.yaml` was missing from the database, we traced the startup path:

`BootstrapBaseDataWithConfig` (`backend/internal/server/seed.go:146`) → `pruneProviderImportedModelCatalog` (`seed.go:186`) → `seedDefaultModelCatalog` (`seed.go:194`) → `store.AddModel` (`model_route_store.go:10`).

Two problems:

1. **`AddModel` swallows insert errors.** `model_route_store.go:13` does `model, _ = createModelRecord(...)`. If `createModelRecord` fails (e.g. unique-index conflict when a seed's `name` matches an existing row whose `ID` differs), the model is silently dropped and the operator has no way to know the catalog entry never made it into the DB.

2. **`pruneProviderImportedModelCatalog` deletes unconditionally.** It removes every model whose `metadata.source == "public-provider-conf"` before re-seeding, which can destroy user edits (custom prices, status) on those entries, and interacts badly with the fact that seeding is add/update-only (renamed entries are never reconciled).

### Suggestion

- Log (or return) the error from `AddModel` so a failed seed is visible in startup logs.
- Make the prune step selective (only prune entries that no longer exist in the catalog file), and/or surface a warning when a seed conflicts with an existing row.

Mr0bean:
## Summary

Make catalog seed write failures visible instead of silently dropping a model when a legacy database row conflicts with the catalog.

## Related Issue

Fixes #220.

## Changes

- Use the existing transactional model creation path while seeding the catalog.
- Return a model-specific wrapped error to the startup bootstrap path.
- Add a real SQLite legacy ID/name conflict regression test.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor or maintenance
- [ ] Documentation

## Verification

- [x] go test ./...
- [x] go vet ./...
- [x] node --test tools/*.test.mjs
- [x] node tools/check-ui-translations.mjs
- [x] node tools/check-source-lines.mjs
- [x] git diff --check
- [x] Two consecutive independent post-fix scans found no actionable issues.

A malformed or conflicting catalog seed now fails startup with the model name and database cause instead of continuing silently. No API, security, configuration, or deployment changes are required.

## Checklist

### DIFF
--- a/backend/internal/server/seed.go
+++ b/backend/internal/server/seed.go
@@ -224,7 +224,9 @@ func seedDefaultModelCatalog(store Store, catalogFile string) error {
 			model.OutputPriceUSDPer1M = existing.OutputPriceUSDPer1M
 			model.EmbeddingPriceUSDPer1M = existing.EmbeddingPriceUSDPer1M
 		}
-		store.AddModel(model)
+		if _, err := store.CreateModelWithRoutes(model, nil); err != nil {
+			return fmt.Errorf("seed catalog model %q: %w", model.Name, err)
+		}
 	}
 	return nil
 }
--- a/backend/internal/server/seed_test.go
+++ b/backend/internal/server/seed_test.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 )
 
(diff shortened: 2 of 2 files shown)


astaxie (approved):
Reviewed the complete current diff at . I found no substantive blocker in surfacing transactional catalog seed failures. The full backend suite, , all 122 repository tests and gates, , and all GitHub checks pass on this head.

PR #243, opened 2026-08-16
fix(catalog): preserve routes during catalog refresh

### LINKED ISSUE #231: fix(catalog): seed never creates model routes; pruneProviderImportedModelCatalog is dead code
## Bug: seeded catalog models have no routes; prune step is dead code

Found while investigating why `deepseek-v4-flash` (bare catalog name) started returning `503 No available provider route` after an app reinstall, while `claude-deepseek-v4-flash` and `deepseek/deepseek-v4-flash` kept working.

### Root cause (two related defects in the startup path)

**1. `pruneProviderImportedModelCatalog` can never match anything (dead code).**

`backend/internal/server/seed.go:199-205`:

```go
func pruneProviderImportedModelCatalog(store Store) {
	for _, model := range store.ListModels() {
		if model.Metadata != nil && model.Metadata["source"] == "public-provider-conf" {
			_ = store.DeleteModel(model.Name)
		}
	}
}
```

But the catalog loader unconditionally overwrites `source` for every seeded entry — `backend/internal/server/model_catalog_data.go:108`:

```go
if strings.TrimSpace(metadata["source"]) == "" {
	metadata["source"] = "tokenhub-standard-catalog"
}
```

and every entry in `data/model-catalog.yaml` already carries `source: "tokenhub-standard-catalog"`. So no model in the DB ever has `source == "public-provider-conf"`, the prune never fires, and `DeleteModel` (which **does** cascade-delete routes, `store_routing_calls.go:163`) is effectively unreachable from this path.

**2. `seedDefaultModelCatalog` only creates model entries — never routes.**

`seed.go:207-226` calls `store.AddModel(model)` for each catalog entry but never `AddRoute`. Routes are only created later via the admin API when a user explicitly "接入"s a model. Consequence:

- After a fresh install / DB rebuild, seeded models are `active` in the directory but have **zero routing rules**.
- Any request to a bare catalog name (e.g. `deepseek-v4-flash`) fails with `503 provider_unavailable / No available provider route` even though the model shows up as available in the UI.
- Restarting never heals it; there is no startup warning either (see #220 — `AddModel` also swallows errors silently).

### Reproduction

1. Fresh DB → start backend → catalog seeds models (`deepseek-v4-flash` present, `models` table, `status=active`).
2. `GET /api/admin/routing-rules` → no entry for `deepseek-v4-flash`.
3. `POST /v1/messages {"model":"deepseek-v4-flash", ...}` with a valid API key → `503 {"error":{"code":"provider_unavailable","message":"No available provider route"}}` (latency ~0, rejected at gateway).
4. Manually flag a model's metadata as `public-provider-conf` in the DB → restart → model is **not** pruned (confirms defect 1).

### Expected

Either of:

- `seedDefaultModelCatalog` also creates a default route for catalog entries that declare a matching provider (e.g. `provider_catalog_id` in metadata), so freshly seeded models are actually callable; or
- If routes are intentionally manual-only, then (a) remove the dead prune code or fix its match key, and (b) surface "model has no route" as a warning in startup logs / model list UI so operators aren't surprised by 503s on active models.

### Environment

- Self-built image `0.5.1-pr191` (same seed code as master), SQLite backend, running as a Lazycat micro-service app.

Related: #220 (same startup path — the swallowed-error bug meant the missing routes never showed up in any log).

Mr0bean:
## Summary

Preserve configured routes while refreshing catalog models by removing an obsolete startup prune path that could delete a model and cascade-delete its routes.

## Related Issue

Fixes #231.

## Changes

- Remove the unreachable and destructive provider-imported catalog prune step.
- Keep catalog refresh as an upsert that preserves routes.
- Add a regression test using the tracked deepseek-v4-flash catalog entry.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor or maintenance
- [ ] Documentation

## Verification

- [x] go test ./...
- [x] go vet ./...
- [x] node --test tools/*.test.mjs
- [x] node tools/check-ui-translations.mjs
- [x] node tools/check-source-lines.mjs
- [x] git diff --check
- [x] Two consecutive independent post-fix scans found no actionable issues.

The change does not invent default Provider routes. It retains operator-configured routes during catalog refresh and continues to expose unrouted models through the existing administrative state. No API, credential, configuration, or deployment changes are required.

## Checklist

### DIFF
--- a/backend/internal/server/seed.go
+++ b/backend/internal/server/seed.go
@@ -156,7 +156,6 @@ func BootstrapBaseDataWithConfig(store Store, config Config) error {
 	if err := seedBuiltinProviderCatalog(store); err != nil {
 		return err
 	}
-	pruneProviderImportedModelCatalog(store)
 	if err := seedDefaultModelCatalog(store, config.ModelCatalogFile); err != nil {
 		return err
 	}
@@ -198,14 +197,6 @@ func seedDefaultProject(store Store) {
 		CostCenter:  "AI-PLATFORM",
 		Status:      StatusActive,
 	})
-}
-
-func pruneProviderImportedModelCatalog(store Store) {
-	for _, model := range store.ListModels() {
-		if model.Metadata != nil && model.Metadata["source"] == "public-provider-conf" {
-			_ = store.DeleteModel(model.Name)
-		}
-	}
 }
 
 func seedDefaultModelCatalog(store Store, catalogFile string) error {
--- a/backend/internal/server/seed_test.go
+++ b/backend/internal/server/seed_test.go
@@ -213,3 +213,47 @@ models:
 		t.Fatalf("new catalog model did not use catalog prices: %+v", newModel)
 	}
 }
+
+func TestBootstrapBaseDataPreservesRoutesWhenRefreshingLegacyCatalogModel(t *testing.T) {
+	store := NewMemoryStore()
+	const modelName = "deepseek-v4-flash"
+	store.AddModel(Model{
+		Name:     modelName,
+		Modality: "chat",
+		Status:   StatusActive,
+		Metadata: map[string]string{"source": "public-provider-conf"},
+	})
+	provider := store.AddProvider(Provider{
+		ID:      "prv_catalog_route",
+		Name:    "Catalog Route Provider",
+		Type:    ProviderOpenAICompatible,
+		Status:  StatusActive,
+		Healthy: true,
+	})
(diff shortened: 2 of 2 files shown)


astaxie (approved):
Reviewed the complete current diff at . I found no substantive blocker in removing the destructive catalog prune path and preserving routes. The full backend suite, , all 122 repository tests and gates, , and all GitHub checks pass on this head.

PR #244, opened 2026-08-16
perf(gateway): enlarge upstream idle connection pool

legendtkl:
## Summary

Every upstream provider call goes through the SSRF-guarded transport, which clones `http.DefaultTransport` and therefore inherits its default of **2 idle connections per host**. A gateway fans a large amount of concurrent traffic out over a handful of upstream hosts (`api.openai.com` etc.), so under any real concurrency the surplus connections are retired the moment they go idle and nearly every request to a busy host pays a fresh TCP + TLS handshake. This PR sizes the idle pool for a gateway workload.

## Related Issue

N/A (first PR from an internal performance review of the request hot path).

## Changes

- `ssrfGuardedProviderTransport` now sets `MaxIdleConnsPerHost = 64` and `MaxIdleConns = 256` after cloning. These cap idle-pool retention only; `MaxConnsPerHost` stays unset so the guard never throttles concurrency.
- The rare fallback path (when `http.DefaultTransport` is not an `*http.Transport`) now sets `IdleConnTimeout: 90 * time.Second` explicitly, so the enlarged pool can never hold connections forever on that path.
- New test `TestSSRFGuardedProviderTransportPoolsIdleConnections`: an 8-way concurrent burst against a local server must open 8 connections, keep all of them pooled (asserted via `httptrace.ClientTrace.PutIdleConn`), and a follow-up serial round must open **no** new connections. Under the old default of 2 the test fails with `http: putIdleConn: too many idle connections for host`.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor or maintenance
- [ ] Documentation

## Verification

- `gofmt -l .` — clean.
- `go test ./... -count=1` (backend) — all packages ok, run twice independently.
- `go test ./internal/server/ -run TestSSRFGuardedProviderTransportPoolsIdleConnections -race -count=10` — ok.
- `go vet ./...` — clean.
- `golangci-lint run ./...` (v2.12.2, same version/working-directory as CI) — 10 pre-existing `staticcheck SA5011` findings in unrelated test files; verified via stash diff that the output is byte-identical with and without this change (no new findings).
- `git diff --check` — clean.
- End-to-end smoke on a real process (temp SQLite, seeded demo data): `/livez`, non-streaming and streaming `/v1/chat/completions` via the mock provider, and `/api/admin/usage/summary` all pass.
- Discriminating check: temporarily reverting `MaxIdleConnsPerHost` to 2 makes the new test fail as expected, then restored.

- No API or behavior change other than connection reuse. SSRF guard, DialContext, proxy disablement, and redirect policy are untouched.
- Rotation of synthetic-DNS transports already calls `CloseIdleConnections()` on the retired transport; connections still serving in-flight requests during a rotate strand in the retired pool (unreachable for new requests) until the 90 s idle timeout — same behavior as before, now with a larger upper bound (64/host instead of 2/host) on that transient.
- File-descriptor budget: the gateway keeps 2 long-lived upstream transports (stream + non-stream), so the idle-pool ceiling is ~512 FDs process-wide (128 per host). Worth keeping in mind for containers with `ulimit -n 1024`; no deployment file change included.
- HTTP/2 upstreams multiplex and are unaffected; the win applies to HTTP/1.1 upstreams.

## Checklist

### DIFF
--- a/backend/internal/server/provider_catalog_ssrf_test.go
+++ b/backend/internal/server/provider_catalog_ssrf_test.go
@@ -4,13 +4,16 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"io"
 	"net"
 	"net/http"
 	"net/http/httptest"
+	"net/http/httptrace"
 	"net/url"
 	"os"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


astaxie (approved):
Reviewed the complete current diff at . I found no substantive blocker in the guarded transport pool sizing or fallback timeout behavior. The full backend suite, , all 122 repository tests and gates, , and all GitHub checks pass on this head.

PR #245, opened 2026-08-16
perf(gateway): probe stream frames with a single decode

legendtkl:
## Summary

The OpenAI-compatible streaming path (`copyOpenAIStreamAndUsageForProvider`) fully decoded **every** SSE frame into `map[string]any` up to twice: once in `providerStreamEventIsError` (to decide whether the frame needs secret redaction) and once in `usageFromServerSentEvent` (to read usage for billing). A token stream is almost entirely frames that are neither an error nor a usage report, so both decodes were pure overhead on the gateway's hottest path. This PR answers both questions from a single decode per frame.

## Related Issue

N/A (performance-review series; previous: #244).

## Changes

- New `providerStreamEventProbe` — one `json.Unmarshal` of the payload into `map[string]json.RawMessage`; `isError()` and `usage()` share it. The main copy loop probes once per frame.
- A **map, not a struct**, on purpose: `encoding/json` matches struct fields case-insensitively, so a struct-based probe would let a frame hide its `error` behind a later `ERROR` key and skip redaction (verified experimentally). Map keys match exactly, like the old lookups.
- All raw values stay `json.RawMessage`, so a field of an unexpected JSON type cannot fail the whole decode and take the other answer with it — the same per-key independence the old type assertions had.
- Multi-line payloads (the single-newline-joined chunks some OpenAI-compatible servers emit, or legal multi-`data:`-line frames) keep the **verbatim legacy parsing**, because which joined segment wins depends on which segments parse at all. Token streams are single-line frames and never pay for it.
- `providerStreamEventIsError` / `usageFromServerSentEvent` / `usageFromSSEPayload` keep their signatures; the first two now have only test callers and are kept deliberately as the named, comparison-tested definition of this behavior.
- New `provider_stream_probe_test.go`: 77 table-driven payloads asserted against verbatim copies of the legacy parsers (escaped keys, shadowed/duplicate keys, nested `response.error`, null/non-object fields, non-object payloads, joined segments, multi-line objects, whitespace, invalid UTF-8), plus a pinned test for the one intended difference and a benchmark.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor or maintenance
- [ ] Documentation

## Verification

- `gofmt -l .` — clean; `go vet ./...` — clean; `git diff --check` — clean.
- `go test ./... -count=1` (backend) — all packages ok, run independently twice (implementer + reviewer session).
- `golangci-lint run ./...` (v2.12.2, CI-matching) — 10 pre-existing SA5011 findings in unrelated files; stash-diff confirms zero new findings.
- `node tools/check-source-lines.mjs` — pass.
- Benchmarks (arm64, mean of 6): chat chunk **2328 ns/op / 27 allocs** vs 6327 / 88 legacy (2.7x, −69% allocs); Responses delta 1.9x; usage frame 1.4x.
- End-to-end smoke on a real process (temp SQLite, seeded demo data): `/livez`, non-streaming and streaming `/v1/chat/completions`, `/api/admin/usage/summary` — pass.
- Known flake: `go test ./...` failed once on `TestTeamLeaderScopedResourceFilteringUsesConstantQueries` / `TestBackgroundResponsesSuccessPersistsEncryptedPayloadAndAccountsOnce`; reproduced 2/3 runs on unmodified `origin/main` in a separate worktree — pre-existing load-sensitive flake, unrelated to this change.

- Wire format unchanged: frames reach the client byte-identical (`event.Raw`), redaction call sites untouched.
- One intended behavior difference, pinned by test: a single-line frame that is legal JSON except for a number too large for float64 in a field the probe does not read used to fail the whole decode — dropping billable usage and, worse, letting an error frame **skip redaction**. The probe now reads such frames. Both directions are strictly safer (usage that was lost is now billed; errors that escaped redaction are now redacted); no previously-billed value can change, since single-line frames had no fallback path.
- Security-reviewed via four Codex review rounds (high effort): round 1 caught the struct case-insensitivity redaction bypass; rounds 2–3 removed any input whose billed value could change on the multi-line path; round 4 verdict "ready to commit — no input found where old=error / new=not-error".

## Checklist

### DIFF
--- a/backend/internal/server/provider_stream_probe_test.go
+++ b/backend/internal/server/provider_stream_probe_test.go
@@ -1,0 +1,297 @@
+package server
+
+import (
+	"encoding/json"
+	"fmt"
+	"reflect"
+	"strings"
+	"testing"
+)
+
+// The probe replaced two full map[string]any decodes per frame with one decode
+// of four raw fields. These tests pin the new parsing to the old one it stands
+// in for: legacyProviderStreamEventIsError and legacyUsageFromServerSentEvent
+// below are verbatim copies of the code the probe replaced, and every payload
+// class the streaming path can meet is run through both.
+
+func legacyProviderStreamEventIsError(event serverSentEvent) bool {
+	eventName := strings.ToLower(strings.TrimSpace(event.Event))
+	if eventName == "error" || strings.HasSuffix(eventName, ".failed") {
+		return true
+	}
+	payload := strings.TrimSpace(event.Data)
+	if payload == "" || payload == "[DONE]" {
+		return false
+	}
+	var decoded map[string]any
+	if json.Unmarshal([]byte(payload), &decoded) != nil {
+		return false
+	}
+	if eventType, _ := decoded["type"].(string); strings.EqualFold(strings.TrimSpace(eventType), "error") || strings.HasSuffix(strings.ToLower(strings.TrimSpace(eventType)), ".failed") {
+		return true
+	}
+	if errorValue, hasError := decoded["error"]; hasError && errorValue != nil {
+		return true
+	}
+	response, _ := decoded["response"].(map[string]any)
+	errorValue, hasResponseError := response["error"]
+	return hasResponseError && errorValue != nil
+}
+
+func legacyUsageFromServerSentEvent(frame serverSentEvent) (Usage, bool) {
+	if usage, ok := legacyUsageFromSSEPayload(frame.Data); ok {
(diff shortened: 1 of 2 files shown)


astaxie (approved):
Reviewed the complete current diff at . I found no substantive blocker in the single-decode stream probe or its error-redaction and usage-preservation behavior. The full backend suite, , all 122 repository tests and gates, , and all GitHub checks pass on this head.

PR #247, opened 2026-08-16
perf(metrics): cache the model catalog behind the rejection label

legendtkl:
## Summary

`knownModelLabel` bounds a metrics label to catalog-known model names; its only production caller is `RecordRejectedRequest` — the rejection path. Every rejected request paid one `COUNT` query for that check, so the cheapest outcome in the gateway (refusing a request) carried a database round trip that a client looping over invented model names or bad keys could multiply at will. This PR replaces the per-request query with a 60-second cached snapshot of the catalog's name set.

## Related Issue

N/A (performance-review series; previous: #244, #245).

## Changes

- New `modelLabelCache` (`model_label_cache.go`): the set of catalog model names behind a mutex, refreshed at most once per 60 s TTL, held as a **pointer field** on `GormStore` so `WithContext`'s shallow copy shares one cache (same pattern as the store's `mu` and `leaseHeartbeats`).
- `knownModelLabel` answers from the snapshot. Failure semantics are deliberately conservative: a failed refresh keeps serving the previous snapshot and backs off 5 s; a cache that has **never** loaded collapses names to `"unknown"` instead of falling back to per-request queries against a database that is already failing. The label set can never widen past the catalog under any failure.
- Refresh timestamps are taken **after** the load returns, so a slow failing query cannot hand back an already-elapsed backoff window (caught by Codex review round 1).
- All four writers of the `models` table invalidate the snapshot after their transaction commits: `AddModel`, `CreateModelWithRoutes`, `UpdateModel` (renames), `DeleteModel`. Cross-replica and out-of-band changes (e.g. SQLite backup restore) converge within one TTL — acceptable for a label that takes no part in authentication or routing, documented in code.
- 400-line test file: behavior parity, invalidation on each mutator, TTL expiry, refresh-failure fallback, database-outage query bound (≤1 query across 10 rejections), backoff/recovery timing with a fake clock, `-race` concurrency. Each guard was mutation-tested (4 targeted mutations each make specific tests fail).

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor or maintenance
- [ ] Documentation

## Verification

- `gofmt -l .` / `go vet ./...` / `git diff --check` — clean.
- `go test ./internal/server/ -run 'ModelLabel|RejectedRequest' -count=3 -race` — ok, no flake.
- `go test ./... -count=1` — green (run independently in two sessions). One unattributed full-suite failure occurred during iteration; 8 of 9 full/package runs green, the failure's log signature matches the pre-existing load-sensitive flake already reproduced 2/3 runs on unmodified `origin/main` during PR #245 validation (`TestBackgroundResponsesSuccessPersistsEncryptedPayloadAndAccountsOnce` area), and this change's only production call path (`RecordRejectedRequest`) does not intersect response-job workers.
- `golangci-lint run ./...` (v2.12.2, CI-matching) — 10 pre-existing SA5011 findings in unrelated files, none in changed files, no new findings.
- Micro-benchmark (in-memory SQLite): cache hit ~43 ns vs ~5.3 µs per query (~120x; larger over a networked PostgreSQL).
- End-to-end smoke on a real process (temp SQLite, seeded demo): `/livez`, non-streaming + streaming `/v1/chat/completions`, `/api/admin/usage/summary` — pass.
- Codex review, 3 rounds (high effort): round 1 found timestamp-before-load backoff bug (fixed); round 2 found first-load fallback re-amplifying queries during outages (fixed by collapsing to `"unknown"`); round 3 — no new findings, one extra recovery-path regression test added on its suggestion.

- Metrics label values are unchanged for a healthy database; during an outage, known models may label as `"unknown"` for up to one 5 s backoff window, and cross-replica model changes converge within 60 s. The security property — arbitrary client input can never mint a metrics label — holds under all failure modes.
- No API, schema, or configuration change.

## Checklist

### DIFF
--- a/backend/internal/server/model_label_cache.go
+++ b/backend/internal/server/model_label_cache.go
@@ -1,0 +1,92 @@
+package server
+
+import (
+	"sync"
+	"time"
+)
+
+const (
+	// modelLabelTTL bounds how stale a cached catalog snapshot may become. Model
+	// changes made through this process invalidate the snapshot immediately;
+	// changes made by another replica are picked up within one TTL. The snapshot
+	// only decides whether a model name is safe to use as a metrics label — it
+	// takes no part in authentication or routing — so a label that lags the
+	// catalog by up to one TTL is acceptable.
+	modelLabelTTL = 60 * time.Second
+	// modelLabelRetryBackoff keeps a database outage from turning every rejected
+	// request back into a query: a failed refresh is not retried before it elapses.
+	modelLabelRetryBackoff = 5 * time.Second
+)
+
+// modelLabelCache holds the set of model names the catalog knows, so bounding a
+// metrics label to that set costs a map lookup instead of a query per request.
+type modelLabelCache struct {
+	mu sync.Mutex
+	// names stays nil until a refresh has succeeded at least once.
+	names       map[string]struct{}
+	fetchedAt   time.Time
+	nextAttempt time.Time
+}
+
+func newModelLabelCache() *modelLabelCache {
+	return &modelLabelCache{}
+}
+
+// invalidate marks the snapshot stale so the next lookup reloads it. Callers must
+// not hold an open database transaction: a refresh already in flight runs its
+// query while holding the same mutex.
+func (c *modelLabelCache) invalidate() {
+	if c == nil {
+		return
+	}
+	c.mu.Lock()
+	c.fetchedAt = time.Time{}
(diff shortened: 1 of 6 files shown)


astaxie (approved):
Reviewed the complete current diff at . I found no substantive blocker. The focused race test, full backend suite, , all 122 repository tests, translation/environment/source-line gates, , and all GitHub checks pass on this head.

PR #248, opened 2026-08-16
perf(store): throttle last-used bookkeeping writes

legendtkl:
## Summary

Every gateway request performed up to three bookkeeping UPDATEs — the API key's `last_used_at` during validation, and the route's and provider resource's marks after a successful call — each taking the store-wide mutex and its own write transaction, for columns that only feed "last used" displays in the admin console. The API-key write could even fail an otherwise valid request. This PR routes all three through a throttle that persists at most one write per object per 60-second window.

## Related Issue

N/A (performance-review series; previous: #244, #245, #247).

## Changes

- New `lastUsedThrottle` (`last_used_throttle.go`): namespaced keys (`key:` / `route:` / `resource:`), a claim protocol (reserve → write → commit / failed-at backoff) so failed **or panicking** writes suppress repeated attempts and logs for one window without leaving a permanent in-flight placeholder, and concurrent claimers collapse to a single writer.
- Hard capacity bound (100k entries) with pressure-driven sweeping: reservation uses a CAS counter, a full map can trigger at most one sweep per minute, and when no slot frees up the write proceeds **unthrottled** (degrades to today's behavior) instead of growing memory. `entries ≤ maxEntries` is an invariant, not a soft target.
- Internal clock uses monotonic `time.Now()` (a wall-clock rollback cannot stretch the suppression window); the persisted timestamp is sampled by the caller under the store mutex.
- `ValidateAPIKey`: throttled write; a failure no longer rejects a valid key and records a failed-at state that backs off repeated attempts and logs for one window. The returned key still reports the current request's timestamp (callers never write it back), documented in code.
- `MarkRouteUsed` / `MarkProviderResourceUsed`: throttled; the store mutex is only taken when a write is actually due. `updated_at` keeps being written together with the resource's `last_used_at` (existing semantics), so its use-triggered updates share the 60 s resolution.
- `CompleteImageJob`: the two inline mark UPDATEs move out of the job transaction and go through the same throttle after commit (verified the surrounding closure releases the store mutex first — no self-deadlock; marks and job completion are no longer atomic, acceptable for display columns and documented).
- 19 new tests (`-race`): window semantics, per-namespace isolation, failure backoff/retry, panic recovery, 64/256-goroutine concurrency collapse, capacity bound under concurrency, successful/failed state sweeping and rate limits, `WithContext` sharing, `ValidateAPIKey` degradation, image-job paths.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor or maintenance
- [ ] Documentation

## Verification

- `gofmt -l .` / `go vet ./...` / `git diff --check` — clean.
- Narrow suite `-race` — ok (implementer and reviewer sessions independently).
- `go test ./... -count=1` — green runs in both sessions; intermittent failures observed across repeated runs are the two pre-existing load-sensitive flakes (`TestTeamLeaderScopedResourceFilteringUsesConstantQueries/project_members`, `TestBackgroundResponsesSuccessPersistsEncryptedPayloadAndAccountsOnce`), each reproduced on an unmodified `origin/main` worktree at comparable rates (2/8 and 1/10); neither intersects this change's paths (admin resource filtering query counting; response-job audit-event ordering).
- `golangci-lint run ./...` (v2.12.2, CI-matching) — 10 pre-existing SA5011 findings in untouched files; one transient errcheck introduced during development was caught and fixed before review; no new findings.
- Existing `last_used_at` assertions audited: all assert first-use nil→non-nil transitions, which still write immediately; zero tests needed changes.
- End-to-end smoke on a real process (temp SQLite, seeded demo): `/livez`, non-streaming + streaming `/v1/chat/completions`, `/api/admin/usage/summary` — pass.
- Codex review, 2 rounds (high effort): round 1 found the soft capacity bound, a panic-leaves-placeholder bug, and wall-clock/monotonic issues — all fixed with tests; round 2 passed with no blocking findings.

- Semantics change (intended, display-only): persisted `last_used_at` (and the resource's use-triggered `updated_at`) resolution drops from per-request to one write per object per 60 s window; the API returns exact per-request values in `ValidateAPIKey`'s copy. These columns take no part in authentication, quotas, or routing (audited call sites).
- A failed bookkeeping write no longer fails the request (previously `ValidateAPIKey` rejected the key). While an object is tracked, failures retry and log at most once per object per window.
- Multi-replica deployments throttle independently; cross-replica timestamp skew is bounded by one window plus the write duration. No SQL CASE guard added — deliberate trade-off, documented in code.
- No API, schema, or configuration change.

## Checklist

- [x] Shared user-facing behavior is documented consistently in English, Simplified Chinese, and Japanese where applicable. (Admin-console "last used" columns simply update less often; no copy changes.)


backend/internal/server/last_used_throttle.go
```diff
@@ -0,0 +1,238 @@
...
+	if !ok {
+		return nil
+	}
+	// The claim is released on every path that is not a successful write,
+	// panics included: a leaked in-flight placeholder would suppress this
+	// object's writes until the process restarts.
+	committed := false
+	defer func() {
+		if !committed {
+			t.rollback(claim)
```
astaxie: [P2] Back off failed bookkeeping writes

Rolling the claim back immediately means a persistent write failure makes every subsequent request reissue the same UPDATE, and each caller logs the error. On a hot API key, route, or resource, SQLite lock contention or a database write outage therefore restores the per-request write load and creates an unbounded log storm, contradicting the PR's stated guarantee that failures log at most once per object per window. Please preserve retryability with a short failed-at/backoff state (and throttle the corresponding log) so failures cannot bypass the window entirely.

astaxie (commented):
Reviewed the complete current diff at `34bb27e35`. The full backend test suite, `go vet ./...`, and `git diff --check` pass, and GitHub CI is green. I found one operational throttling issue inline. The PR is also currently conflicting with the base branch, so this is a comment-only review.

backend/internal/server/last_used_throttle.go
legendtkl: Addressed in b4689bef. Failed and panicking tracked writes now record an explicit failed-at state, so subsequent requests skip both the UPDATE and the caller-side error log until the 60-second backoff expires. The next request after the backoff retries normally, and stale failed states are reclaimable by the bounded sweep. The focused race suite and the full backend test suite pass.

astaxie (approved):
Re-reviewed the complete current diff at b4689bef0. The prior failed-write throttling issue is fixed: failed and panicking writes now back off for 60 seconds without suppressing later retry. Full backend tests, go vet, repository gates, and diff checks pass; required CI is green. Approving this exact head.

PR #250, opened 2026-08-16
fix(gateway): expose codex cache and reasoning metadata

Mr0bean:
## Summary

Expose Codex cache accounting and encrypted reasoning continuation through the OpenAI-compatible fields understood by standard clients, including DeepSeek Harness and pi-ai.

TokenHub previously returned cache and reasoning counters only through TokenHub-specific top-level usage fields. Clients that read the standard nested OpenAI usage objects therefore reported zero cache hits. The Codex Chat bridge also emitted a standalone reasoning signature that pi-ai displayed but did not persist for the next tool-call turn.

## Related Issue

N/A.

## Changes

- Emit standard `prompt_tokens_details` and `completion_tokens_details` for Chat Completions, and the corresponding `input_tokens_details` and `output_tokens_details` for Responses.
- Preserve existing TokenHub usage aliases so current clients remain compatible.
- Emit tool-bound encrypted `reasoning_details` in streaming and non-streaming Codex Chat responses and restore the continuation on the next tool-result request.
- Add focused regression tests for usage mapping, streaming output, route conversion, tool binding, and rejection of foreign or unbound reasoning details.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor or maintenance
- [ ] Documentation

## Verification

- [x] `cd backend && go test ./...`
- [x] `node tools/check-doc-translations.mjs`
- [x] `node tools/check-ui-translations.mjs`
- [x] `git diff --check`

The response schema gains standard nested token details while retaining all existing TokenHub aliases. No request field is removed, no environment variable or database migration is introduced, and no credential, backup, or runtime data is included. Codex upstream requests and cache behavior are unchanged; this change fixes response accounting and multi-turn reasoning-state preservation.

## Checklist

### DIFF
--- a/backend/internal/server/codex_compat_request.go
+++ b/backend/internal/server/codex_compat_request.go
@@ -481,7 +481,7 @@ func chatMessageToCodex(message ChatMessage, toolNames map[string]string) ([]any
 		}}, nil
 	case "assistant":
 		items := make([]any, 0, 3)
-		if encrypted, ok := decodeProviderSignature(codexSignatureProvider, message.ReasoningSignature); ok {
+		if encrypted, ok := chatCodexReasoningContinuation(message); ok {
 			reasoning := map[string]any{
 				"type":              "reasoning",
 				"summary":           []any{},
@@ -548,6 +548,42 @@ func chatMessageToCodex(message ChatMessage, toolNames map[string]string) ([]any
 	default:
 		return nil, NewHTTPError(http.StatusBadRequest, "invalid_message", fmt.Sprintf("unsupported message role %q", message.Role))
 	}
+}
+
+func chatCodexReasoningContinuation(message ChatMessage) (string, bool) {
+	if encrypted, ok := decodeProviderSignature(codexSignatureProvider, message.ReasoningSignature); ok {
+		return encrypted, true
+	}
+	if message.raw == nil {
+		return "", false
+	}
+	var details []map[string]any
+	if err := json.Unmarshal(message.raw["reasoning_details"], &details); err != nil {
+		return "", false
+	}
+	toolCallIDs := map[string]struct{}{}
+	if calls, ok := anySlice(message.ToolCalls); ok {
+		for _, rawCall := range calls {
+			call, _ := rawCall.(map[string]any)
+			if id, _ := call["id"].(string); id != "" {
+				toolCallIDs[id] = struct{}{}
+			}
+		}
+	}
+	for _, detail := range details {
+		id, _ := detail["id"].(string)
+		data, _ := detail["data"].(string)
+		if detail["type"] != "reasoning.encrypted" || id == "" {
+			continue
+		}
+		if _, exists := toolCallIDs[id]; !exists {
+			continue
+		}
+		if encrypted, ok := decodeProviderSignature(codexSignatureProvider, data); ok {
+			return encrypted, true
+		}
+	}
+	return "", false
 }
 
 func chatContentToCodex(content any, assistant bool) ([]any, error) {
--- a/backend/internal/server/codex_compat_response.go
+++ b/backend/internal/server/codex_compat_response.go
@@ -154,6 +154,15 @@ func codexResponsesToChat(
 	}
 	if len(toolCalls) > 0 {
 		message["tool_calls"] = toolCalls
+		if reasoningSignature != "" {
+			firstCall, _ := toolCalls[0].(map[string]any)
+			callID, _ := firstCall["id"].(string)
+			message["reasoning_details"] = []any{map[string]any{
+				"type": "reasoning.encrypted",
+				"id":   callID,
+				"data": reasoningSignature,
+			}}
+		}
 		if text.Len() == 0 {
 			message["content"] = nil
 		}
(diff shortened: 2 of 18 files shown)


astaxie (commented):
Reviewed the complete current diff at `a4ebeb53`. I found two blocking correctness/compatibility issues and one documentation contract gap; details are inline. The focused Codex/usage regressions, full `go test ./...`, `go vet ./...`, and `git diff --check` pass, and GitHub CI is green. Those checks do not cover the foreign-provider replay path or the legacy Responses usage keys described below, so this is a comment-only review and not an approval.

backend/internal/server/codex_compat_request.go
```diff
@@ -550,6 +550,42 @@ func chatMessageToCodex(message ChatMessage, toolNames map[string]string) ([]any
...
 
+func chatCodexReasoningContinuation(message ChatMessage) (string, bool) {
+	if encrypted, ok := decodeProviderSignature(codexSignatureProvider, message.ReasoningSignature); ok {
+		return encrypted, true
+	}
+	if message.raw == nil {
+		return "", false
+	}
+	var details []map[string]any
+	if err := json.Unmarshal(message.raw["reasoning_details"], &details); err != nil {
```
astaxie, line 561: [P1] Keep Codex reasoning details gateway-local

This accepts `reasoning_details` as a TokenHub/Codex continuation field, but the non-Codex path in `withoutGatewayExtensions` still strips only `reasoning_content`, `reasoning_signature`, and `redacted_reasoning_content`. I reproduced decoding a request containing this field and marshaling `withoutGatewayExtensions(req, false)`; the resulting OpenAI-compatible upstream body still contained both `reasoning_details` and `codex:opaque-state`. A later routed or failover request can therefore leak the opaque Codex continuation to a foreign provider, and strict upstreams can reject the unknown field. Please strip or provider-scope these details on non-Codex routes, as the legacy signature is, and add a foreign-route regression.

backend/internal/server/openai_usage.go
```diff
@@ -0,0 +1,91 @@
...
+		result["completion_tokens_details"] = outputDetails
+	}
+	return result
+}
+
+// openAIResponsesUsageObject renders TokenHub's internal accounting in the
+// public Responses shape while retaining TokenHub's compatibility aliases.
+func openAIResponsesUsageObject(usage Usage) map[string]any {
+	result := tokenHubUsageAliases(usage)
+	result["input_tokens"] = usage.PromptTokens
```
astaxie, line 32: [P1] Preserve the existing Responses usage totals

Before this change, `MockAdapter.Responses` and `responseObject` stored `Usage` directly, whose JSON included `prompt_tokens` and `completion_tokens`. This formatter replaces those keys with `input_tokens` and `output_tokens`, while `tokenHubUsageAliases` does not restore the old totals. Every synthetic Mock/Anthropic/Gemini `/v1/responses` result therefore silently drops fields that existing clients may read, contrary to the PR promise to retain TokenHub aliases and the repository compatibility rule. Please emit both legacy totals alongside the standard Responses names, or scope the normalization without removing the public keys, and add a route-level compatibility regression.

backend/internal/server/codex_compat_response.go
```diff
@@ -154,6 +154,15 @@ func codexResponsesToChat(
 	}
 	if len(toolCalls) > 0 {
 		message["tool_calls"] = toolCalls
+		if reasoningSignature != "" {
+			firstCall, _ := toolCalls[0].(map[string]any)
+			callID, _ := firstCall["id"].(string)
+			message["reasoning_details"] = []any{map[string]any{
```
astaxie, line 160: [P2] Document the new round-trip extension in all three guides

`message.reasoning_details` is now both emitted and accepted, with a tool-call ID binding that clients must preserve. The reasoning-continuation table in `docs/user-guide.md` and its Simplified Chinese and Japanese counterparts still lists only `reasoning_signature` and the other older extensions. Please add this field and its echo/binding semantics to all three guides so the documented client contract matches the new behavior.

backend/internal/server/codex_compat_request.go
astaxie: [P1] Preserve non-Codex/OpenRouter reasoning details

The Codex leak itself is fixed, but the sanitizer is now overbroad: `withoutGatewayExtensions` deletes every `reasoning_details` value on ordinary OpenAI-compatible and DeepSeek routes. This PR itself describes the field as an OpenRouter-compatible extension, and `ChatMessage.raw` previously preserved unknown upstream fields, so a client replaying legitimate non-Codex/OpenRouter reasoning details now silently loses provider continuation. Please remove only TokenHub-owned Codex encrypted details (for example, entries whose encoded data is provider-scoped to `codex`) instead of deleting the entire upstream extension. I am keeping this thread unresolved for that compatibility issue.

astaxie (approved):
Re-reviewed the complete current diff at 797f570f1. The latest change now removes only TokenHub-scoped Codex reasoning details while preserving legitimate non-Codex/OpenRouter entries, including mixed arrays, so the previous compatibility blocker is fixed without reintroducing opaque-state leakage. The prior usage-alias and three-language documentation findings remain fixed. Focused regressions, the full backend suite, go vet, all 122 repository tests, translation/environment/source-line gates, and diff checks pass; required GitHub CI is green and the branch is cleanly mergeable.

PR #253, opened 2026-08-16
perf(routing): batch route candidate lookups

legendtkl:
## Summary

Eliminate the `SelectRouteCandidates` provider and resource N+1 queries while preserving routing behavior and PostgreSQL read consistency.

## Related Issue

N/A.

## Changes

- Batch provider, explicit-resource, and implicit-resource lookups with bounded 500-ID `IN` query chunks.
- Preserve outer route order, validate explicit resource ownership, and define resource tie-breaking as `priority ASC, weight DESC, created_at ASC, id ASC`.
- Read PostgreSQL route candidates in a read-only repeatable-read transaction.
- Keep optional adaptive runtime-stat failures best-effort by rolling back the failed observation query to a savepoint.
- Add query-count, chunking, ordering, ownership, fallback, error-path, race, and PostgreSQL multi-instance regression coverage.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor or maintenance
- [ ] Documentation

## Verification

- `gofmt` on all changed Go files.
- `go test ./...`
- `go vet ./...`
- `go test -race ./internal/server -run '^TestSelectRouteCandidates' -count=1`
- `node --test tools/*.test.mjs` (122 tests)
- `git diff --check`
- PostgreSQL multi-instance tests compiled locally but were skipped because `TEST_POSTGRES_URL` was not configured; the new subtests are included in the existing CI PostgreSQL suite.
- `golangci-lint` was not run because it is unavailable in the local environment.

No public API, persistence schema, configuration, deployment, or user-visible behavior changes. Explicit resource ownership validation prevents a route from selecting a resource belonging to another provider. The change can be rolled back by reverting this PR.

## Checklist

### DIFF
--- a/backend/internal/server/store_route_candidates.go
+++ b/backend/internal/server/store_route_candidates.go
@@ -2,6 +2,8 @@ package server
 
 import (
 	"database/sql"
+	"errors"
+	"fmt"
 	"strings"
 	"time"
 
@@ -15,16 +17,22 @@ func (s *GormStore) SelectRouteCandidates(modelName string) ([]RouteSelection, e
 	defer s.mu.Unlock()
 
 	var selections []RouteSelection
-	read := func(db *gorm.DB) error {
-		var err error
-		selections, err = s.loadRouteCandidates(db, modelName, time.Now().UTC())
-		return err
+	now := time.Now().UTC()
+	runRead := func(load func(*gorm.DB, string, time.Time) ([]RouteSelection, error)) error {
+		read := func(db *gorm.DB) error {
+			var err error
+			selections, err = load(db, modelName, now)
+			return err
+		}
+		if s.dbDriver == "postgres" {
+			return s.db.Transaction(read, &sql.TxOptions{Isolation: sql.LevelRepeatableRead, ReadOnly: true})
+		}
+		return read(s.db)
 	}
-	var err error
-	if s.dbDriver == "postgres" {
-		err = s.db.Transaction(read, &sql.TxOptions{Isolation: sql.LevelRepeatableRead, ReadOnly: true})
-	} else {
-		err = read(s.db)
+	err := runRead(s.loadRouteCandidates)
+	var batchErr *routeCandidateBatchLookupError
+	if errors.As(err, &batchErr) {
+		err = runRead(s.loadRouteCandidatesIndividually)
 	}
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


backend/internal/server/store_route_candidates.go
```diff
@@ -0,0 +1,182 @@
...
+	if err := db.Where("model_name = ? AND status = ?", modelName, StatusActive).
+		Order("priority asc, weight desc, created_at asc").
+		Find(&routes).Error; err != nil {
+		return nil, err
+	}
+
+	providerIDs, explicitResourceIDs, implicitProviderIDs := routeCandidateLookupIDs(routes)
+	providers, err := loadRouteCandidateProviders(db, providerIDs)
+	if err != nil {
+		return nil, err
```
astaxie: [P1] Preserve failover when one candidate lookup errors

These batch-load errors now abort all candidate selection. On the base implementation, provider and explicit-resource lookup errors are scoped to one route and `continue`, so later healthy routes remain available. I verified this with the same regression test on both revisions: when only the first provider lookup fails transiently, base `5b6024610` selects the second provider, while this head returns the transient error. That violates the PR's routing-behavior preservation requirement and can turn one failed lookup batch into a model outage. Please fall back to per-route/per-item recovery after a batch error, or otherwise retain unaffected candidates.

backend/internal/server/store_route_candidates_test.go
```diff
@@ -0,0 +1,293 @@
...
+	store.AddRoute(ModelRoute{
+		ID: "route_lookup_error", ModelName: modelName, ProviderID: provider.ID,
+		ProviderModel: modelName, Priority: 1, Weight: 100, Status: StatusActive,
+	})
+
+	wantErr := errors.New("provider batch lookup failed")
+	callbackName := "test:route-candidate-query-error"
+	if err := store.db.Callback().Query().Before("gorm:query").Register(callbackName, func(tx *gorm.DB) {
+		if tx.Statement.Schema != nil && tx.Statement.Schema.Table == "providers" {
+			tx.AddError(wantErr)
```
astaxie: [P1] Handle AddError so the required Backend check passes

The required Backend CI job is failing because `errcheck` reports this ignored `tx.AddError` result (`store_route_candidates_test.go:227`, run 31951811077 / job 95176322164). Please handle or explicitly validate the returned error; until then the aggregate CI check also remains failed.

astaxie (commented):
Reviewed the complete current diff at `2e6d57e92`. Focused routing tests, the full backend test suite, `go vet ./...`, and `git diff --check` pass locally, but I found a reproduced failover regression and the required Backend CI check is failing; details are inline. This is a comment-only review and not an approval.

backend/internal/server/store_route_candidates.go
legendtkl: Addressed in 6e321005. Batch provider and explicit-resource lookup errors now trigger a per-route fallback, preserving later healthy candidates. PostgreSQL retries in a new read-only repeatable-read transaction and isolates each best-effort item lookup with savepoints. Added SQLite and PostgreSQL regressions covering both lookup families with a failed batch and failed first item while the unaffected route still succeeds.

backend/internal/server/store_route_candidates_test.go
legendtkl: Addressed in 6e321005. The AddError return values are now handled explicitly, and golangci-lint v2.12.2 reports 0 issues locally.

astaxie (approved):
Re-reviewed the complete current diff at 6e321005e. The prior failover and errcheck issues are fixed: failed batch reads retry in a fresh transaction and PostgreSQL item lookups use savepoints. Full backend tests, go vet, repository gates, and diff checks pass; required CI is green. Approving this exact head.

PR #257, opened 2026-08-16
fix(security): harden administrator authentication and secrets

deepjerry-ai:
## Summary

This PR closes multiple administrator-session and credential exposure paths. It replaces bearer tokens in OAuth redirects with short-lived, single-use PKCE-bound codes, constrains browser session persistence, adds nonce-based Content Security Policy enforcement, and prevents sensitive administrator resources from leaking through API, audit, alert-delivery, or CSV responses.

It also fixes related issues found during review: OAuth username account takeover, explicitly unverified email linking, password-reset tokens in query strings, cross-instance reset-token replay, and the inability to explicitly clear stored integration credentials.

## Related Issue

N/A.

## Changes

- Return a 60-second, single-use administrator OAuth exchange code in the URL fragment and atomically persist/consume OAuth state and exchange records across instances.
- Require S256 PKCE, exact return-origin validation, trusted forwarded-host handling, stable redirect error codes, verified email claims when supplied, and email-only existing-account linking.
- Keep administrator bearer sessions in `sessionStorage`, remove legacy `localStorage` sessions, and reject bearer tokens supplied through URL parameters.
- Add per-request CSP nonces and security headers; propagate trusted proxy headers safely in the multi-instance Nginx configuration.
- Redact identity-provider and notification-channel credentials from API responses, audit snapshots, alert deliveries, and CSV exports while preserving masked update semantics and supporting explicit JSON `null` clearing.
- Enforce database-level role, project, and API-key scoping for image-generation audit records before pagination and decryption.
- Move password-reset tokens to URL fragments, derive reset links from trusted console origins, and atomically consume reset tokens across instances.
- Remove the committed SDK credential fallback and document the hardened OAuth, session, proxy, and credential-management behavior in English, Simplified Chinese, and Japanese.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor or maintenance
- [x] Documentation

## Verification

- [x] `cd backend && go test ./...` (all packages passed; `internal/server` completed in 147.180s)
- [x] `cd frontend && npm test` (64 domain tests and 5 component tests passed)
- [x] `cd frontend && npm run typecheck`
- [x] `cd frontend && npm run lint`
- [x] `cd frontend && npm run build` (Next.js 16.2.9 production build passed)
- [x] Production response checks for `/`, `/overview`, and a 404 route: six unique CSP nonces, every script nonce matched its response policy, and `script-src` contained neither `unsafe-inline` nor `unsafe-eval`
- [x] Password-reset cross-instance concurrency regression repeated 10 times with exactly one successful consumer per token
- [x] `node --test tools/*.test.mjs` (123 tests passed)
- [x] Documentation and UI translation checks, environment contract, source-line gate, Compose rendering, and `git diff --check`
- [x] Added-line credential scan reported no high-severity findings; medium findings were deterministic test fixtures and reserved example data
- [x] `cd frontend && npm run test:e2e` (3 Chromium smoke tests passed against the isolated frontend/backend/provider stack)
- [ ] SDK live smoke tests were not run because compatible provider credentials were not available

AI-assessed test coverage is 88%. The remaining non-blocking gaps are browser-level OAuth/session E2E coverage, a committed response-level CSP test, explicit log-capture validation, PostgreSQL reset-token contention coverage, and an automated SDK secret-scan regression.

- The bundled console and backend must be deployed together because administrator OAuth now requires the PKCE exchange endpoint and no longer accepts bearer tokens from redirect URLs.
- Closing a browser tab now ends its administrator session locally; reloading the same tab preserves it.
- Sensitive field values `""`, `"********"`, and `"[redacted]"` retain the stored credential during updates. JSON `null` explicitly clears the credential and related legacy aliases.
- Identity-source mutations are now restricted to platform administrators. Security administrators retain masked read access but can no longer create, update, or delete identity sources; migrate those workflows to platform administrators before rollout.
- Reverse proxies in `TOKENHUB_TRUSTED_PROXY_CIDRS` must overwrite `X-Forwarded-For`, `X-Forwarded-Host`, and `X-Forwarded-Proto` instead of passing client values through.
- Revoke and rotate the credential that was previously committed as the SDK fallback. Invalidate existing administrator sessions during rollout because a code change cannot revoke already issued credentials or sessions.
- Residual risks: `sessionStorage` remains readable by a successful same-origin XSS or malicious extension; identity-provider and notification-channel credentials are redacted at response boundaries but are not yet encrypted at rest. Follow-up work should move administrator sessions to an HttpOnly cookie/refresh design and encrypt these resource fields.
- Rollback is a normal application rollback, but doing so would restore the vulnerable OAuth/session behavior and is not recommended after credential/session rotation.

## Checklist

- [x] Environment variable changes are synchronized across examples, Compose, `start.sh`, and deployment documentation where applicable. No environment variables were added or removed.
- [x] `data/model-catalog.yaml` remains tracked and was not changed by this PR.


backend/internal/server/store_admin.go
```diff
@@ -505,30 +505,33 @@ func (s *GormStore) ResetAdminUserPassword(token string, password string) (Admin
...
-	}
-	if item.UsedAt != nil || time.Now().UTC().After(item.ExpiresAt) {
-		return AdminUser{}, NewHTTPError(400, "invalid_reset_token", "Reset token is invalid or expired")
-	}
-	var user AdminUser
-	if err := s.db.First(&user, "id = ?", item.UserID).Error; err != nil {
-		return AdminUser{}, notFound(err, "admin_user_not_found", "Admin user not found")
-	}
-	now := time.Now().UTC()
 	passwordHash, err := hashPassword(password)
```
astaxie: [P1] Validate the reset token before paying the bcrypt cost

This endpoint is unauthenticated, but the new ordering holds the store-wide mutex and runs bcrypt before checking whether the reset token exists. I measured eight requests with the same invalid token taking 358 ms on this head; the base implementation rejected an invalid token with the indexed lookup before hashing. Concurrent random-token POSTs can therefore consume bcrypt CPU and queue every operation that needs this mutex, turning the reset endpoint into a cheap denial-of-service path. Please perform a cheap valid/unexpired-token lookup before hashing (while retaining the conditional update as the final atomic claim), and avoid holding the global mutex during attacker-controlled bcrypt work.

backend/internal/server/admin_auth_http.go
```diff
@@ -157,54 +154,126 @@ func (s *Server) handleAdminOAuthStart(w http.ResponseWriter, r *http.Request) {
 		writeError(w, r, err)
 		return
 	}
+	redirectTarget, _ := url.Parse(redirectURI)
+	cookieSecure := redirectTarget != nil && strings.EqualFold(redirectTarget.Scheme, "https")
+	if err := s.store.SaveAdminOAuthFlow(adminOAuthFlow{
```
astaxie, line 164: [P2] Bound unauthenticated OAuth flow persistence

Every valid GET to the public OAuth start endpoint inserts a fresh row here before the caller proves browser ownership or completes an IdP flow. There is no rate limit, per-browser replacement, or live-row cap; expired rows are only opportunistically deleted after their ten-minute lifetime. I reproduced 200 unauthenticated start requests leaving 200 live admin_o_auth_flow_records. An external caller can continuously grow the primary database and create shared SQLite write contention without following redirects or retaining the cookies. Please bound outstanding flows (for example per source/provider/browser plus a global cap or cluster-aware rate limit) and add an abuse regression.

deploy/nginx.multi-instance.conf
```diff
@@ -29,6 +29,7 @@ server {
         proxy_http_version 1.1;
         proxy_set_header Host $host;
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_set_header X-Forwarded-Host $host;
```
astaxie, line 26: [P2] Make the bundled proxy overwrite every trusted forwarded header

This block still preserves client input: X-Forwarded-For uses $proxy_add_x_forwarded_for, and tokenhub_forwarded_proto accepts $http_x_forwarded_proto directly. The backend now treats forwarded host/proto from a trusted proxy as authoritative for OAuth callback construction, while the PR body and all three documentation variants explicitly require trusted proxies to overwrite X-Forwarded-For, X-Forwarded-Host, and X-Forwarded-Proto. The bundled multi-instance proxy should generate these values from trusted connection data (and apply the same fix to the frontend block) rather than pass through request headers; please add a configuration regression for spoofed incoming values.

docs/administrator-guide.md
```diff
@@ -326,6 +326,10 @@ Creating an identity source uses three required steps: choose the source, enter
 
 Use the public TokenHub backend URL with the callback path `/api/admin/auth/oauth/callback`. You may leave Callback URL blank to derive it from the incoming backend host; when setting it explicitly, the complete URL must exactly match the redirect URL registered with the identity provider.
 
+Completing an administrator OAuth login never puts the administrator session token in a redirect URL. TokenHub returns a short-lived, single-use code to the console, which exchanges it once and keeps the resulting session only in the current browser tab. Reloading that tab preserves the session; closing it requires signing in again.
+
```
astaxie, line 330: [P2] Document the complete credential-management contract

The PR promises synchronized documentation for the hardened credential-management behavior, but this addition only covers identity-source client secrets. It does not tell operators that notification-channel credentials and delivery URLs are also redacted, nor how updates distinguish retain from clear: empty string, the mask, and [redacted] retain the stored value, while JSON null clears it and related aliases. Please document those behaviors here and in the Simplified Chinese and Japanese guides so administrators can safely rotate or remove these credentials without relying on the PR description.

astaxie (commented):
Reviewed the complete current diff at d7d6524ce. Full backend tests, go vet, frontend lint/typecheck/tests/build, Chromium E2E, repository gates, Compose rendering, and diff checks pass locally, and required CI is green. I found one unauthenticated resource-exhaustion regression and three additional security/operations gaps inline, so this is a comment-only review.

backend/internal/server/admin_auth_http.go
deepjerry-ai: Fixed in 189031e. OAuth start admission is now serialized in the shared database: PostgreSQL uses a transaction-scoped advisory lock, while SQLite acquires its writer lock before counting. Client sources are normalized to IPv4 /24 or IPv6 /64 and stored only as a SecretKey-backed HMAC. Active limits are 16 per client scope/provider, 32 per client scope across providers, and 4096 globally. Rejections return 429 oauth_start_rate_limited with Retry-After and do not set a cookie or Location. Tests cover untrusted X-Forwarded-For input, cross-provider scope limits, concurrent last-slot admission across two SQLite stores, and committed expired-row cleanup on rejection; the concurrency set passed 20 consecutive runs.

backend/internal/server/store_admin.go
deepjerry-ai: Fixed in 549daba. The reset path now performs an indexed valid/unexpired token preflight before password hashing, runs bcrypt outside the store-wide mutex, and retains the conditional UPDATE as the final atomic claim. Regression coverage verifies that unknown and expired tokens never invoke the hasher, hashing does not block another store operation, and two database instances still produce exactly one successful reset. The focused tests passed with -count=10, and go test ./... plus go vet ./... pass locally.

deploy/nginx.multi-instance.conf
deepjerry-ai: Fixed in a2a5e11. Both backend and frontend locations now overwrite X-Forwarded-For with $remote_addr, X-Forwarded-Host with $host, and X-Forwarded-Proto with $scheme. The client-derived proto map and $proxy_add_x_forwarded_for are removed. A repository test rejects either client-forwarded variables or append behavior, and the remote-PostgreSQL Compose configuration renders successfully with required placeholder values.

docs/administrator-guide.md
deepjerry-ai: Fixed in 7965826. The English, Simplified Chinese, and Japanese administrator guides now document identity-source and notification-channel redaction across management API responses, CSV exports, audit snapshots, and alert-delivery output. They also define the complete update contract: empty string, the mask, and [redacted] retain a stored value; JSON null clears it and removes related notification-channel aliases. The full translation co-change check passes against origin/main.

backend/internal/server/admin_auth_http.go
```diff
@@ -402,85 +493,115 @@ func firstForwardedValue(value string) string {
...
+	requestOrigin, valid := requestOAuthOrigin(r)
+	if valid && targetOrigin == requestOrigin {
+		return true
 	}
-	body := base64.RawURLEncoding.EncodeToString(data)
-	mac := hmac.New(sha256.New, []byte(s.oauthStateSecret()))
-	_, _ = mac.Write([]byte(body))
-	signature := base64.RawURLEncoding.EncodeToString(mac.Sum(nil))
-	return body + "." + signature, nil
+	return valid && isOAuthLoopbackHost(target.Hostname()) && isOAuthLoopbackHost(requestHostName(r.Host))
```
astaxie: [P1] Keep loopback returns on an exact origin

This fallback accepts every scheme/host/port combination whenever both hostnames are loopback. A page served from `http://localhost:4444` can navigate a victim to the OAuth start endpoint on `127.0.0.1:8080` with its own PKCE challenge and return URL; after the victim completes SSO, that local origin receives the code and already owns the verifier, so it can mint the victim administrator session. This also contradicts the stated exact return-origin validation. Please require an exact configured CORS/public origin for loopback too; local development can enumerate the intended console origins explicitly.

backend/internal/server/admin_resource_secrets.go
```diff
@@ -0,0 +1,204 @@
...
+	return cloned
+}
+
+func isAdminResourceSecretPlaceholder(value any) bool {
+	text, ok := value.(string)
+	if !ok {
+		return false
+	}
+	text = strings.TrimSpace(text)
+	return text == "" || text == providerHeaderMask || text == "[redacted]"
```
astaxie: [P2] Honor the placeholder documented in the PR contract

The Compatibility section says the ASCII value `********` retains a stored credential, but this predicate accepts only the empty string, the bullet mask `••••••••`, and `[redacted]`. A client or operator following the PR contract will silently replace the real credential with eight asterisks. Please either accept the documented ASCII placeholder here or correct the PR body so the update contract is unambiguous.

backend/internal/server/admin_identity_oauth_flow.go
```diff
@@ -0,0 +1,392 @@
...
+	if exchange.CreatedAt.IsZero() {
+		exchange.CreatedAt = now
+	}
+	record := adminOAuthExchangeRecord{
+		ID:            NewID("oauth_exchange"),
+		CodeHash:      HashSecret(exchange.Code),
+		CodeChallenge: exchange.CodeChallenge,
+		UserID:        exchange.UserID,
+		CreatedAt:     exchange.CreatedAt,
+		ExpiresAt:     exchange.CreatedAt.Add(adminOAuthExchangeTTL),
```
astaxie: [P2] Base the exchange expiry on the database clock

The issuing replica derives `ExpiresAt` from its local `time.Now()`, while `ConsumeAdminOAuthExchange` compares it with the consuming replica local clock. With ordinary cross-node clock skew, a fresh code can expire immediately or remain usable longer than the promised 60 seconds. Because this record is explicitly consumed across instances, please derive and compare the TTL using the shared database clock inside the transactions.

frontend/features/admin/domain/password-reset.ts
```diff
@@ -0,0 +1,21 @@
...
+  }
+  if (fragment.has("reset_token")) {
+    fragment.delete("reset_token");
+    target.hash = fragment.toString();
+    changed = true;
+  }
+  if (changed) {
+    replaceURL(`${target.pathname}${target.search}${target.hash}`);
+  }
+  return fragmentToken || queryToken;
```
astaxie: [P2] Reject reset credentials in the query string

This fallback still accepts `reset_token` from the query when no fragment token exists. `replaceState` runs only after the initial document request, so the credential has already reached the Next.js server and any proxy/access logs. That leaves legacy or manually constructed links on the exposure path this PR says it removes. Please reject query tokens rather than consuming them; only the fragment value should be returned.

backend/internal/server/admin_authorization.go
```diff
@@ -1248,7 +1248,7 @@ func adminResourcePermission(path string) string {
 		return "security"
 	}
 	if strings.Contains(path, "/identity-providers") {
-		return "security"
+		return "identity_provider"
```
astaxie, line 1251: [P2] Account for the security-admin write regression

Before this remapping, identity-provider routes used the `security` permission, so `security_admin` could create, patch, and delete them. The new `identity_provider` permission appears only in the read branch above, making every existing security-admin mutation return 403. If platform-admin-only management is intentional, this breaking RBAC change needs to be called out in the PR compatibility/rollout notes; otherwise preserve the existing write permission while returning masked fields.

backend/internal/server/admin_identity_oauth_flow.go
deepjerry-ai: Fixed in e5c2fa0. OAuth flow and exchange records now obtain CreatedAt and ExpiresAt from databaseNow inside their write transactions, and both consumption paths compare expiry against the database clock inside the locking transaction. This also closes the same skew issue for the ten-minute flow record. SQLite tests simulate a database clock one hour behind the application clock and verify creation plus consumption across two store instances. The existing PostgreSQL multi-instance CI test now exercises the same shared-clock assertions; it compiled locally and skipped because TEST_POSTGRES_URL is not configured. The full backend suite and go vet pass locally.

backend/internal/server/admin_auth_http.go
deepjerry-ai: Fixed in e5c2fa0 and documented for local development in 5ccadbf. Loopback return URLs now follow the same exact-origin rule as every other return: they must match an explicit CORSAllowedOrigins entry or PublicBaseURL, including scheme, host spelling, and port. An unconfigured loopback candidate is rejected and falls back to the backend origin without returning the exchange code to the caller-controlled page. Regression coverage includes unconfigured loopback, localhost/IPv4/IPv6 aliases, scheme and port changes, and exact configured CORS/public origins; the focused tests passed repeatedly and go test ./... passes.

frontend/features/admin/domain/password-reset.ts
deepjerry-ai: Fixed in 2430188. The frontend now returns a reset credential only from the URL fragment. A reset_token in the query is removed from the visible URL but ignored even when no fragment token exists; when both are present, only the fragment value is consumed. The focused domain tests, the full 64-test domain suite, 5 component tests, production build, and 3 Chromium smoke tests pass.

backend/internal/server/admin_authorization.go
deepjerry-ai: Addressed in the PR Compatibility, Security, and Operations section. The platform-admin-only mutation boundary is intentional: security administrators retain read access to masked identity-source configuration but can no longer create, update, or delete identity sources. The PR now explicitly identifies this as a rollout change and instructs operators to migrate those workflows to platform administrators before deployment. The synchronized administrator guides already document the same role boundary.

backend/internal/server/admin_resource_secrets.go
deepjerry-ai: Fixed in dd42b28. The update predicate now accepts the documented ASCII mask ******** in addition to the empty string, the Unicode mask, and [redacted]. The API regression patches both identity-source and notification-channel credentials and verifies stored secrets are retained, while a table test rejects real replacement values as placeholders. The English, Simplified Chinese, and Japanese administrator guides now state the same four retain values.

backend/internal/server/admin_auth_http.go
```diff
@@ -135,19 +125,26 @@ func (s *Server) handleAdminOAuthStart(w http.ResponseWriter, r *http.Request) {
...
 	}
-	redirectURI, err := identityProviderRedirectURI(provider, r)
+	codeChallenge, err := validateAdminOAuthPKCE(
+		r.URL.Query().Get("code_challenge"),
+		r.URL.Query().Get("code_challenge_method"),
+	)
 	if err != nil {
 		writeError(w, r, err)
 		return
 	}
```
astaxie, line 135: [P1] Forward the PKCE challenge to the identity provider

The start endpoint validates and persists `codeChallenge`, but this call still builds the provider authorization URL with only `state`; `buildOAuthAuthorizeURL` never adds `code_challenge` or `code_challenge_method`. Consequently the IdP authorization-code hop is not PKCE-bound even though this PR requires S256 PKCE. An intercepted authorization code could be redeemed at the provider before TokenHub applies its own exchange-code PKCE. Please pass the challenge through this helper (including the provider-specific URL builders) and add a regression asserting the redirect `Location` contains both PKCE parameters.

deepjerry-ai: Fixed in bddb28f.

Rather than forwarding the console challenge (whose verifier only exists in browser memory, so a PKCE-enforcing provider would then reject our own confidential-client token exchange), the backend now generates an independent S256 PKCE pair per flow. The challenge is sent to the provider in every authorize redirect — `buildOAuthAuthorizeURL` sets `code_challenge` + `code_challenge_method=S256`, and the DingTalk, Feishu, and WeCom rewrites preserve both parameters. The verifier is persisted in the flow record (schema column added, required at save time) and presented as `code_verifier` at the generic OIDC token endpoint in both the client_secret_post and client_secret_basic paths. The DingTalk/Feishu/WeCom token APIs expose no verifier parameter, so their exchanges are unchanged; the challenge is still sent for forward compatibility.

Regression coverage: `TestAdminOAuthAuthorizationIsPKCEBoundToIdentityProvider` runs the full start → callback → exchange login against a fake provider and asserts the redirect Location carries a valid independent S256 challenge, the token endpoint receives a `code_verifier` whose S256 digest equals that challenge, and the console exchange still completes with the original browser verifier. The three protocol tests additionally assert both PKCE parameters survive the provider-specific URL rewrites.

Validation: focused OAuth/provider suites, `go test ./internal/server` three consecutive clean runs, `go test ./...` (one unrelated flake in a run whose truncated output pointed at the response-jobs worker; it did not reproduce in three re-runs), and `go vet ./...` pass.

frontend/features/admin/domain/password-reset.ts
deepjerry-ai: Fixed in 2430188 (apologies for the missing reply on this thread). `parsePasswordResetLocation` no longer falls back to `reset_token` in the query string; only the fragment value is accepted, so query-carried credentials are ignored rather than consumed. `password-reset.test.mjs` covers the rejection.

astaxie (commented):
Re-reviewed the complete current diff. The upstream PKCE gap is fixed for generic identity providers with an independent S256 pair persisted in the flow and presented at token exchange; DingTalk, Feishu, and WeCom correctly omit an unsupported challenge while the browser-to-TokenHub exchange remains S256-bound. The query-string reset-token fallback is also removed. Focused OAuth, vendor-protocol, and password-reset tests pass locally, and Backend/PostgreSQL/Frontend/Deployment/Repository/aggregate CI checks pass on this head. I am not approving because Frontend browser smoke is CANCELLED on the current head; cancelled CI remains a merge blocker.

astaxie (approved):
Reviewed the complete current diff at `b19359e1`. The new CI change uses the runner-provided Chrome through a validated Playwright channel and fixes the previously cancelled browser smoke check. Local `go test ./...`, `go vet ./...`, frontend lint/typecheck/unit/build, Chrome E2E (3/3), all 125 repository tests, translation/environment/source-line gates, and diff checks pass. All required GitHub checks are green; the optional LiteLLM migration E2E job is skipped by design. No unresolved substantive blocker remains, so I approve this current head.

PR #263, opened 2026-08-17
feat(quota): enforce aggregate user limits

### LINKED ISSUE #259: feat(quota): add per-user aggregate usage limits
## Affected area

Authentication and API keys

## Problem or use case

TokenHub supports daily and monthly request, token, and cost limits, but quota counters are currently maintained per API key. A quota policy applied at project or team scope still gives each matching API key an independent allowance.

API keys can be attributed to an owner through `owner_user_id`, with fallback attribution through the key creator or project owner. However, this attribution is used for usage reporting and audit records rather than quota enforcement. As a result, one user can consume a separate daily token allowance through each of their API keys, and rotating or creating another key effectively resets the user's available allowance.

Enterprise administrators need to configure a daily token limit for a user and have usage aggregated across every API key attributed to that user. This is required for fair-use controls, predictable cost governance, and consistent enforcement when users own multiple project keys.

## Proposed solution

Add a `user` quota-policy scope that applies the existing quota fields to an attributed user, including at least `daily_tokens` and preferably the complete existing `QuotaLimits` surface.

- Accept `scope: "user"` with an administrator user ID in `scope_id` through the Admin API.
- Expose the user scope and a user selector in the admin console.
- Resolve the attributed user consistently with usage accounting: API-key owner first, then the existing key-creator/project-owner fallback.
- Aggregate counters across every API key attributed to the same user instead of maintaining independent per-key allowances.
- Combine user limits with applicable API-key, project, team, and global policies using the existing strictest-limit behavior.
- Enforce admission atomically across concurrent requests and multiple API keys. PostgreSQL replicas must share the same counters; SQLite should retain its supported single-backend behavior.
- Reconcile reserved and actual token usage exactly once for buffered, streaming, background Responses, image, and other metered gateway calls.
- Preserve usage and quota continuity when an API key is rotated, revoked, or replaced.
- Include the effective scope in quota diagnostics, audit records, alerts, and bounded-cardinality metrics without exposing key secrets or sensitive identity data.

### Issue one API key per user

This approximates a per-user quota only while the user has exactly one key. It breaks when a user needs separate keys for environments or projects, and key rotation can unintentionally reset the allowance.

### Apply a project quota policy

Project policies currently apply their limit independently to each API key, so they do not provide an aggregate user allowance. Users may also own keys in multiple projects.

### Enforce the limit outside TokenHub

An external identity-aware proxy could maintain user counters, but it would duplicate TokenHub's usage reconciliation, streaming behavior, background job handling, audit, and multi-instance coordination.

## Additional context

Suggested acceptance criteria:

- [ ] An administrator can create, update, disable, and delete a user-scoped quota policy through the Admin API and console.
- [ ] `daily_tokens` is enforced against the total metered usage of all API keys attributed to the user.
- [ ] Key rotation or replacement does not reset the user's current daily or monthly counters.
- [ ] The strictest applicable user, API-key, project, team, and global limit is enforced.
- [ ] Concurrent calls through different keys cannot independently pass the same remaining user allowance.
- [ ] A rejected call returns `429 quota_exceeded` before invoking an upstream provider and records which non-sensitive quota scope blocked admission.
- [ ] Usage settlement is idempotent for buffered, streaming, interrupted-stream, background Responses, and image jobs.
- [ ] PostgreSQL multi-instance and SQLite single-instance behavior is covered by automated tests.
- [ ] The console shows user-scope configuration and current consumption without exposing complete API keys.
- [ ] English, Simplified Chinese, and Japanese documentation and UI translations are updated.

## Submission checklist

- [x] I searched open and closed issues for related requests.
- [x] The issue title and all explanatory text are written in English.
- [x] I removed or redacted all credentials and other secrets.

astaxie:
## Summary

Adds aggregate user-scoped quota policies so one user's request, token, cost, rate, and concurrency limits apply across all API keys attributed to that user. This closes the per-key bypass caused by multiple keys and key rotation, while preserving ownership history when a key changes owner within one billing period.

## Related Issue

Closes #259

## Changes

- Add `scope: "user"` quota policies with API-key owner, key creator, and project owner attribution fallback.
- Enforce aggregate daily/monthly request, token, cost, RPM/TPM, and concurrency limits across keys, including atomic PostgreSQL multi-instance admission.
- Store API-key quota totals in canonical rows and owner-specific history in attribution rows, so owner transfers preserve each user's usage.
- Migrate legacy quota buckets to an explicit unattributed canonical marker; never assign unverifiable history to the current owner.
- Commit image admission, concurrency leases, and image job creation in one database transaction; persist response/image admission state for recovery.
- Preserve settlement and reservation correctness for buffered, streaming, image, and background Responses calls.
- Extend the admin API and console with user selection, current request/token/cost usage, lifecycle controls, and team-leader authorization.
- Update English, Simplified Chinese, and Japanese administrator documentation and UI translations.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Refactor or maintenance
- [x] Documentation

## Verification

- `gofmt -w` on changed Go files
- `go test ./...`
- `go vet ./...`
- `npm run lint`
- `npm run typecheck`
- `npm test`
- `npm run build`
- `npm run test:e2e`
- `git diff --check`
- Added regression coverage for same-bucket owner transfers, legacy-safe attribution, persisted image admission rollback, and transactional background admission.
- SDK smoke tests were skipped because no compatible backend credentials/environment were configured; deployment checks were skipped because this PR does not change deployment files.

The existing OpenAI-compatible `/v1` contract remains unchanged; user-scoped policies are additive and existing scopes retain their strictest-limit behavior. This change updates the frozen database adoption baseline: `quota_buckets` now uses `(key_id, scope, bucket, attributed_user_id)` as its primary key, legacy rows are retained under an explicit unattributed marker rather than guessed ownership, and image/response admission recovery fields are persistent. Pre-ledger databases are upgraded under the shared schema lock before verification. The attribution column defaults to the canonical unattributed marker so verified N-1 releases can continue writing during the rollback window; operators should still preserve the database and take the usual backup before rollout.

Image admission and job creation now commit together, so a crash cannot leave a reservation or concurrency lease without a recoverable job. User metrics and quota errors expose only bounded scope labels, never raw user IDs or secrets. Team leaders can configure only user policies within their own teams; platform administrators retain global access. Disabling or deleting a policy provides an application-level rollback without changing stored API keys.

## Checklist

- [x] Tests were added or updated for behavior changes, including cross-key, owner-transfer, concurrency, settlement, lifecycle, migration, and PostgreSQL coverage.


backend/internal/server/store_backup_quota.go
```diff
@@ -408,6 +408,106 @@ func (s *GormStore) quotaBucketForUpdate(tx *gorm.DB, keyID, scope, bucket strin
...
+			mergeQuotaCounterMax(period.counter, aggregated)
+		}
+	}
+	return usage, true, nil
+}
+
+func (s *GormStore) aggregateUserQuotaCounter(tx *gorm.DB, userID string, scope string, bucket string) (QuotaCounter, error) {
+	var aggregate QuotaCounter
+	creatorExpression := apiKeyCreatorExpression(s.dbDriver)
+	attributionExpression := "COALESCE(NULLIF(TRIM(ak.owner_user_id), ''), NULLIF(TRIM(" + creatorExpression + "), ''), NULLIF(TRIM(p.owner_user_id), ''))"
```
astaxie: [P1] Preserve request-time attribution when aggregating historical usage. This query reconstructs every quota bucket owner from the API key and project current owner fields. I reproduced the failure on this head: after user A consumed 10 tokens and the key owner was changed to user B, GetQuotaPolicyUsage for B reported those 10 historical tokens even though B made no request, while A dedicated user bucket still retained the same 10 tokens. The same usage is therefore charged to both users, and a pre-policy reassignment can move it away from the original user entirely. Aggregate from persisted request-time attribution, or persist attribution alongside the quota bucket, and add a key/project ownership-change regression.

astaxie (commented):
Reviewed the complete current diff at head cde69dd97bf515f43df619643bfb5949d1e59e66. Local full Go tests and go vet, frontend lint/typecheck/tests/build, all 122 repository tests, translation/UI/environment/source-line gates, and git diff --check passed. The required Backend and aggregate CI checks are still failing because four newly added helper functions are unused. I also reproduced one blocking attribution-continuity issue inline, so this is a comment-only review.

backend/internal/server/store_backup_quota.go
astaxie: The fix is incomplete on `a20bd517`. `QuotaBucket` still has one `attributed_user_id` for the `(key_id, scope, bucket)` primary key, and `quotaBucketForUpdate` only fills it while blank. I reproduced a same-day transfer: user A consumed 3 tokens, the key moved to user B, B consumed 7 before a B policy existed, then `GetQuotaPolicyUsage(B)` returned 0 and the 7-token policy did not block B. The existing regression verifies only that A keeps pre-transfer usage; it does not verify B receives post-transfer usage. Please model attribution at request granularity or split durable counters by attributed user, with a same-period owner-change regression. Pre-upgrade rows with blank attribution also need an explicit backfill/reconstruction path rather than assigning the entire row to the first post-upgrade owner.

astaxie (commented):
Re-reviewed the complete diff at `a20bd517`. All current GitHub checks are green, and local backend tests/vet, frontend lint/typecheck/tests/build, 125 repository tests, explicit translation/UI/environment/source-line gates, and `git diff --check` pass. The prior attribution finding is only partially fixed (I replied on that thread with a failing same-period transfer reproduction), and the following additional issues still block approval.

[P1] Make image admission atomic with durable image-job persistence (`backend/internal/server/image_generation.go:95-109,241-255`, `backend/internal/server/store_routing_calls.go:450-462`). `startImageCall` commits quota reservations and concurrency leases before `CreateImageJob` runs in a separate transaction. A process crash between those calls leaves no image-job row for `FailUnfinishedImageJobs` to discover, so request/token reservations remain charged and leases have no durable owner to recover. Persist the image job and its admission in one transaction, as the background Responses path does, and add a crash/restart regression.

[P1] Backfill existing current-period quota buckets safely (`backend/internal/server/store_backup_quota.go:395-425,479-486`). AutoMigrate adds `attributed_user_id`, but existing rows start blank. The first post-upgrade request stamps the whole pre-existing day/month row with the current owner, silently assigning all earlier requests, tokens, and cost to that user; untouched rows are ignored by user aggregation. Reconstruct attribution from request-time records or define an explicit conservative rollout path, with upgrade coverage for ownership changes.

[P2] Correct the PR's database and rollback impact. The Compatibility section says no database migrations are required, but this head adds persisted columns to `quota_buckets`, `image_jobs`, and `response_jobs`, and changes key/project deletion to retain attributed quota history (`backend/internal/server/store.go:18-23`, `backend/internal/server/types.go:518-588`, `backend/internal/server/store_projects.go:187-196,684-696`). Document the automatic schema changes, retained user-attribution data, upgrade/backfill behavior, and rollback compatibility.

[P2] Document team-leader authorization in all three administrator guides (`docs/administrator-guide.md:73` and locale counterparts). The docs say platform administrators configure user quotas, while the frontend exposes quota policies to team leaders and the backend lets them manage policies for users in their own team (`frontend/features/admin/core/navigation.tsx:210-216`, `backend/internal/server/admin_authorization.go:1034-1108`). Describe that scoped capability and its restrictions consistently, or remove the access.

astaxie (commented):
Found one remaining startup blocker in the PostgreSQL upgrade path:

[P1] Add the new column before the PostgreSQL backfill (`backend/internal/server/store_database.go:378-380`). `migrate()` invokes `ensureQuotaBucketAttributionSchema` before `db.AutoMigrate` (`:214-232`). On a database created by the base revision, `quota_buckets` has only the original `(key_id, scope, bucket)` columns, so this branch executes `UPDATE quota_buckets SET attributed_user_id ...` against a column that does not exist and aborts startup before `AutoMigrate` can add it. The SQLite branch handles the missing column explicitly, but the PostgreSQL branch does not. Please add/guard `ALTER TABLE quota_buckets ADD COLUMN attributed_user_id ...` (and make the backfill/PK replacement safe for both legacy and partially upgraded schemas), plus a PostgreSQL migration regression test.

astaxie (commented):
The original legacy-PostgreSQL startup blocker is fixed at `c34adddb`, but two issues still block approval:

[P1] Run the new PostgreSQL regressions in CI (`.github/workflows/ci.yml:100-103`). Both `TestQuotaBucketMigrationPostgresAddsAttributionColumnBeforeBackfill` and `TestUserQuotaIsAtomicAcrossPostgresInstances` are behind the `integration` build tag, while the PostgreSQL job filters execution to exactly `^TestMultiInstancePostgresE2E$`; default `go test ./...` excludes them as well. The current green checks therefore execute neither the migration regression added by this commit nor the PR's direct cross-instance user-quota test. Expand the filter or add an explicit invocation so these tests actually run on every PR.

[P2] Scope the PostgreSQL primary-key metadata query to the active schema (`backend/internal/server/store_database.go:390`). It currently selects every primary key named on a `quota_buckets` table across all schemas. In a database containing `public.quota_buckets` plus a TokenHub schema (the new migration test creates exactly this shape), `pkColumns` combines both schemas, `len(pkColumns)` is never 4, and every startup unnecessarily drops and recreates the active table's primary key under an exclusive lock. Add `tc.table_schema = current_schema()` (or equivalent) and make the second migration assertion verify that the PK was not rebuilt.

astaxie (commented):
Re-reviewed the complete current diff at exact head `d636b4dd`. No remaining code, standards, or spec findings were found. All previously reported blockers are fixed, and exact-head CI is fully green, including the legacy PostgreSQL quota-bucket migration, primary-key idempotence, and cross-instance aggregate user-quota tests. Focused local quota/image/migration tests, `go vet ./...`, all 125 repository tests, explicit translation/UI/environment/source-line gates, and `git diff --check` pass. GitHub rejected the attempted APPROVE review because the authenticated `astaxie` account is also this PR's author; a non-author reviewer must approve if approval is required. Not merged.

astaxie (commented):
Re-reviewed the complete current diff at exact head 4739b3e39c40079d23c4037eb6e1e5e61285aa71. One new quota-settlement blocker remains.

[P1] finishCallTransaction wraps minute-token reconciliation, attributed key usage, and all user day/month reservation settlement inside the successful lookup of the live API key row at store_routing_calls.go:655-743. If an administrator deletes a key after StartCall admits a request but before FinishCall runs, the request and usage logs are still written, but the user reservation is never reconciled to actual tokens and the attributed quota buckets are not updated. This can under-enforce when actual usage exceeds the reservation or over-enforce when the reservation exceeds actual usage, breaking continuity across key deletion and rotation.

Please settle the immutable CallContext quota data even when the key row no longer exists, keep only key-dependent alerting or model data conditional, and add a regression that deletes the key between admission and completion.

All GitHub checks are green. The authenticated astaxie account is also the PR author, so GitHub cannot accept an APPROVE review from this account.

backend/internal/server/store_routing_calls.go
```diff
@@ -621,8 +705,46 @@ func (s *GormStore) finishCallTransaction(tx *gorm.DB, call CallContext, route R
...
-		if err := raiseQuotaAlerts(tx, key, &dayCounter.QuotaCounter, &monthCounter.QuotaCounter); err != nil {
-			return err
+		for _, period := range []struct {
+			scope  string
+			bucket string
+		}{
+			{scope: "day", bucket: dayBucket(admittedAt)},
+			{scope: "month", bucket: monthBucket(admittedAt)},
+		} {
+			if err := s.addAttributedQuotaUsage(tx, call.Key.ID, period.scope, period.bucket, attributedUserID, usage); err != nil {
```
astaxie: This writes the attributed per-user history with the raw provider `usage`, but the direct user quota bucket below is settled with `actualTokens` after the interrupted-stream fallback. When a stream has already emitted data and the provider never returns usage, `actualTokens` is raised to the reservation, while this attributed row records zero tokens. Because later user-scope usage/admission merges these attributed API-key rows through `aggregateUserQuotaCounter`, policy creation or owner/key rotation can undercount interrupted streams. Consider passing the same quota-settlement usage here, or otherwise applying the `actualTokens` fallback before updating attributed history.

astaxie: Fixed in 5c44ee661. Attributed day/month history and direct user day/month settlement now use the same quotaUsage derived from actualTokens after the interrupted-stream reservation fallback. Added TestInterruptedStreamUsageAppliesToPoliciesCreatedLater, which reproduces committed output with no provider usage and verifies that a later user quota policy enforces the preserved reservation. Targeted tests, the full backend suite, repository gates, and exact-head CI all pass.

backend/internal/server/store_call_admission.go
```diff
@@ -78,14 +99,59 @@ func (s *GormStore) admitCallTransaction(ctx context.Context, tx *gorm.DB, key A
...
+		// limits here: an API-key-specific limit must not become an aggregate
+		// limit for every key attributed to the same user.
+		userMinuteScopes := MinuteLimitScopes{}
+		if userPolicy.Limits.RateLimitRPM > 0 {
+			userMinuteScopes.RPM = "user"
+		}
+		if userPolicy.Limits.TokenLimitTPM > 0 {
+			userMinuteScopes.TPM = "user"
+		}
+		userMinuteCounter, err = s.consumeAPIKeyMinuteRequest(tx, userQuotaID, userPolicy.Limits, userMinuteScopes, tokenReservation, now, userPolicy.UserID)
```
astaxie, line 124: Issue #259 says a rejected user-quota call should return `429 quota_exceeded` while recording the non-sensitive blocking scope. This user-scope RPM/TPM path still delegates to `consumeAPIKeyMinuteRequest`, which returns `api_key_rpm_exceeded` / `api_key_tpm_exceeded`; the response has `details.scope = "user"`, but clients still see an API-key-specific error code for a user-scoped quota policy. If preserving the existing minute-limit codes is intentional for compatibility, the acceptance criteria/docs/tests should say so explicitly; otherwise this branch should map user minute-limit failures to `quota_exceeded`.

astaxie: Fixed in 677ef0f1a. User-scoped RPM and TPM admission failures are now normalized to HTTP 429 quota_exceeded with details.scope=user, while preserving the existing Retry-After and rate-limit headers. Aggregate user concurrency rejection now follows the same Issue #259 contract; API-key-scoped RPM, TPM, and concurrency errors retain their existing codes. Added TestUserMinuteQuotaRejectionsUseGenericQuotaError and updated the aggregate-concurrency regression. Targeted tests, the full backend suite, repository gates, two-axis review, and exact-head CI all pass.

PR #265, opened 2026-08-17
feat(db): add forward-only database evolution with migration ledger

deepjerry-ai:
## Summary

TokenHub currently evolves its database implicitly through the ORM schema flow, with no migration ledger, no checksum or dirty-state tracking, and no way to certify that an older release can safely serve a newer database. Upgrades and rollbacks therefore have no database compatibility guarantee.

This PR implements the forward-only database evolution system documented by the repository-managed `docs/database-evolution.md`: an explicit migration ledger with a narrow `database/sql`-based runner, expand migrations that run at startup, contract migrations that only run through a maintenance command, verified repair of dirty migrations, a data-backfill ledger with blocking and online modes, per-instance heartbeats, rollback compatibility verdicts, and managed-upgrade preflight plus one-shot auto-rollback. CI runs a v0.4.0 N-1 two-way CRUD contract on SQLite and PostgreSQL, plus a separate real-v0.5.0 SQLite fixture, adoption, and reboot check.

## Related Issue

N/A. The normative lifecycle and safety contract is maintained in `docs/database-evolution.md`, with synchronized Simplified Chinese and Japanese copies.

## Changes

- `backend/internal/dbschema`: migration runner with ledger (`schema_migrations`), append-only attempt audit (`migration_attempts`, now including the executing instance), adoption baseline from frozen dialect SQL, semantic verification against a reference snapshot, expand/contract phases, verified repair without force-version, data-backfill executor, and per-migration lock/statement budgets. Migration bodies and required postconditions share one statement budget on both SQLite and PostgreSQL; repair verification is also bounded.
- Build-time migration manifest: `backend/cmd/manifestgen` generates the embedded `migrations_manifest.json` (dialect baselines plus registry checksums and budgets); CI fails when the embedded copy is stale.
- `tokenhub db` maintenance CLI: `status`, `verify`, `migrate`, `repair`, and `contract` (dry-run, built-in verified SQLite backup, operator backup reference for PostgreSQL, maintenance assertion).
- Server: startup adoption under the schema section, readiness gated on dirty/blocking/unverifiable state, fail-closed instance heartbeats with TTL, managed rollback compatibility preflight, and managed upgrade preflight (target binary `db migrate` followed by `db verify` before activation) with a one-shot auto-rollback guard on failed boots. Before the first migration mutation, the runner rejects unknown versions or checksums and dirty ledger state.
- Admin console: read-only database evolution status and rollback compatibility badges (compatible / incompatible / unknown), with stable localized reason codes, parameters, and locale-aware numeric formatting.
- Immutable N-1 legacy schema fixtures verified by `backend/cmd/n1check`: v0.4.0 on SQLite and PostgreSQL, plus v0.5.0 on SQLite.
- CI: v0.4.0 N-1 two-way contracts on SQLite and PostgreSQL cover legacy creation, adoption, readiness, an API contract (auth, project, API key, provider, model and route, one gateway request, and audit writes), rollback-window reboot, and selected current-release read-back checks. A separate SQLite leg validates the real v0.5.0 fixture, adoption, and both releases booting; it does not claim a v0.5.0 PostgreSQL or CRUD leg. The stable aggregate `CI` check requires the N-1 job.
- Documentation: `docs/database-evolution.md` is the repository-managed normative source, synchronized in English, Simplified Chinese, and Japanese. Stale references to unavailable ADR 0005–0007 documents were removed.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor or maintenance
- [ ] Documentation

## Verification

- `go test ./...`, `go vet ./...`, and `gofmt` pass across the backend.
- PostgreSQL 16 integration tests pass, including the baseline/adoption coverage and the combined migration body/postcondition statement-budget regression.
- SQLite regression coverage verifies that a postcondition exceeding the shared statement budget leaves the non-transactional migration dirty; the matching PostgreSQL regression verifies the same dialect path.
- The N-1 workflow covers the v0.4.0 SQLite/PostgreSQL two-way contract and the separate v0.5.0 SQLite fixture/adoption/reboot leg described above.
- Frontend lint, typecheck, 64 domain tests, 9 component tests, and production build pass; Japanese assertions cover locale-aware database evolution reason parameters.
- All 126 repository-gate tests pass, along with explicit document/UI translation, environment contract, source-line, full translation co-change, and `git diff --check` gates.
- Skipped locally: browser smoke (`npm run test:e2e`); the `frontend-browser-smoke` CI job covers it with runner-provided Chrome.

- No OpenAI-compatible `/v1` API changes; the gateway contract is preserved.
- New `tokenhub db` maintenance commands; no new environment variables (reuses `TOKENHUB_DATABASE_URL`).
- Startup behavior changes on unadopted databases: fresh databases are created from frozen baseline SQL, and legacy databases are adopted with semantic verification; unrecognized databases refuse to start.
- Managed upgrades run `db migrate` and then `db verify` with the target binary before activation, and may auto-reactivate the previous release once after a failed first boot; the database keeps completed expand migrations (no automatic backup restore).
- `migration_attempts` gains an `executor` column; the table is created by the runner itself and is not part of the frozen baseline.
- Migration content is checksum-verified on every start; the embedded manifest gate keeps the release fingerprint in sync with source.

## Checklist


backend/internal/server/version_update.go
```diff
@@ -259,9 +263,259 @@ func (s *versionService) applyNativeRelease(ctx context.Context, release githubR
...
+// against the current database before activation (ADR 0005): `db verify` is
+// the read-only semantic preflight and `db migrate` applies the pending expand
+// migrations under the target's own runner. Completing expands before
+// activation is intentional; the database keeps them even if the upgrade is
+// later rolled back (ADR 0005: no automatic backup restore).
+func (s *versionService) runTargetDatabasePreflight(ctx context.Context, binary, version string) error {
+	if s.databaseURL == "" {
+		return errors.New("database URL is not configured; cannot preflight the target release")
+	}
+	for _, args := range [][]string{{"db", "verify"}, {"db", "migrate"}} {
```
astaxie: This preflight order looks like it will reject the first real expand migration. `tokenhub db verify` runs semantic verification against the target binary reference schema, and `CompareObjects` reports missing target tables/columns as violations. Before `db migrate` runs, a valid pre-upgrade database is expected to be missing exactly those new expand objects, so the managed upgrade would fail before it gets the chance to apply them. The read-only preflight needs to verify ledger compatibility/current-state invariants, or run after the target expand step.

backend/internal/dbcli/cli.go
```diff
@@ -0,0 +1,403 @@
...
+	return nil
+}
+
+// requireNoLiveInstances refuses contract execution while any instance
+// publishes a fresh heartbeat. Compatibility-range comparison arrives with
+// the release manifest; until then any live instance counts as potentially
+// incompatible, the conservative reading of ADR 0005. A missing heartbeat
+// table also refuses: without it the preflight cannot prove that no instance
+// is serving, and a server whose heartbeat publication failed keeps running.
+func requireNoLiveInstances(ctx context.Context, s *session) error {
```
astaxie, line 370: Counting only fresh heartbeat rows leaves a startup race for contract migrations. Server startup releases the schema migration lock when `OpenStoreWithConfig` returns, but the heartbeat is not published until `NewWithConfig` runs later. A process in that gap has already passed schema startup and can go on to serve, while `tokenhub db contract` can acquire the lock, see zero live heartbeats, and execute destructive contracts. Contract preflight needs to exclude booting instances too, for example by publishing/registering before releasing the schema lock or by holding a startup/serving marker through heartbeat publication.

backend/internal/dbschema/dbschema.go
```diff
@@ -0,0 +1,493 @@
...
+	Postcondition    func(ctx context.Context, db Execer) error
+	// SafeRetry declares that re-executing the statements after a failed
+	// non-transactional run is harmless. Repair may then drop the dirty row
+	// and re-apply the migration; without it a failed postcondition refuses
+	// repair (ADR 0005: no generic force-version escape hatch).
+	SafeRetry bool
+	// ChecksumOverride pins the checksum explicitly. It is required for Go
+	// migrations, whose source checksum is produced by the build-time manifest.
+	ChecksumOverride string
+	// LockTimeoutSeconds declares how long this migration may hold the
```
astaxie, line 144: The lock and statement budgets are recorded in the manifest, but I do not see them enforced when migrations run. `acquireLock` still uses the runner-wide `r.lockWait`, and `runMigrationBody` executes all statements/callbacks without deriving a per-migration deadline or checking the statement count against `StatementBudget`. That means an expand can still hold the startup lock or run unbounded work despite the safety fields declared here.

.github/workflows/ci.yml
```diff
@@ -100,7 +110,252 @@ jobs:
...
+          run_crud_contract "http://127.0.0.1:18083"
+          kill "$crud_pid"; wait "$crud_pid" 2>/dev/null || true
+          # The current release returns to the same database and must read
+          # every write the old release made.
+          cur_pid="$(boot_keep "$RUNNER_TEMP/current-tokenhub" 18084 /readyz "$RUNNER_TEMP/cur-e.log" "sqlite://${N1_DATABASE}")"
+          verify_current_reads "http://127.0.0.1:18084"
+          kill "$cur_pid"; wait "$cur_pid" 2>/dev/null || true
+          TOKENHUB_DATABASE_URL="sqlite://${N1_DATABASE}" "$RUNNER_TEMP/current-tokenhub" db status
+
+          # Leg 2: v0.5.0 is the other pre-ledger stable release; legacy
```
astaxie, line 351: This comment says the v0.5.0 legacy adoption is validated with its real fixture, but this leg only boots v0.5.0, adopts it, and verifies the current schema after adoption. The only fixture checker built here is `n1check`, and it is hard-wired to the v0.4.0 fixture files. If v0.5.0 created a different legacy shape that the frozen flow accidentally papered over, this CI leg would still pass. Please add a v0.5 fixture/checker invocation before adoption or adjust the claim.

backend/internal/dbschema/baseline.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+	result.Applied = append(result.Applied, migrated.Applied...)
+	return result, nil, true
+}
+
+// beginAttemptOn records an attempt start on a held SQLite connection and is
+// the single-connection-pool-safe variant of beginAttempt.
+func (r *Runner) beginAttemptOn(ctx context.Context, db Execer, version int64) (int64, error) {
+	result, err := db.ExecContext(ctx,
+		"INSERT INTO migration_attempts (version, app_release, started_at) VALUES (?, ?, ?)",
```
astaxie: This SQLite fresh-adoption attempt path does not write the `executor` column, so successful fresh SQLite starts are recorded with the default empty executor even when the runner was configured with `WithExecutor`. That undercuts the new migration_attempts executor audit for the common fresh-database path; this helper should mirror `beginAttempt` and include `executor` in the insert.

frontend/features/admin/views/database-model-pricing.tsx
```diff
@@ -173,6 +186,48 @@ export function DatabaseStatusView({ api }: { api: ApiContext; isDark: boolean }
...
+                <Database />
+              </span>
+              <h2>{tx("数据库演进")}</h2>
+            </div>
+            <div className="database-status-value-row">
+              <span className={`database-status-state ${schema.ready ? "normal" : "error"}`}>
+                {schema.ready ? tx("就绪") : tx("未就绪")}
+              </span>
+              <span className="database-status-card-value">
+                {tx("数据库状态版本")} {schema.schema_version}
```
astaxie: This panel builds localized UI text by concatenating translated fragments with values and punctuation here and below (`数据库状态版本` + value, `待执行迁移`: count, `数据回填`: count, `在线实例`: list). The repo guidance asks dynamic UI text to use complete translated templates rather than stitched fragments, so these should be catalog entries such as `数据库状态版本 {version}` and `待执行迁移：{count}` rendered with `formatTranslationTemplate`.

backend/cmd/tokenhub/main.go
```diff
@@ -34,10 +40,26 @@ func main() {
...
+	}
+	cancelGuard()
+
 	store, err := server.OpenStoreWithConfig(config.DatabaseURL, config)
 	if err != nil {
 		log.Fatal(err)
 	}
+	// The target release completed its database schema flow; settle the
+	// pending upgrade so a later boot never auto-rolls back a healthy release.
+	if err := server.RecordStartupGuardSuccess(config); err != nil {
```
astaxie: This settles the pending managed-upgrade state before `RunStartupBootstrap` has succeeded. The PR/docs describe auto-rollback after a failed first boot, but if the target release fails during bootstrap (catalog/default data seeding, etc.) the state is already removed, so the next boot no longer re-activates the previous release. Please move `RecordStartupGuardSuccess` until after all startup work that must complete before the first boot is considered healthy, or narrow the documented guarantee to schema-flow failures only.

frontend/features/admin/views/database-model-pricing.tsx
```diff
@@ -173,6 +186,48 @@ export function DatabaseStatusView({ api }: { api: ApiContext; isDark: boolean }
...
+                <span className="database-status-note">
+                  {formatTranslationTemplate(tx("兼容范围 {min} – {max}"), {
+                    min: String(schema.compatibility.min_compatible),
+                    max: String(schema.compatibility.max_compatible),
+                  })}
+                </span>
+              )}
+            </div>
+            {!schema.ready && schema.reason && (
+              <div className="database-status-note" role="alert">{schema.reason}</div>
```
astaxie: This renders the backend-generated readiness reason directly into the localized admin UI. These reasons are application diagnostics, not user/provider data, so they bypass the canonical `tx("...")` catalog path and will always appear in English for zh-CN/ja users. Consider returning a stable reason code plus parameters, or mapping known reasons through complete translated templates before display.

backend/internal/server/http.go
```diff
@@ -218,6 +222,21 @@ func (s *Server) handleHealth(w http.ResponseWriter, r *http.Request) {
 		writeJSON(w, http.StatusServiceUnavailable, map[string]any{"status": "unavailable", "service": "tokenhub-backend"})
 		return
 	}
+	// Readiness reflects the database evolution state: a dirty or unverifiable
```
astaxie, line 225: This changes `/readyz` and `/healthz` from pure database availability probes into database-evolution readiness gates, but the deployment and architecture docs still say these endpoints return 503 only when the database is unavailable. Operators using these probes need to know that dirty/unverifiable migrations and incomplete blocking backfills also make readiness fail; please update the English, zh-CN, and ja docs alongside this shared behavior change.

backend/internal/server/version_update.go
deepjerry-ai: Fixed in b467c3a: the preflight now runs `db migrate` first and `db verify` after. The migrate step's runner already rejects an incompatible ledger (unknown checksums, dirty state) before touching anything, and the semantic verify can only pass once the expands it expects are in place — verifying first would reject every valid pre-upgrade database for exactly the objects the pending migrations add.

backend/internal/dbcli/cli.go
deepjerry-ai: Fixed in b467c3a: OpenStoreWithConfig now publishes the initial instance heartbeat while the schema migration lock is still held, after adoption and the data repairs. The published instance ID is kept on the store and StartInstanceHeartbeat refreshes that row instead of creating a second one, so the lock-release-to-publication gap is closed; a failed initial publish stays non-fatal and readiness keeps gating the instance until the refresher succeeds.

backend/internal/dbschema/dbschema.go
deepjerry-ai: Fixed in b467c3a: applyMigration now derives a per-migration context deadline from LockTimeoutSeconds (migration body and postcondition included), and runMigrationBody wraps the Execer in a budgetedExecer that counts every executed statement against StatementBudget and fails the migration when either budget is exceeded. Oversized SQL statement lists are rejected up front, and transactional/NonTransactional outcomes follow the existing rollback/dirty-marker paths. Tests in internal/dbschema/budgets_test.go cover both budgets. The lock-wait for acquiring the cluster lock stays runner-wide, which we read as acquire-side policy rather than the per-migration hold budget.

.github/workflows/ci.yml
deepjerry-ai: Fixed in b467c3a: n1check gained `-fixture <name>`, and a v0.5.0 SQLite legacy fixture generated from the real v0.5.0 binary is now committed (`n1-legacy-v050-sqlite.json`; it differs materially from v0.4.0 — 10 additional tables plus new columns). The CI leg runs `n1check -fixture v050` on the v0.5.0-created database before the current release adopts it. A v0.5.0 PostgreSQL fixture is not added because the PostgreSQL N-1 job only exercises v0.4.0; it can follow when that leg grows a v0.5.0 boot.

backend/internal/dbschema/baseline.go
deepjerry-ai: Fixed in b467c3a: beginAttemptOn now inserts the executor column, mirroring beginAttempt, so fresh SQLite adoptions record the configured executor in migration_attempts.

frontend/features/admin/views/database-model-pricing.tsx
deepjerry-ai: Fixed in b467c3a: the panel now renders complete translated templates through formatTranslationTemplate (`数据库状态版本 {version}`, `待执行迁移：{count}`, `数据回填：{count}`, `在线实例：{instances}`), with matching en/ja catalog entries in db-evolution.tsx.

backend/cmd/tokenhub/main.go
deepjerry-ai: Fixed in b467c3a: RecordStartupGuardSuccess now runs after RunStartupBootstrap succeeds, so a target release that fails during bootstrap keeps the upgrade state armed and the next boot still auto-rolls back. Provider catalog initialization stays after settling deliberately — it is soft-failing and falls back to the database snapshot.

frontend/features/admin/views/database-model-pricing.tsx
deepjerry-ai: Fixed in b467c3a: DatabaseEvolutionStatus now carries a stable reason_code (baseline_missing, heartbeat_failing, dirty_migration, ledger_verification_failed, expand_pending, blocking_backfills_pending, ...) and schema-status also returns dirty_version. The panel maps the known codes to complete translated templates and falls back to the raw reason only for unknown codes.

backend/internal/server/http.go
deepjerry-ai: Fixed in b467c3a: deployment and architecture docs in English, Simplified Chinese, and Japanese now state that /readyz and /healthz also return 503 on dirty or unverifiable migration ledgers and incomplete blocking backfills, and that pending online backfills keep the instance ready.

docs/database-evolution.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Operational notes
+
+- `tokenhub db migrate` on a database without an adoption baseline points you at a normal server start: adoption happens there, inside the serialized schema section.
+- A refused contract tells you which precondition failed; nothing was executed.
+- After a rollback to a previous release, the previous release keeps working on the current database; when the newer release returns, it re-verifies the ledger and continues.
+- A managed upgrade runs the target release's own binary against the database first — a read-only preflight (`db verify`) and then the pending expand migrations (`db migrate`) — and only activates the target once both pass. If the activated release then fails its first boot, the previous release is re-activated automatically once (only when the upgrade ran no contract and the previous release's compatibility record covers the database state); a second failure stops version switching for operator recovery.
```
astaxie: [P2] Keep the managed-upgrade documentation consistent with the executable order

The current implementation intentionally runs db migrate before db verify, but this English page still promises a read-only verify followed by migrate; the zh-CN/ja copies carry the same stale order. Operators cannot rely on the documented preflight semantics as written. Please update all three documents and the PR description to describe the chosen migrate-then-verify behavior, including what compatibility checks happen before the first mutation.

frontend/features/admin/shell/version-status.tsx
```diff
@@ -577,32 +579,50 @@ export function VersionStatus({ api, user }: { api: ApiContext; user: AdminUser
...
+                              return (
+                                <div
+                                  className={selectedVersion === version.version ? "version-option selected" : "version-option"}
+                                  key={version.version}
                                 >
-                                  <ExternalLink size={14} />
-                                </a>
-                              </div>
-                            ))}
+                                  <label title={blocked ? version.compatibility_reason : undefined}>
```
astaxie: [P2] Localize rollback compatibility reasons instead of exposing backend English

Blocked options render compatibility_reason directly as a tooltip, while version_compat.go constructs English interface prose (including versions and ranges). This bypasses the canonical Simplified-Chinese tx key and EN/JA catalog rule. Return a stable reason code plus parameters and render complete translated templates here.

.github/workflows/ci.yml
```diff
@@ -100,7 +110,256 @@ jobs:
...
+              "${base}/api/admin/provider-models/import")"
+            [ "$(printf '%s' "$imported" | jq -r '.created_routes')" = "1" ]
+            gateway="$(curl -fsS -X POST -H "Authorization: Bearer ${key}" -H "$ct" \
+              -d '{"model":"n1-upstream","messages":[{"role":"user","content":"hello from n1"}]}' \
+              "${base}/v1/chat/completions")"
+            printf '%s' "$gateway" | grep -q '"content":"Echo: hello from n1"'
+            curl -fsS -H "$auth" "${base}/api/admin/audit/events" | jq -e '.data | length >= 1' >/dev/null
+            echo "PostgreSQL N-1 CRUD contract passed"
+          }
+          verify_current_reads() {
```
astaxie, line 190: [P2] Make the N-1 assertions match the documented coverage

The PostgreSQL leg checks out only v0.4.0, while v0.5.0 is exercised only by the SQLite leg. This verify_current_reads function then checks only projects and models, although the PR body and database-evolution docs claim both previous releases across both dialects and read-back of every team, API-key, provider, route, gateway, and audit write. Please expand the matrix/read-back assertions or narrow the claims to the evidence actually enforced.

astaxie (commented):
Re-reviewed the complete current diff and the nine author replies. Full backend tests, go vet, frontend lint/typecheck/tests/build, Chromium E2E, all 122 repository tests, explicit translation/environment/source-line gates, and diff checks pass locally; all current GitHub checks are green. The earlier fixes are largely present, but two prior threads remain incomplete and three additional documentation/UI/contract-evidence blockers are inline, so this remains a comment-only review.

backend/internal/dbcli/cli.go
astaxie: [P1] The success path is fixed, but the failure path still permits the original race. newStoreWithDialect logs and continues when publishInitialInstanceHeartbeat fails under the schema lock. If StartInstanceHeartbeat then fails its CREATE TABLE call, it returns a no-op without setting heartbeatState to failing; DatabaseEvolutionStatus sees heartbeatOff and can report ready while no row exists. A transient failure can therefore release the lock and later serve with zero heartbeat, allowing contract preflight to miss the booting instance. Please fail startup or set a fail-closed state whenever the under-lock publication cannot be established, and add a regression for that path.

frontend/features/admin/views/database-model-pricing.tsx
astaxie: [P2] The stable-code approach is in place, but this remains incomplete. evolutionReasonText maps six codes and falls back to schema.reason; DatabaseEvolutionStatus also emits handle_unavailable, runner_error, ledger_unreadable, and backfill_ledger_unreadable, so each of those known paths still displays backend English in zh-CN/ja. Please map every known code to complete translated templates and reserve a non-prose fallback that does not expose backend English as localized UI.

astaxie (commented):
Re-reviewed the complete current diff at `1f4da64e`. The runner-provided Chrome change is sound: the focused repository test passes, the full GitHub browser smoke check passes, and all current checks are green. This commit does not address the four existing open blockers.

[P1] Include the N-1 contract in the stable CI aggregate (`.github/workflows/ci.yml:215,575-585`). The new `backend-db-n1` job is not present in `ci.needs`, which still expects only five jobs. Branch protection can therefore receive a green `CI` check even when this compatibility contract fails. Please add `backend-db-n1` to `ci.needs`, update `EXPECTED_JOBS` to 6, and pin the aggregate behavior with a repository-gate regression.

This remains a comment-only review.

frontend/features/admin/views/database-model-pricing.tsx
deepjerry-ai: **Fixed** in 19c992c. The frontend now maps all ten DatabaseEvolutionStatus reason codes, including handle_unavailable, runner_error, ledger_unreadable, and backfill_ledger_unreadable, through complete English/Japanese translation entries. Unknown codes use a localized generic message and never render schema.reason. db-evolution-reasons.test.tsx covers every known code and the unknown-code fallback.

frontend/features/admin/shell/version-status.tsx
deepjerry-ai: **Fixed** in 19c992c. Rollback verdicts now return compatibility_reason_code plus structured parameters. version-status.tsx renders those codes through complete translated templates and uses a localized generic fallback, so compatibility_reason remains diagnostic-only and backend English is not exposed in localized tooltips. The formatter has focused unit coverage.

.github/workflows/ci.yml
deepjerry-ai: **Fixed** in 19c992c by narrowing the claims to the enforced evidence. All three database-evolution documents and the PR description now state that v0.4.0 runs the SQLite/PostgreSQL CRUD contract, current read-back checks selected durable records, and v0.5.0 is a separate SQLite fixture/adoption/reboot leg with no PostgreSQL or CRUD claim.

docs/database-evolution.md
deepjerry-ai: **Fixed** in 19c992c. The English, Simplified Chinese, and Japanese database-evolution pages now document db migrate before db verify and state that the runner rejects unknown versions, checksum drift, and dirty state before the first mutation. The PR description has been updated to the same executable order.

backend/internal/dbcli/cli.go
deepjerry-ai: **Fixed** in 19c992c. Initial heartbeat publication now returns a startup error while the schema lock is still held, so the lock is never released without a visible instance row. StartInstanceHeartbeat also marks heartbeatState as failing when table setup fails. Regression coverage: TestOpenStoreFailsWhenInitialHeartbeatCannotPublish and TestHeartbeatTableSetupFailureFailsReadinessClosed.

deepjerry-ai:
**Fixed** in 19c992c. The stable CI aggregate now includes `backend-db-n1` in `ci.needs`, `EXPECTED_JOBS` is 6, and `tools/ci-aggregate.test.mjs` pins the exact required job set and count. The branch is also merged with current `main` at 285eeb6 and is mergeable.

backend/internal/dbschema/apply.go
```diff
@@ -0,0 +1,491 @@
...
+		return Applied{}, "failed", err
+	}
+	if _, err := conn.ExecContext(ctx, "COMMIT"); err != nil {
+		return Applied{}, "failed", err
+	}
+	open = false
+	if err := runMigrationBody(ctx, conn, m); err != nil {
+		return Applied{}, "dirty", err
+	}
+	if err := m.Postcondition(ctx, conn); err != nil {
```
astaxie: [P1] Charge postcondition queries to the migration budget

`runMigrationBody` creates a `budgetedExecer`, but this required non-transactional postcondition receives the raw connection; the PostgreSQL path likewise passes raw `r.db` at line 431. A postcondition can therefore execute unlimited SQL while the schema lock is held, even though `Migration.StatementBudget` promises to bound the work one migration performs and the previous fix stated that body and postcondition were included. Please enforce one combined budget across the body and postcondition, with regression coverage for both dialect paths.

docs/database-evolution.md
```diff
@@ -0,0 +1,46 @@
...
+- Regenerate the SQLite baseline after model changes with `UPDATE_BASELINE=1 go test ./internal/server -run TestSQLiteBaselineSQLIsCurrent`; the PostgreSQL baseline regenerates the same way with `TEST_POSTGRES_URL` set and the `integration` build tag. Tests fail on a stale baseline.
+- CI runs the PostgreSQL integration suite and a v0.4.0 N-1 two-way contract on SQLite and PostgreSQL: the old binary creates a database, the current release adopts it and reports ready, the old binary boots again and completes an API contract (auth, project, API key, provider, model and route, one gateway request, and audit writes), and the current release returns and reads selected durable records (project and model on both dialects, plus provider on SQLite). A separate SQLite leg pins the real v0.5.0 schema shape, adopts it, and proves both releases can boot; it does not run the CRUD contract or a PostgreSQL v0.5.0 leg. The committed immutable fixtures under `backend/internal/dbschema/fixtures/` cover v0.4.0 on both dialects and v0.5.0 on SQLite; CI checks each covered database with `go run ./cmd/n1check` before adoption (ADR 0005).
```
astaxie: [P2] Commit the design sources cited as normative

The PR body says this system is specified by `docs/adr/0005-use-forward-only-database-evolution.md` and ADR 0006, while these lines and production code/CI repeatedly cite ADR 0005, 0006, and 0007. None of those ADRs exists in this head, `main`, or the GitHub tree. That leaves the contract, repair, budget, backfill, and rollback rules pointing to an unavailable design authority. Please commit the referenced ADRs, or replace the claims and citations with an actual committed normative source.

frontend/features/admin/views/database-model-pricing.tsx
```diff
@@ -173,6 +187,58 @@ export function DatabaseStatusView({ api }: { api: ApiContext; isDark: boolean }
...
+              </span>
+              <h2>{tx("数据库演进")}</h2>
+            </div>
+            <div className="database-status-value-row">
+              <span className={`database-status-state ${schema.ready ? "normal" : "error"}`}>
+                {schema.ready ? tx("就绪") : tx("未就绪")}
+              </span>
+              <span className="database-status-card-value">
+                {formatTranslationTemplate(tx("数据库状态版本 {version}"), {
+                  version: String(schema.schema_version),
```
astaxie: [P2] Format numeric template values with the active locale

The schema version here, compatibility bounds, and counts below are rendered with `String(...)`; `db-evolution-reasons.tsx` does the same for dirty versions and compatibility state/range parameters. `AGENTS.md` explicitly requires user-visible numbers to use `languageLocale()` and `Intl`. Please format numeric values with the active locale (while leaving release identifiers as data) and add a non-English regression assertion.

astaxie (commented):
Re-reviewed the complete current diff at `285eeb67`. The five prior blockers are substantively fixed: heartbeat publication now fails closed, managed upgrades and all three documentation sets consistently use migrate-then-verify, evolution and rollback reasons are localized through stable codes, the N-1 claims match the enforced matrix, and the stable `CI` aggregate requires `backend-db-n1`. Full backend tests/vet, frontend lint/typecheck/tests/build, all 126 repository tests, explicit translation gates, environment/source-line gates, and `git diff --check` pass; all eight GitHub checks are also green. The three new findings below still block approval of the current head.

frontend/features/admin/views/database-model-pricing.tsx
deepjerry-ai: **Fixed** in e85fa3e. Database evolution schema versions, compatibility bounds, migration/backfill counts, dirty versions, and rollback state/range parameters now use `Intl.NumberFormat(languageLocale())` through `formatLocaleNumber`; release identifiers remain unchanged data. `db-evolution-reasons.test.tsx` includes Japanese assertions for grouped numeric reason parameters, and frontend lint, typecheck, tests, and production build pass.

docs/database-evolution.md
deepjerry-ai: **Fixed** in e85fa3e without adding unmanaged ADR files. `docs/database-evolution.md` and its synchronized Simplified Chinese and Japanese copies now identify that managed page as the repository's normative lifecycle and safety contract. All stale ADR 0005–0007 references were removed from code, tests, CI-facing commands, documentation, and the PR description. The full `origin/main..HEAD` documentation co-change check passes.

backend/internal/dbschema/apply.go
deepjerry-ai: **Fixed** in e85fa3e. `Migration.Postcondition` now receives the same budgeted execution surface as the migration body, and both phases share one `budgetedExecer` counter on SQLite and PostgreSQL. Repair postcondition verification is also bounded by the migration timeout and statement budget. `TestSQLiteStatementBudgetIncludesPostconditionQueries` and `TestPostgresStatementBudgetIncludesPostconditionQueries` prove that a postcondition cannot exceed the remaining budget and that the dirty marker remains set.

astaxie (commented):
Re-reviewed the complete current diff at `e85fa3e` and reran the applicable local validation: backend tests, `go vet`, frontend lint/typecheck/tests/build, and `git diff --check` pass; current GitHub checks are green. I found two new SQLite concurrency blockers, so this remains a comment-only review.

[P1] Keep non-transactional migration execution serialized across the dirty-marker commit (`backend/internal/dbschema/apply.go:372-394`, `backend/internal/dbschema/apply.go:459-466`). The SQLite `BEGIN IMMEDIATE` transaction is committed at line 378 to persist the dirty marker and then `open` is cleared before the body and postcondition execute. `acquireLock` is explicitly a no-op for SQLite, and `Repair` uses that same no-op path, so another process can repair or re-enter while the original migration is still executing its body/postcondition. That violates the PR's per-migration cross-process lock/safety contract and can let the retry or repair interleave with schema work. Keep a writer/coordination lock through body, postcondition, and dirty cleanup (or serialize repair through an equivalent lock), and add a concurrent SQLite regression.

[P1] Close the SQLite contract precondition-to-execution race (`backend/internal/dbschema/contract.go:75-95`, `backend/internal/dbschema/apply.go:459-466`). `ApplyContract` runs `RequireCluster` while `acquireLock` is a no-op on SQLite, then releases no coordination before each destructive contract migration later obtains its own `BEGIN IMMEDIATE`. A new server can publish a fresh heartbeat after the no-live-instances check but before the contract's write lock, and the contract will still execute against a serving instance. This contradicts the documented requirement that contract execution have no live instance heartbeats. Hold a SQLite cross-process coordination lock across the preconditions and migrations, or re-check the heartbeat under an equivalent lock immediately before mutation, with a regression covering a server start racing contract execution.

astaxie (commented):
Additional P1 from the complete upgrade-path review: managed-upgrade preflight cannot handle an unbaselined legacy database (`backend/internal/server/version_update.go:282-303`, `backend/internal/dbcli/cli.go:181-193`). `runTargetDatabasePreflight` always invokes the target binary's `db migrate` before activation, but `runMigrate` maps `ErrCodeBaselineMissing` to a hard error telling the operator to start the server once for adoption. A v0.4/v0.5 database without `schema_migrations` cannot start the target server before activation, so the managed upgrade is rejected before the normal adoption path can run. This conflicts with `docs/database-evolution.md`'s documented unbaselined-database adoption path and leaves the advertised managed upgrade unable to upgrade legacy databases. Either make the target preflight perform a safe adoption/startup flow before expand migration, or explicitly narrow the managed-upgrade compatibility contract and documentation, with an integration regression for an unbaselined legacy database.

astaxie (commented):
Further complete-diff findings on `e85fa3e`:

[P1] Bound PostgreSQL startup schema-lock acquisition (`backend/internal/server/store_database.go:887-914`). `runSchemaMigrationLocked` uses `context.Background()` and an unbounded `pg_try_advisory_lock` loop with `time.Sleep(50ms)`. A long-lived `tokenhub db contract` or hung migration can therefore make every PostgreSQL server startup wait forever, even though the database-evolution runner advertises bounded lock waiting and fail-closed timeout behavior. Pass a startup context/deadline or enforce a finite lock timeout, and add a regression that proves startup returns not-ready/error when the lock remains held.

[P2] Remove the remaining references to unavailable ADR 0005/0006 sources (`.github/workflows/ci.yml:72,198,283,332,351`). The PR body says stale ADR 0005-0007 references were removed and `docs/database-evolution.md` is now the normative source, but the CI workflow still cites those nonexistent ADRs; `git ls-tree` confirms there is no `docs/adr` tree. These comments leave manifest and N-1 requirements pointing at an unavailable design authority. Replace them with the committed normative page or commit the referenced ADRs, and keep the documentation-source cleanup claim accurate.

astaxie (approved):
Re-reviewed the complete current diff at exact head `cad6e902c22c32be90b1e36062184f8da91234c3`. The final schema-coordination changes are sound: SQLite migration and contract work now share a cross-process lock file, startup lock acquisition is bounded by context, managed upgrades use the heartbeat-free `db prepare` adoption path, and the added concurrency/prepare regressions pass.

Local validation passed: full `go test ./... -count=1` (including `internal/server`), `go vet ./...`, frontend `npm ci`, lint, typecheck, domain/component tests, production build, all 126 repository tests, explicit English/zh-CN/ja documentation translation, UI translation, environment contract, source-line, and diff checks. GitHub Backend, PostgreSQL, database N-1, Frontend, browser smoke, Deployment, Repository gates, and aggregate CI are all green.

No merge was performed; a separate non-author approval is still required if branch policy requires it.

PR #269, opened 2026-08-19
fix(codex): omit unset background probe parameter

### LINKED ISSUE #228: fix(gateway): Responses API always sends background:false, breaking strict-whitelist upstreams

Mr0bean:
## Summary

Codex subscription probes started failing with `Unsupported parameter: background` after Responses request marshaling began forcing `background: false` onto requests where the client never supplied the field. The ChatGPT Codex Responses endpoint rejects the presence of this unsupported parameter.

This change preserves the original field-presence semantics: an omitted `background` field stays omitted, while explicitly supplied `background: false` or `background: true` values remain on the wire. This is the original fix by `chuhao` and differs from #230 by preserving an explicit false value.

## Related Issue

Fixes #228.

Alternative implementation to #230.

## Changes

- Omit `background` when it was absent from the incoming Responses request.
- Preserve explicit `background: false` and `background: true` values.
- Add a Provider-level regression assertion for the default Codex subscription probe.
- Add local Codex compatibility coverage and align its usage assertions with the current public OpenAI-compatible usage object.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor or maintenance
- [ ] Documentation

## Verification

- [x] `cd backend && go test ./... -count=1`
- [x] Relevant Codex, compatibility, and metrics tests passed five consecutive runs.
- [x] `node tools/check-ui-translations.mjs --base origin/main --head HEAD`
- [x] `git diff --check`
- [ ] Frontend checks were not run because this change only touches Go request marshaling and backend tests.
- [ ] SDK live smoke tests were not run because they require compatible external provider credentials.

The public Responses request contract remains compatible: explicitly supplied boolean values are preserved, while an absent optional field is no longer synthesized. The change only affects outbound Codex subscription requests that did not provide `background`.

No database, credential format, environment variable, deployment, or user-interface changes are introduced. Rollback is a normal code rollback, but it restores the Codex subscription probe failure for strict upstream endpoints.

## Checklist

- [x] Environment variable changes are synchronized across examples, Compose, `start.sh`, and deployment documentation where applicable. No environment variables were changed.
- [x] Shared user-facing behavior is documented consistently in English, Simplified Chinese, and Japanese where applicable. No user-facing documentation or copy changed.
- [x] `data/model-catalog.yaml` remains tracked and catalog changes were reviewed where applicable. The catalog was not changed.

### DIFF
--- a/backend/internal/server/types.go
+++ b/backend/internal/server/types.go
@@ -1061,7 +1061,10 @@ func (r ResponsesRequest) MarshalJSON() ([]byte, error) {
 	setRawJSONField(raw, "model", r.Model, r.Model != "")
 	setRawJSONField(raw, "input", r.Input, r.Input != nil)
 	setRawJSONField(raw, "stream", r.Stream, true)
-	setRawJSONField(raw, "background", r.Background, r.Background || raw["background"] != nil)
+	background, backgroundPresent := raw["background"]
+	if !backgroundPresent || r.Background || strings.TrimSpace(string(background)) != "null" {
+		setRawJSONField(raw, "background", r.Background, r.Background || backgroundPresent)
+	}
 	setRawJSONField(raw, "max_output_tokens", r.MaxTokens, r.MaxTokens != 0)
 	setRawJSONField(raw, "temperature", r.Temperature, r.Temperature != nil)
 	setRawJSONField(raw, "instructions", r.Instructions, r.Instructions != "")
--- a/backend/internal/server/types.go
+++ b/backend/internal/server/types.go
@@ -1061,7 +1061,7 @@ func (r ResponsesRequest) MarshalJSON() ([]byte, error) {
 	setRawJSONField(raw, "model", r.Model, r.Model != "")
 	setRawJSONField(raw, "input", r.Input, r.Input != nil)
 	setRawJSONField(raw, "stream", r.Stream, true)
-	setRawJSONField(raw, "background", r.Background, true)
+	setRawJSONField(raw, "background", r.Background, r.Background || raw["background"] != nil)
 	setRawJSONField(raw, "max_output_tokens", r.MaxTokens, r.MaxTokens != 0)
 	setRawJSONField(raw, "temperature", r.Temperature, r.Temperature != nil)
 	setRawJSONField(raw, "instructions", r.Instructions, r.Instructions != "")
--- a/backend/internal/server/codex_compat_local_test.go
+++ b/backend/internal/server/codex_compat_local_test.go
@@ -392,12 +392,11 @@ func TestLocalCodexCompatNonStreamingCompletedAndIncompleteResponses(t *testing.
 			if message["content"] != testCase.text {
 				t.Fatalf("Chat content = %#v", message["content"])
 			}
-			gotUsage, ok := chat["usage"].(Usage)
-			if !ok ||
-				gotUsage.PromptTokens != usage.PromptTokens ||
-				gotUsage.CachedInputTokens != usage.CachedInputTokens ||
-				gotUsage.CompletionTokens != usage.CompletionTokens ||
-				gotUsage.TotalTokens != usage.TotalTokens {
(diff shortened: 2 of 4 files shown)


backend/internal/server/types.go
```diff
@@ -1061,7 +1061,7 @@ func (r ResponsesRequest) MarshalJSON() ([]byte, error) {
 	setRawJSONField(raw, "model", r.Model, r.Model != "")
 	setRawJSONField(raw, "input", r.Input, r.Input != nil)
 	setRawJSONField(raw, "stream", r.Stream, true)
-	setRawJSONField(raw, "background", r.Background, true)
+	setRawJSONField(raw, "background", r.Background, r.Background || raw["background"] != nil)
```
astaxie: [P2] Preserve or explicitly reject an explicit null value. `UnmarshalJSON` decodes `background:null` into `Background == false` while retaining `raw["background"]`; this condition therefore rewrites the caller's `null` to `background:false` on the wire. Issue #228 explicitly lists null as a rejected presence, and the surrounding comment describes a transparent raw passthrough. Please either preserve the raw null (or reject it at ingress) and add a table-driven regression for omitted/false/true/null; the current provider-probe assertion only covers the omitted path and the existing `Background:true` tests use `raw == nil`, bypassing this branch.

astaxie (commented):
Reviewed the complete current diff. The production change fixes the omitted-background case and all current GitHub checks plus the full backend validation pass. I found one P2 compatibility/test gap inline, so this is a comment-only review.

astaxie (approved):
Re-reviewed the complete current diff at `ed100a28`. The explicit-null follow-up preserves omitted/false/true/null presence semantics through the actual raw passthrough path, and the focused regression covers all four cases. The provider probe assertion also confirms the original omitted-field failure is fixed. Full backend tests, `go vet`, all 122 repository tests, translation/environment/source-line gates, and `git diff --check` pass; all current GitHub checks are green. No blocking issues remain on this head.

PR #273, opened 2026-08-20
feat(provider): add configurable global egress proxy

### LINKED ISSUE #262: [Bug] v0.6.0 Codex Subscription ignores HTTP(S)_PROXY with SSRF transport, causing provider_resource_missing
## Summary

After upgrading from TokenHub v0.5.0 to v0.6.0, Codex Subscription requests no longer use the container's `HTTP_PROXY` / `HTTPS_PROXY`. The request directly dials `chatgpt.com:443`, fails, marks the only resource unhealthy after repeated failures, and eventually returns the misleading secondary error:

```text
provider_resource_missing
Codex Subscription resource is missing
```

The resource is not missing. Its database row and valid OAuth credentials remain present. Rolling back only the application image to v0.5.0 immediately restores HTTP 200 Codex SSE requests and resource health.

## Environment

- TokenHub: `v0.6.0` (regression not present in `v0.5.0`)
- Image: `ghcr.io/astaxie/tokenhub-backend:0.6.0`
- Deployment: Docker Compose, two replicas, shared PostgreSQL, HAProxy/Keepalived VIP
- Egress: HTTP forward proxy required to reach `chatgpt.com`
- Container variables: `HTTP_PROXY` and `HTTPS_PROXY` set to the forward proxy
- Provider: OpenAI Codex Subscription with valid OAuth resource
- API: Responses API over SSE

## Steps to reproduce

1. Configure a working Codex Subscription provider and account resource on v0.5.0.
2. Run TokenHub where `chatgpt.com` must be reached through an HTTP proxy.
3. Set `HTTP_PROXY=http://proxy.example:3128` and `HTTPS_PROXY=http://proxy.example:3128` in the container.
4. Confirm a Codex Responses request succeeds on v0.5.0.
5. Upgrade only the TokenHub image to v0.6.0, preserving the database, provider, resource, route and environment.
6. Send a streaming Codex request.

## Actual behavior

```text
Codex request failed: Post "https://chatgpt.com/backend-api/codex/responses":
dial tcp <public-ip>:443: i/o timeout

Codex request failed: Post "https://chatgpt.com/backend-api/codex/responses":
dial tcp [public-ipv6]:443: network is unreachable
```

After three failures, the existing resource changes to the equivalent of:

```text
status=active
healthy=false
failure_count=3
cooldown_until=<timestamp>
```

The next routed request returns HTTP 400 `provider_resource_missing`.

## Expected behavior

- Codex Subscription should support an operator-configured outbound proxy without weakening SSRF protection for arbitrary provider URLs.
- Upgrading from v0.5.0 should not silently change Codex egress from proxy to direct dial.
- The error should distinguish an unhealthy/cooling-down resource from a missing resource.

## Isolation evidence

The proxy path itself is healthy:

```bash
curl -x http://proxy.example:3128 \
  https://chatgpt.com/backend-api/codex/responses
# HTTP 405 returned immediately, which confirms connectivity for this method/path-only test.
```

Rolling back both replicas to the official v0.5.0 image, without changing PostgreSQL, OAuth credentials, resource, route, API key or proxy variables, restores new TRAE-originated SSE requests with HTTP 200. The resource automatically returns to `healthy=true` and `failure_count=0`.

## Likely regression point

In v0.6.0, `NewWithConfig` builds the Codex client using:

```go
Transport: guardProviderUpstreamRequests(
    rotatingProviderUpstreamTransport(
        allowedProviderUpstreams,
        syntheticDNSPolicy,
        nil,
    ),
    allowedProviderUpstreams,
)
```

The guarded production transport validates DNS and redirect targets, but does not set `http.Transport.Proxy` to `http.ProxyFromEnvironment`, so Codex bypasses `HTTP_PROXY` / `HTTPS_PROXY`.

Relevant code:

- https://github.com/astaxie/TokenHub/blob/v0.6.0/backend/internal/server/http.go
- https://github.com/astaxie/TokenHub/blob/v0.6.0/backend/internal/server/provider_upstream_ssrf.go

## Suggested fix

Add a dedicated setting such as `TOKENHUB_CODEX_SUBSCRIPTION_PROXY_URL`:

- Apply it only to the built-in Codex endpoint host `chatgpt.com`.
- Preserve current DNS, redirect and private-address validation.
- Keep custom Codex base URLs on the guarded direct transport unless explicitly configured.
- Reject malformed URLs and unsupported proxy schemes.
- Never log proxy credentials.

This is safer than globally restoring environment proxy behavior for all provider URLs.

## Acceptance criteria

- A real streaming Codex Responses request returns HTTP 200 through the explicit proxy.
- With no explicit Codex proxy setting, current v0.6.0 behavior remains unchanged.
- `chatgpt.com.evil.example` and arbitrary custom hosts do not match the built-in host rule.
- Redirects to loopback, private, link-local or metadata endpoints remain blocked.
- Resource health stays true when the proxy path is healthy.
- Errors distinguish missing, disabled, unhealthy and cooling-down resources.

## Security and privacy

All account identifiers, tokens, API keys, internal addresses, resource IDs and request IDs in this report are redacted. No credentials are required to reproduce the transport behavior.

deepjerry-ai:
## Summary

Restore HTTP(S) proxy support for Codex Subscription after the v0.6.0 SSRF transport regression, and provide one operator-controlled Provider egress policy for every Provider upstream channel.

## Related Issue

Fixes #262.

The issue originally proposed an explicit Codex-only proxy. The subsequent product decision intentionally supersedes that narrower suggestion: TokenHub uses one global Provider egress policy for all Provider HTTP channels, and existing installations default to standard `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY` behavior. This broader scope does not relax Provider URL or runtime DNS/IP SSRF enforcement.

## Changes

- Add `inherit_environment`, `direct`, and `configured_proxy` modes to `cfg_gateway`, with five-second multi-replica refresh and last-known-valid fallback.
- Route inference, streaming, images, model discovery, public Provider catalog refresh, quota calls, and Provider OAuth credential refresh through the shared proxy-aware transport.
- Preserve Provider, Provider Resource, import, and custom catalog save-time SSRF/public-HTTP validation in every proxy mode; TokenHub locally resolves and validates proxied target addresses and pins proxy requests or CONNECT tunnels to validated IPs while preserving the original Provider Host and TLS SNI.
- Add encrypted and masked Basic proxy credentials plus an unsaved connection test that checks proxy connect/auth/CONNECT and target TLS without sending Provider credentials or model requests.
- Treat identifiable proxy configuration, authentication, connection, timeout, and HTTPS CONNECT failures as resource-neutral platform egress failures.
- Move proxy configuration into System Settings, remove the legacy Proxy Egress page and demo seed, and retain existing hidden database rows for compatibility.
- Pass standard proxy environment variables through local and Compose startup paths and document behavior in English, Simplified Chinese, and Japanese.
- Stabilize existing Kronk cancellation and OAuth return URL tests by replacing a timing assumption with an upstream handshake and cleaning up background workers.
- Distinguish missing, disabled, unhealthy, and cooling-down Codex resources before any upstream request, while preserving provider-level credential fallback for other adapters.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Refactor or maintenance
- [x] Documentation

## Verification

- `go test ./... -count=1`
- `go vet ./...`
- `golangci-lint v2.12.2` (0 issues)
- Focused Provider proxy, settings secret, concurrency, SSRF, cancellation, and worker cleanup regression tests
- `npm run lint`
- `npm run typecheck`
- `npm test` (65 domain tests and 11 component tests)
- `npm run build`
- `npm run test:e2e` (3 Chromium smoke tests)
- `node --test tools/*.test.mjs` (126 tests)
- Documentation, UI translation, environment contract, source line, and `git diff --check` gates
- Rendered SQLite, bundled PostgreSQL, and remote PostgreSQL Compose configurations
- Local backend, vet, lint, repository gates, race checks, and repeated focused regressions passed on `b9cc0bf`
- GitHub Actions passed on `b9cc0bf`, including Backend, PostgreSQL, N-1 database, Frontend, browser smoke, Deployment, Repository gates, aggregate CI, and Migration Unit Tests

Existing installations intentionally default to `inherit_environment`, restoring standard `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY` behavior without requiring a new setting across all Provider HTTP channels. Configured proxy credentials are encrypted at rest, masked in responses, and omitted from logs and audits. Credential encryption fails closed, so a password is never persisted when protection fails. Provider and Provider Resource URLs retain save-time scheme and literal-address validation in every proxy mode. At runtime, TokenHub locally resolves Provider hostnames, rejects private, loopback, link-local, and metadata addresses unless an existing explicit operator exception applies, and pins proxy requests or CONNECT tunnels to the validated IP; direct and `NO_PROXY` paths retain the existing guarded dialer. Proxy configuration, authentication, connection, timeout, and HTTPS CONNECT failures are resource-neutral and stop route failover. Plaintext HTTP proxy responses retain normal upstream error handling because TokenHub cannot reliably distinguish proxy-generated responses from Provider responses. Replicas converge on shared settings within five seconds. Existing `proxies` rows are retained but hidden and have no runtime effect. Rolling back leaves the new `cfg_gateway` fields ignored by older binaries.

## Checklist


astaxie (commented):
Reviewed the complete diff at exact head 27c41ed5b96880ad523a5d6114780b68fda02a3b. Two blocking issue-contract gaps remain.

1. [P1] The proxy trust model contradicts issue #262. Missing settings default to inherit_environment for every Provider channel, and validateProviderUpstreamBaseURLForProxyPolicy returns success whenever a proxy matches. The new tests explicitly accept http://169.254.169.254/v1 through a configured proxy. Issue #262 requires proxy support without weakening SSRF protection for arbitrary Provider URLs, unchanged v0.6 behavior without an explicit Codex proxy setting, and redirects to private, link-local, or metadata targets to remain blocked. Please narrow the proxy scope to the intended Codex/built-in endpoint or retain target validation for proxied custom URLs and redirects.

2. [P2] Issue #262 also requires distinct missing, disabled, unhealthy, and cooling-down errors. provider_account_codex.go still returns provider_resource_missing when route selection yields no usable resource, and this diff adds no state-specific handling or regression coverage.

All GitHub checks are green, but these security and compatibility requirements prevent approval. Please address them and request another review.

astaxie (commented):
Additional local evidence confirms the P1 proxy-scope blocker on exact head 27c41ed5. This host has only the standard lowercase http_proxy and https_proxy variables set. With the new default inherit_environment behavior, go test ./internal/server -count=1 fails the existing SSRF contract tests:

- TestAdminCreateProviderRejectsSSRFBaseURL accepts http://169.254.169.254/latest/meta-data with HTTP 201 instead of 400.
- TestAdminCreateProviderRejectsPublicHTTP accepts http://api.example.com/v1 with HTTP 201 instead of 400.
- TestAdminPatchProviderRejectsSSRFBaseURL accepts the metadata URL with HTTP 200 instead of 400.

GitHub passes because its runner does not expose the same proxy environment. This makes the security contract depend on ambient host configuration and demonstrates that the default proxy selection bypasses existing save-time guards. Please preserve these rejection tests under both uppercase and lowercase proxy environments as part of the fix.

deepjerry-ai:
Addressed both review blockers in 16c3bb2. Provider, Provider Resource, import, and custom catalog URLs now retain save-time SSRF/public-HTTP validation under uppercase/lowercase environment proxies and configured proxy mode. Codex routing now returns distinct missing, disabled, unhealthy, and cooling-down resource errors before any upstream request; the capacity transaction rechecks status, and non-Codex provider-level credential fallback remains compatible. Focused regressions passed 10 times, the full backend suite passed, and vet/lint/repository gates are green locally. Please re-review the updated head.

backend/internal/server/provider_upstream_proxy.go
```diff
@@ -0,0 +1,152 @@
...
+		proxied:     proxied,
+		selectProxy: selectProxy,
+	}
+}
+
+func (transport *providerEnvironmentProxyTransport) RoundTrip(request *http.Request) (*http.Response, error) {
+	if request == nil || request.URL == nil {
+		return nil, NewHTTPError(http.StatusBadRequest, "provider_base_url_invalid", "Base URL is invalid")
+	}
+	if err := validateProviderUpstreamURLSyntax(request.URL); err != nil {
```
astaxie: [P1] Proxied requests still bypass runtime target-address SSRF validation. Save-time validation rejects disallowed literal IPs and public HTTP, but an accepted custom URL such as `https://attacker.example/v1` relies on the guarded `DialContext` to resolve and validate the hostname on every connection. Once `selectProxy` returns a proxy, this path performs only syntax validation and the forward proxy resolves the target, so DNS rebinding can make `attacker.example:443` connect to a private, link-local, or metadata service; same-origin redirect checks do not prevent that. This still violates issue #262’s requirements to avoid weakening SSRF protection for arbitrary Provider URLs and to scope proxying to the exact built-in Codex host (`chatgpt.com`, not lookalikes/custom hosts). Please keep arbitrary custom hosts on the guarded direct path (or provide equivalent target-address enforcement at the proxy boundary), and add a regression where a proxied hostname resolves/rebinds to a denied address.

deepjerry-ai: Fixed in b9cc0bf. Proxied Provider requests now resolve targets locally on every hop, apply the same runtime IP policy as direct egress, and pin CONNECT to the validated IP while preserving the original Provider Host and TLS SNI. The implementation also retries validated multi-IP targets before sending Provider bytes, distinguishes proxy establishment failures from Provider TLS failures, enforces cancellable CONNECT deadlines, and bounds/rotates proxy transport pools without retaining raw credentials. Regression coverage includes public-to-metadata DNS rebinding, IP-pinned CONNECT with Host/SNI preservation, multi-IP fallback, legacy public-HTTP rejection before proxying, silent-proxy timeout/closure, and pool credential lifecycle.

astaxie (commented):
Re-reviewed exact head `b9cc0bf1a82ce76fbbb7c6dda34546965620afb3`. The previous proxy-boundary SSRF blocker is addressed by the new local resolution/IP-pinned CONNECT path, but one issue-contract blocker remains.

Issue #262 asked for an explicit Codex-only proxy path: "With no explicit Codex proxy setting, current v0.6.0 behavior remains unchanged" and "Keep custom Codex base URLs on the guarded direct transport unless explicitly configured." This PR still makes `inherit_environment` the default and falls through to `policy.environmentProxy(request)` in `providerProxyPolicy.proxyForRequest`, so any existing `HTTP_PROXY`/`HTTPS_PROXY` changes Provider egress by default. It also wraps the shared Provider upstream transport in `providerTransportWithEnvironmentProxy`, so the behavior applies to every Provider upstream channel, not only built-in Codex `chatgpt.com`; allowed custom Codex hosts use the same globally proxied client as well.

Please either narrow proxy selection to the issue contract (built-in Codex endpoint / explicit Codex proxy only, with custom Codex hosts staying guarded direct unless explicitly configured) or update/confirm the product decision in the issue/PR so this broader global Provider egress behavior is an intentional contract change rather than an accidental expansion.

deepjerry-ai:
@astaxie Confirmed: the global all-Provider scope is an intentional product decision made after Issue #262 was filed, not an accidental expansion. The issue decision record is here: https://github.com/astaxie/TokenHub/issues/262#issuecomment-5372189010. I also updated the PR body to state that this supersedes the original Codex-only suggestion, corrected the runtime security description to local DNS/IP validation plus IP-pinned proxy/CONNECT, and updated verification to `b9cc0bf`. The implementation remains global while preserving equivalent SSRF enforcement for arbitrary Provider URLs. Please re-review the updated contract and head.

Issue #274, opened 2026-08-20
feat(deployment): add a secure first-run initialization mode


legendtkl:
## Affected area

Deployment and configuration

## Problem or use case

In a production Kubernetes or custom-orchestrator deployment, TokenHub exits during startup when the bootstrap settings are missing, still use placeholders, or do not meet the minimum length requirements. The pod then enters a restart loop before an operator can access the admin console and complete the initial setup.

For example, startup currently fails with:

```text
unsafe prod configuration: TOKENHUB_ADMIN_TOKEN must be at least 32 bytes after trimming whitespace; TOKENHUB_SECRET_KEY must not use a default placeholder value; TOKENHUB_BOOTSTRAP_ADMIN_PASSWORD must be at least 12 bytes after trimming whitespace
```

Requiring every bootstrap value before the process can become healthy makes first-time deployment cumbersome, especially when the desired operational model is to claim and configure the instance after it starts. However, weakening the validation or falling back to default credentials would create an account-takeover or data-encryption risk.

## Proposed solution

Provide a secure first-run initialization mode for instances that have not yet been configured.

- Separate settings that must exist before any persistent or encrypted state is used from settings that can safely be completed during first-run setup.
- Keep `TOKENHUB_SECRET_KEY` strong and stable. It should either remain an explicit startup prerequisite or be generated and persisted safely by supported deployment tooling; TokenHub must never silently use an ephemeral or placeholder encryption key.
- Allow `TOKENHUB_ADMIN_TOKEN` to be omitted when session-based administration is used and no operational workflow depends on the static token.
- Allow the initial admin credential to be established through a one-time, authenticated claim/setup flow instead of requiring `TOKENHUB_BOOTSTRAP_ADMIN_PASSWORD` before startup.
- While initialization is incomplete, expose only the minimum health and setup surfaces. Keep `/v1`, normal admin APIs, background workers, and access to existing data unavailable.
- Make the initialization state and the required operator action visible through readiness status and redacted diagnostics without logging credentials.
- Guarantee single-winner initialization across restarts and multiple replicas, and permanently disable the claim flow after an administrator has been created.
- Update the supported deployment examples and the English, Simplified Chinese, and Japanese documentation for the new flow.
- Add tests for missing and placeholder settings, restarts, concurrent claims, existing databases, and secret persistence.

## Alternatives considered

[...]

PR #281, opened 2026-08-21
fix(deploy): make production bootstrap restart-safe

legendtkl:
## Summary

Prevent production deployments from entering a restart loop when optional bootstrap credentials are missing or still use known placeholders, while preserving the stable encryption-key boundary.

New file-backed SQLite deployments can now generate a durable encryption key and an encrypted, retrievable initial administrator password. Unsafe configurations that cannot be repaired automatically stay live but not ready instead of exiting immediately.

## Related Issue

Refs #274.

This PR addresses restart-safe bootstrap and secret handling only. The setup-only authenticated claim flow remains tracked by #274 and is not implemented here.

## Changes

- Treat the static Admin API token as optional and generate an initial administrator password when no explicit password is configured.
- Persist generated initial passwords as encrypted bootstrap credentials and delete the retrievable copy after the first successful login, password update, or reset.
- Add the `tokenhub initial-admin-password` command for local operator retrieval.
- Generate a mode-0600 secret-key file beside a brand-new file-backed SQLite database and reuse it across restarts; never generate a replacement key for an existing database.
- Keep PostgreSQL and other shared deployments dependent on an explicit, stable `TOKENHUB_SECRET_KEY`.
- Serve `/livez` with 200 while returning 503 from readiness, health, and application routes when startup configuration is unsafe.
- Update Compose/install behavior, the frozen SQLite and PostgreSQL bridge baselines, the migration manifest, and English, Simplified Chinese, and Japanese documentation.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor or maintenance
- [x] Documentation

## Verification

- `go vet ./...` — passed.
- Focused bootstrap, startup-blocked, production-config, SQLite-baseline, CLI, and manifest tests — passed.
- PostgreSQL baseline integration test with PostgreSQL 16 — passed.
- `bash deploy/install_test.sh` — passed.
- `node --test tools/*.test.mjs` — passed (126 tests).
- Documentation co-change, UI translation, environment contract, source line, Compose rendering, and `git diff --check` gates — passed.
- `go test ./...` — all packages except `internal/server` passed. The server package is blocked by the pre-existing flaky `TestTeamLeaderScopedResourceFilteringUsesConstantQueries/project_members` query-count assertion. The same `small=5 large=9` failure reproduced four times in 100 runs on the unmodified `origin/main`; the test passes when run once in isolation.

No public API compatibility changes.

For new file-backed SQLite databases, TokenHub writes `<database-path>.secret-key` with mode 0600 and requires that file to remain on the same persistent volume as the database. Existing databases never receive an automatically generated replacement key. PostgreSQL and multi-replica deployments must continue to provide the same explicit `TOKENHUB_SECRET_KEY` to every replica.

Generated initial passwords are encrypted with the configured root key and are retrievable only through the local CLI until the first successful login or password change. Static Admin API access remains disabled when `TOKENHUB_ADMIN_TOKEN` is omitted.

A configuration-blocked process is live but not ready and serves no application traffic. Operators must correct the configuration and restart the process.

## Checklist


backend/cmd/tokenhub/main.go
```diff
@@ -117,6 +134,68 @@ func main() {
...
 }
 
+func printInitialAdminPassword(output io.Writer) error {
+	config, err := server.ConfigFromEnv().PrepareForStartup()
+	if err != nil {
+		return err
+	}
+	if err := config.ValidateForStartup(); err != nil {
+		return err
+	}
```
astaxie, line 144: This command should open the store in maintenance mode. `OpenStoreWithConfig` publishes an instance heartbeat during schema initialization, but `initial-admin-password` is a short-lived read-only CLI command and never starts/stops the server heartbeat loop, so it can leave a stale live-instance row until the TTL expires. That can make `tokenhub db contract`/other maintenance preflights think a serving instance exists. Please use `OpenStoreForMaintenance` here, like the other DB maintenance CLI paths.

backend/internal/server/startup_config.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+	return false
+}
+
+func sqliteSecretKeyPaths(databaseURL string) (keyPath string, databasePath string, ok bool, err error) {
+	driver, dsn, err := parseDatabaseURL(databaseURL)
+	if err != nil {
+		return "", "", false, err
+	}
+	if driver != "sqlite" || dsn == "" || dsn == ":memory:" || strings.HasPrefix(dsn, "file:") {
```
astaxie: This skips valid file-backed SQLite DSNs. `parseDatabaseURL`/`sqliteDSN` support `file:` DSNs, and SQLite commonly uses forms like `file:/app/data/tokenhub.db?cache=shared` or `file:tokenhub.db?mode=rwc`; those are durable file-backed databases, but this branch returns `ok=false`, leaving production startup blocked when the secret key is a placeholder. The PR promise is to generate the key for brand-new file-backed SQLite deployments, so this should only exclude in-memory `file:` modes rather than every `file:` DSN.

backend/cmd/tokenhub/main.go
legendtkl: Updated `initial-admin-password` to open the database with `OpenStoreForMaintenance` and to close the store before returning. The CLI regression test now bootstraps through maintenance mode and asserts that password retrieval does not create the `instance_heartbeats` table.

backend/internal/server/startup_config.go
legendtkl: Updated SQLite key-path classification to accept durable `file:` URI DSNs and exclude only actual in-memory forms (`:memory:`, `file::memory:`, and `mode=memory`). Added coverage for absolute and relative file URIs plus the in-memory variants.

deploy/install.sh
```diff
@@ -263,11 +263,17 @@ elif [[ "$environment" != "dev" && "$environment" != "development" && "$environm
     local blocked
     for blocked in "$@"; do
       if [ "$value" = "$blocked" ]; then
-        validation_errors+=("$name must not use a default placeholder value")
-        return
+        # Known defaults are interpreted by the backend as an unset bootstrap
+        # value. A new SQLite deployment generates and persists the root key and
+        # initial admin password; custom non-empty weak values remain errors.
+        return 0
```
astaxie: `validate_secret` now accepts placeholders and empty values for every secret, including `TOKENHUB_SECRET_KEY`. That only matches the backend for a brand-new file-backed SQLite database; `PrepareForStartup` still leaves PostgreSQL (and existing SQLite) with the placeholder/empty key and `ValidateForStartup` then blocks startup. As a result `deploy/install.sh --check-only` can report a production PostgreSQL or existing-SQLite configuration as valid even though the backend will enter configuration-required mode. Please keep the relaxed handling scoped to the optional admin token/bootstrap password, or make this validation conditional on the SQLite-new-database case before accepting an unset root key.

backend/internal/server/store_bootstrap.go
```diff
@@ -0,0 +1,91 @@
...
+		return nil
+	})
+	if err != nil && AsHTTPError(err).Code == "admin_user_conflict" {
+		return AdminUser{}, false, nil
+	}
+	return created, createdNow, err
+}
+
+// InitialAdminPassword decrypts the generated first-run password for the local
+// administrative CLI. It disappears after the first successful login or any
```
astaxie, line 69: This is not true for every successful login path yet. Password login calls `AuthenticateAdminUser` and deletes the bootstrap credential, but admin OAuth exchange creates the session via `CreateAdminSession`, which does not call `deleteInitialAdminPassword`. If the bootstrap `usr_admin` signs in through an identity provider before a password login/change, `tokenhub initial-admin-password` remains retrievable after that successful login. Please clear the credential from `CreateAdminSession` as well, or narrow the documented guarantee.

deploy/install.sh
legendtkl: Updated the installer preflight so only the optional admin token and bootstrap password accept unset or known-placeholder values. TOKENHUB_SECRET_KEY remains required: an empty, weak, or placeholder root key now fails --check-only instead of being reported as valid. Added regression coverage for the rejected root-key placeholder and for optional placeholders paired with a strong root key.

backend/internal/server/store_bootstrap.go
legendtkl: Updated CreateAdminSession to delete the generated initial-admin-password credential in the same transaction after the session is created. This covers the OAuth exchange path as well as password login. Added a regression test that bootstraps usr_admin, creates a session directly, and verifies the initial password is no longer retrievable.

deploy/install.sh
```diff
@@ -257,28 +257,36 @@ elif [[ "$environment" != "dev" && "$environment" != "development" && "$environm
...
       validation_errors+=("$name must be at least $minimum_length bytes after trimming whitespace")
       return
     fi
   }
 
-  validate_secret "TOKENHUB_ADMIN_TOKEN" "$admin_token" 32 \
+  validate_secret "TOKENHUB_ADMIN_TOKEN" "$admin_token" 32 true \
     "dev_admin_token" "change-me-tokenhub-admin-token"
-  validate_secret "TOKENHUB_SECRET_KEY" "$secret_key" 32 \
+  validate_secret "TOKENHUB_SECRET_KEY" "$secret_key" 32 false \
```
astaxie: This still diverges from the backend and the updated deployment docs, just in the other direction. The default Compose deployment is a brand-new file-backed SQLite database (`TOKENHUB_DATABASE_URL` defaults to `sqlite:///app/data/tokenhub.db`), and `PrepareForStartup` can safely replace the placeholder root key by writing `<db>.secret-key`. But the installer now rejects the placeholder unconditionally; `./deploy/install.sh --env-file deploy/.env.example --check-only` fails before that supported first-run path can start. Please make the `TOKENHUB_SECRET_KEY` check conditional on the resolved database type/state, or update the docs/defaults so the installer and runtime agree.

backend/internal/server/seed.go
```diff
@@ -172,15 +173,17 @@ func seedBootstrapAdmin(store Store, password string) error {
 		TeamID:   "team_platform",
 		Status:   StatusActive,
 	}
-	for _, existing := range store.ListAdminUsers() {
-		if existing.ID == admin.ID || existing.Username == admin.Username || existing.Email == admin.Email {
-			return nil
+	retainInitialPassword := false
+	if strings.TrimSpace(password) == "" {
```
astaxie, line 177: Because this PR closes #274, this fallback still misses the issue’s core initialization model. When the bootstrap password is omitted, TokenHub now generates a password, creates `usr_admin`, and then `main` continues into the normal server path with the scheduler, `/v1`, and regular admin APIs enabled. The issue asked for a one-time authenticated claim/setup flow and for normal application surfaces to stay unavailable while initialization is incomplete. Please either keep the instance in a setup-only state until the operator claim completes, or narrow the PR/issue linkage so this does not claim to implement that contract.

deploy/install.sh
legendtkl: Updated the installer to read the resolved TOKENHUB_DATABASE_URL and inspect the tokenhub-data volume before deciding whether an unset or placeholder root key is safe. A missing volume, an empty SQLite database, or an existing database with its paired .secret-key sidecar is accepted; PostgreSQL and a non-empty SQLite database without the sidecar are rejected. Added regression coverage for each state, and ./deploy/install.sh --env-file deploy/.env.example --check-only now passes on a fresh default deployment.

backend/internal/server/seed.go
legendtkl: Scoped the PR as suggested. The Related Issue section now uses Refs #274 instead of a closing reference and explicitly states that this PR covers restart-safe bootstrap and secret handling only. The setup-only authenticated claim flow remains tracked by the still-open #274 and is not claimed as implemented here.

deploy/install.sh
```diff
@@ -246,43 +250,149 @@ byte_length() {
...
+    return 0
+  fi
+  if [ "$volume_names" != "tokenhub-data" ]; then
+    return 1
+  fi
+
+  if volume_mountpoint="$("$DOCKER_BIN" volume inspect --format '{{.Mountpoint}}' tokenhub-data 2>/dev/null)" &&
+    [ -d "$volume_mountpoint" ]; then
+    if [ ! -s "$volume_mountpoint/$relative_path" ] ||
+      [ -s "$volume_mountpoint/$relative_path.secret-key" ]; then
```
astaxie: This check is weaker than the backend’s sidecar validation. `readSQLiteSecretKey` rejects files that are group/other-readable and files whose trimmed contents are empty, shorter than 32 bytes, or one of the known placeholders, but the installer treats any non-empty `<db>.secret-key` as enough to allow an unset `TOKENHUB_SECRET_KEY`. That means `--check-only` can report a config as valid while the backend immediately enters configuration-required mode on the same sidecar. Please mirror the backend checks here, or do not treat an existing sidecar as sufficient unless it passes the same permission/content validation.

legendtkl: Updated the installer to validate an existing SQLite root-key sidecar before accepting an unset or placeholder TOKENHUB_SECRET_KEY. Both the host-volume path and the container fallback now reject group/other-readable files, Unicode-trimmed values shorter than 32 bytes, and the two known placeholders; the container path uses byte length and the same Unicode White_Space class. Added regression coverage for empty, short, placeholder, group-readable, and valid mode-0600 sidecars, including the container fallback.

astaxie (approved):
Reviewed the current head. The deployment changes are coherent, CI is green, and the applicable backend, deployment, Compose, repository, and install-script checks pass locally. Approved.

PR #284, opened 2026-08-21
feat(docs): add openapi gateway reference

### LINKED ISSUE #260: feat(docs): add interactive OpenAPI documentation portal
## Affected area

Documentation

## Problem or use case

TokenHub has an authenticated `/gateway` console view with role-oriented guidance, code snippets, and quick-access workflows, but it does not provide a complete machine-readable or interactive API reference.

There is currently no stable URL such as `/docs`, `/openapi.json`, or `/swagger` where an application developer can:

- discover every public gateway endpoint and operation;
- inspect request and response schemas, headers, status codes, and error objects;
- understand streaming, background-job, multipart, and binary-response behavior;
- enter a TokenHub project API key and invoke an endpoint from the browser;
- import an OpenAPI document into an SDK generator, API client, test tool, or enterprise API catalog.

The existing documentation is maintained separately from backend route registration and can drift from runtime behavior. For example, newer Responses streaming and background-job capabilities are not represented consistently. Image jobs, image assets, native Gemini endpoints, and several lifecycle operations also lack one complete online reference.

Enterprise users need a self-hosted documentation portal that works in private or offline deployments and accurately reflects the gateway contract without exposing credentials or administrative capabilities unintentionally.

## Proposed solution

Design and implement a versioned OpenAPI 3.1 documentation system for TokenHub's public gateway API, with a self-hosted interactive documentation page.

### Public documentation endpoints

- Serve an interactive Scalar or Swagger UI page at `GET /docs`.
- Serve the machine-readable specification at `GET /openapi.json`; optionally also provide YAML download.
- Derive the displayed server URL from the configured public base URL or the current request origin without hard-coding localhost.
- Keep documentation assets self-hosted so the page works without access to a public CDN.
- Keep the existing `/gateway` console guidance, but link it to the complete reference instead of treating both pages as the same artifact.

### Initial API coverage

The first complete public specification should cover every externally supported gateway operation, including:

- model discovery: `/v1/models` and `/v1/models/{model}`;
- Chat Completions, including buffered and SSE streaming requests;
- Responses, including buffered and SSE streaming requests;
- persistent background Responses retrieval and cancellation;
- Responses compaction where enabled;
- Anthropic Messages and token counting;
- embeddings;
- image generation, image edits, asynchronous image jobs, and signed image-asset retrieval;
- native Gemini `v1beta` model, generation, streaming, and token-counting routes;
- common authentication, request IDs, rate-limit headers, quota errors, provider errors, and model-access errors.

Health, metrics, analytics, and administrative APIs should be explicitly classified rather than accidentally mixed into the public model API. Administrative endpoints should use a separate authenticated specification and documentation page if they are documented later.

### Interactive calls and security

- Support `Authorization: Bearer <TOKENHUB_API_KEY>` in the documentation UI.
- Keep a key entered into the page in memory only by default; do not place it in the URL, local storage, analytics, ordinary logs, examples, or generated exports.
- Never prefill a real project key, administrator session token, provider credential, or subscription token.
- Ensure “Try it” requests are sent only to the documented TokenHub origin or an explicitly configured allowlisted public base URL.
- Document browser and CORS constraints clearly. Do not weaken production CORS or credential policies merely to make the documentation UI work.
- Hide or disable interactive execution for operations that cannot be exercised safely from a browser, while still documenting their contract.
- Render secret and credential fields as write-only and redact sensitive examples.

### Source of truth and drift prevention

- Store the public OpenAPI source in a clearly owned repository location such as `docs/openapi/gateway.openapi.yaml` or generate it deterministically from typed route contracts.
- Add CI validation for OpenAPI 3.1 syntax, duplicate operation IDs, unresolved references, schema examples, and security definitions.
- Add a route-coverage contract test that detects registered public gateway routes missing from the specification and documented routes missing from the server.
- Require endpoint changes to update the OpenAPI contract in the same pull request.
- Include automated checks for representative buffered JSON, SSE, multipart upload, background job, and binary download examples.
- Publish the TokenHub release version and an explicit API contract version in the document.

### Documentation design

- Organize operations by protocol and workflow rather than one flat endpoint list.
- Provide concise quick-start examples for cURL, Python, TypeScript, Go, and Java where practical.
- Explain which features are TokenHub-owned and which depend on the selected model, provider, route, or account entitlement.
- Describe asynchronous state transitions and polling/cancellation flows visually or with linked operation references.
- Use reusable schemas for errors, usage, rate-limit headers, model objects, jobs, and content blocks.
- Keep English as the schema and developer-facing source language while synchronizing user-facing guidance in English, Simplified Chinese, and Japanese according to repository policy.
- Meet basic accessibility requirements for keyboard navigation, contrast, focus states, and readable code samples.

### Suggested delivery phases

1. Establish the OpenAPI source, validation tooling, security model, and `/openapi.json` endpoint.
2. Document the stable OpenAI-compatible and Anthropic public endpoints, then launch the self-hosted `/docs` UI.
3. Add Responses background lifecycle, images, binary assets, and native Gemini coverage.
4. Add route-drift tests, localized guidance, deployment documentation, and release/version integration.
5. Evaluate a separate authenticated Admin API specification without exposing it through the public gateway document.

### Continue maintaining the current console documentation manually

The console guidance is useful for onboarding, but it is not machine-readable, cannot be imported into API tooling, does not provide complete endpoint coverage, and has no automated relationship to registered routes.

### Link to upstream OpenAI, Anthropic, and Gemini documentation

Upstream references explain their own protocols but do not describe TokenHub authentication, routing, quotas, errors, background jobs, image storage, supported subsets, or TokenHub-specific extensions.

### Publish only a static Markdown endpoint list

Markdown is easy to maintain initially, but it does not provide schema validation, SDK/tool import, reusable contracts, or safe interactive requests.

### Generate a specification only from Go annotations

Annotation generation can reduce duplication but often omits protocol nuance and complex examples. It should be adopted only if it produces a readable, reviewable contract and can express SSE, multipart, binary, and asynchronous workflows without hiding important behavior in generator-specific comments.

## Additional context

Suggested acceptance criteria:

- [ ] `GET /docs` serves a self-hosted interactive API reference in a standard deployment.
- [ ] `GET /openapi.json` returns a valid OpenAPI 3.1 document with stable operation IDs.
- [ ] The public document covers every registered public gateway route and excludes administrative routes by default.
- [ ] Chat Completions and Responses document both buffered and SSE streaming behavior accurately.
- [ ] Background Responses retrieval/cancellation, image jobs, multipart image edits, and binary image assets are represented accurately.
- [ ] Native Gemini routes and their protocol-specific authentication and payloads are documented.
- [ ] A developer can enter a project API key and safely execute an eligible request from the documentation UI.
- [ ] Entered keys are not persisted, included in URLs, written to logs, or embedded in generated examples.
- [ ] The page uses the configured public base URL/current origin and works in a private deployment without CDN access.
- [ ] CI validates the specification and fails when public route registration and documentation coverage diverge.
- [ ] Common errors, request IDs, rate-limit headers, quotas, and provider-dependent capability notes use reusable schemas and examples.
- [ ] Deployment and user documentation is updated in English, Simplified Chinese, and Japanese.
- [ ] Backend, frontend, repository-gate, accessibility, and browser tests cover the new endpoints and documentation page.

## Submission checklist

- [x] I searched open and closed issues for related requests.
- [x] The issue title and all explanatory text are written in English.
- [x] I removed or redacted all credentials and other secrets.

astaxie:
## Summary

Adds a self-hosted interactive OpenAPI documentation portal for TokenHub's public gateway API, so application developers can inspect and exercise the gateway contract from private deployments and import the contract into API tooling.

Closes https://github.com/astaxie/TokenHub/issues/260

## Related Issue

https://github.com/astaxie/TokenHub/issues/260

## Changes

- Add the public gateway OpenAPI 3.1 contract, Swagger UI assets, and `/docs`, `/openapi.json`, and `/openapi.yaml` server routes.
- Cover OpenAI-compatible, Anthropic, image, background response, asset, and native Gemini gateway operations while excluding admin and operational routes.
- Add backend OpenAPI validation, route drift, example, security, and representative workflow tests.
- Add browser coverage that authorizes Swagger UI with a generated project key, executes `GET /v1/models` against the documented backend origin, and confirms disabled browser operations cannot send requests.
- Update English, Simplified Chinese, and Japanese user/deployment documentation and nginx forwarding coverage.
- Address review gaps for emitted rate-limit headers, protocol-specific SSE examples, Anthropic Messages/count-tokens error shapes, Anthropic protocol headers and message schemas, inspectable Chat/Responses request and response schemas, image edit 501/error examples, localized reference copy, and browser/CORS guidance.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor or maintenance
- [x] Documentation

## Verification

- Passed: `go test ./internal/server -run 'TestOpenAPI|TestGatewayOpenAPI'`
- Passed: `go test ./...`
- Passed: `go vet ./...`
- Passed: `npm run lint`
- Passed: `npm run typecheck`
- Passed: `npm test`
- Passed: `npm run build`
- Passed: `npx playwright install chromium`
- Passed: `npm run test:e2e` (4/4 browser tests)
- Passed: `node --test tools/*.test.mjs` (127/127 tests)
- Passed: `node tools/check-doc-translations.mjs`
- Passed: `node tools/check-ui-translations.mjs`
- Passed: `node tools/check-env-contract.mjs`
- Passed: `node tools/check-source-lines.mjs`
- Passed: `git diff --check`
- Passed: `git diff --cached --check`
- Passed: `docker compose --env-file deploy/.env.example -f deploy/docker-compose.yml config`

Known warnings / skipped checks:

- `node --test tools/*.test.mjs` prints expected fixture diagnostics for documentation translation failure scenarios, but the suite itself passed 127/127.
- Node emitted existing `MODULE_TYPELESS_PACKAGE_JSON` warnings from repository test tooling; no check failed.
- Playwright web server output emitted existing `NO_COLOR` / `FORCE_COLOR` warnings; the browser suite passed.
- `node tools/check-doc-translations.mjs` ran locally with no diff endpoints, so it completed the existence check and skipped the co-change comparison as documented.
- SDK smoke tests were not run because they require a compatible backend and SDK environment variables outside this OpenAPI/docs change.

The new public documentation endpoints expose only the public gateway contract and intentionally exclude admin, analytics, health, and metrics APIs. The docs UI keeps entered API keys in memory only, disables credential persistence and generated snippets, self-hosts its assets, and restricts interactive requests to the documented TokenHub origin or configured public base URL. No environment variables, database migrations, or model catalog changes are included.

## Checklist


astaxie (commented):
The implementation and applicable local checks look sound, but two contract/documentation gaps remain before merge:

1. The runtime emits X-RateLimit-Reset-Requests, the token limit/remaining/reset headers, and Retry-After, while the OpenAPI document only declares the request limit and remaining headers on successful responses. Please declare every emitted rate-limit header on the relevant success responses and on the 429 QuotaExceeded response so generated clients see the actual wire contract.
2. Issue #260 also requires the browser/CORS prerequisite to be explicit. Please document the same-origin or gateway CORS configuration required for direct browser clients.

CI is green; these are the remaining review blockers.

backend/internal/server/openapi/gateway.openapi.yaml
```diff
@@ -0,0 +1,1439 @@
...
+              $ref: "#/components/headers/XRequestID"
+            x-ratelimit-limit-requests:
+              $ref: "#/components/headers/XRateLimitLimitRequests"
+            x-ratelimit-remaining-requests:
+              $ref: "#/components/headers/XRateLimitRemainingRequests"
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/AnthropicMessageResponse"
+            text/event-stream:
```
astaxie, line 378: The streaming operations need protocol-specific examples, not just `type: string`. Issue #260 calls out streaming behavior and native Gemini payloads; Chat/Responses include SSE examples, but Anthropic Messages and Gemini `streamGenerateContent` do not show their event frame shape. Please add representative `text/event-stream` examples for these protocols so the interactive docs and generated references explain the actual stream format.

backend/internal/server/openapi/gateway.openapi.yaml
```diff
@@ -0,0 +1,1439 @@
...
+        TokenHub extension equivalent to `Prefer: respond-async` when set to
+        `true`.
+      schema:
+        type: boolean
+  headers:
+    XRequestID:
+      description: TokenHub request identifier.
+      schema:
+        type: string
+    XRateLimitLimitRequests:
```
astaxie, line 1003: This header component set is incomplete for the actual gateway wire contract. `apiKeyRateLimitHeaders`/`combinedRateLimitHeaders` can emit `X-RateLimit-Reset-Requests`, `X-RateLimit-Limit-Tokens`, `X-RateLimit-Remaining-Tokens`, `X-RateLimit-Reset-Tokens`, and `Retry-After` on 429s, but the OpenAPI only defines request limit/remaining and the `QuotaExceeded` responses do not declare any of those headers. Please add all emitted rate-limit headers to the relevant success responses and to the 429 response components so generated clients see the real contract.

docs/deployment.md
```diff
@@ -182,6 +183,8 @@ Edit `deploy/.env` before starting:
 - `TOKENHUB_BACKEND_REPLICAS`: Backend replica count for remote PostgreSQL Compose. Default: `2`.
 - `TOKENHUB_FRONTEND_REPLICAS`: Frontend replica count for remote PostgreSQL Compose. Default: `2`.
 
+The backend serves the public gateway API reference at `/docs` and the machine-readable OpenAPI 3.1 contract at `/openapi.json` and `/openapi.yaml`. The document server URL is derived from `TOKENHUB_PUBLIC_BASE_URL` when set, otherwise from the current request origin. Keep `TOKENHUB_PUBLIC_BASE_URL` aligned with the browser-facing backend origin when the service is behind a reverse proxy.
```
astaxie, line 186: Issue #260 asks the OpenAPI portal docs to document browser and CORS constraints clearly. This section explains the docs URL and public base URL, but it still does not tell browser clients when Try It/direct calls must be same-origin versus when `TOKENHUB_CORS_ALLOWED_ORIGINS` must include the frontend origin. Please add that prerequisite here (and mirror it in the localized docs) so operators do not loosen CORS or deploy the docs on an origin that cannot call the gateway.

backend/internal/server/openapi_http_test.go
```diff
@@ -0,0 +1,891 @@
...
+			}
+		}
+	}
+	return operations
+}
+
+func registeredGatewayOperations(t *testing.T) map[openAPIOperationKey]bool {
+	t.Helper()
+	operations := make(map[openAPIOperationKey]bool)
+	for _, file := range []string{"routes.go", "gemini_native_http.go"} {
```
astaxie: This route-drift guard is too narrow for the requirement because it only scrapes `routes.go` and `gemini_native_http.go` with a few regexes. A future public gateway route registered from another file or via a slightly different helper shape would not be compared against the OpenAPI document, so the test would miss exactly the drift it is meant to catch. Please derive this from the registered route surface or a shared public-gateway route registry instead of a fixed source-file allowlist.

backend/internal/server/openapi/gateway.openapi.yaml
```diff
@@ -0,0 +1,1561 @@
...
+        "200":
+          description: Token count.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/AnthropicTokenCount"
+        "401":
+          $ref: "#/components/responses/AnthropicUnauthorized"
+        "403":
+          $ref: "#/components/responses/ModelNotAllowed"
```
astaxie: This response reuses the OpenAI-style `ModelNotAllowed` schema, but `handleAnthropicCountTokens` writes errors through `writeAnthropicError` for this endpoint, including model-access failures. It also returns Anthropic-formatted 400s for invalid request bodies from `decodeAnthropicMessagesRequest`, but this operation does not document a 400 at all. Please switch the 403 to an Anthropic-compatible response component and add the missing 400 response so generated clients see the actual wire shape for `/v1/messages/count_tokens`.

frontend/e2e/gateway-docs.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+  await expectMinimumContrast(page, [
+    "h1",
+    "a[href='/openapi.json']",
+    ".opblock-summary-path",
+  ]);
+
+  const authorizeButton = page.getByRole("button", { name: "Authorize" });
+  await focusByKeyboard(page, authorizeButton);
+  await page.keyboard.press("Enter");
+  const secretInput = page.locator(".auth-container input[type='password']").first();
```
astaxie: This browser smoke test opens the Authorize dialog but never enters a project key, executes an eligible operation, or asserts that unsafe operations remain disabled. Issue #260 specifically accepts the feature on a developer safely executing an eligible request from the documentation UI, so the main Try It path is still only covered by string/unit checks. Please extend this e2e test with a fake key/request against a harmless route such as `GET /v1/models`, assert the request stays on the documented backend origin, and verify at least one `x-tokenhub-interactive: false` operation cannot be executed.

astaxie:
The PR template asks the Verification section to list checks run and their results, but this PR body currently lists commands only. Please update each item with the outcome, and call out any skipped checks or known environment warnings/failures separately.

backend/internal/server/openapi/gateway.openapi.yaml
```diff
@@ -0,0 +1,1581 @@
...
+                $ref: "#/components/schemas/ImageJob"
+        "400":
+          $ref: "#/components/responses/BadRequest"
+        "401":
+          $ref: "#/components/responses/Unauthorized"
+        "403":
+          $ref: "#/components/responses/ModelNotAllowed"
+        "429":
+          $ref: "#/components/responses/QuotaExceeded"
+        "502":
```
astaxie, line 637: The image edits contract is missing a runtime 501 response. `handleImageEdits` returns `501 image_mask_not_supported` when a request includes `mask` with the Codex image model (`backend/internal/server/image_generation.go`), but this OpenAPI response list only documents 400/401/403/429/502 after the success cases. That means generated clients and the interactive reference will treat a supported runtime failure mode as undocumented. Please add the 501 response here, and consider extending the route/header/error contract test so image edits cannot regress this status again.

backend/internal/server/openapi/gateway.openapi.yaml
```diff
@@ -0,0 +1,1600 @@
...
+        application/json:
+          schema:
+            $ref: "#/components/schemas/ErrorResponse"
+          examples:
+            codexMaskUnsupported:
+              value:
+                error:
+                  code: image_mask_not_supported
+                  message: Masks are not supported through Codex subscription accounts; use reference-image editing without a mask
+                  type: not_implemented
```
astaxie: This example does not match the runtime error envelope. `writeError` sets both `error.code` and `error.type` to the `HTTPError.Code`, and this branch raises `NewHTTPError(..., "image_mask_not_supported", ...)`, so clients will see `type: image_mask_not_supported`, not `not_implemented`. Please update the example, and ideally extend the OpenAPI example checks to assert representative error examples against the actual error envelope shape.

frontend/features/admin/views/gateway-view.tsx
```diff
@@ -106,6 +107,7 @@ function GatewayDocsView({
 
           <section className="api-doc-quick-grid" aria-label={docBundle.quickInfoLabel}>
             <GatewayCopyCard label={docBundle.quickCards.baseURL} value={baseURL} />
+            <GatewayCopyCard label={tx("完整 API Reference")} value={referenceURL} href={referenceURL} />
```
astaxie: This new canonical `tx(...)` key mixes English UI prose into the Simplified Chinese source literal. The repo policy keeps Simplified Chinese as the canonical UI copy and only leaves protocol names, identifiers, units, and user/provider data untranslated. `API` is fine as a protocol acronym, but `Reference` is interface copy here, so please make the key fully localized, for example `tx("完整 API 参考")`, and update the en/ja catalog entries to match.

backend/internal/server/openapi/gateway.openapi.yaml
```diff
@@ -0,0 +1,1618 @@
...
+                type: string
+              examples:
+                messageStart:
+                  value: "event: message_start\ndata: {\"type\":\"message_start\",\"message\":{\"id\":\"msg_123\",\"type\":\"message\",\"role\":\"assistant\",\"content\":[],\"model\":\"claude-sonnet-4.5\",\"stop_reason\":null,\"stop_sequence\":null,\"usage\":{\"input_tokens\":8,\"output_tokens\":0}}}\n\n"
+        "400":
+          $ref: "#/components/responses/AnthropicBadRequest"
+        "401":
+          $ref: "#/components/responses/AnthropicUnauthorized"
+        "403":
+          $ref: "#/components/responses/ModelNotAllowed"
```
astaxie: `POST /v1/messages` returns Anthropic-shaped errors through `writeAnthropicError` for model-access failures, but this 403 response still references the OpenAI-style `ModelNotAllowed` schema/example. That makes the published contract disagree with the runtime payload (`type: "error"`, nested `error.type`, etc.) for Anthropic clients. Please use `#/components/responses/AnthropicModelNotAllowed` here (and add a representative assertion next to the count_tokens check so this does not drift again).

backend/internal/server/openapi/gateway.openapi.yaml
```diff
@@ -0,0 +1,1618 @@
...
+        "429":
+          $ref: "#/components/responses/AnthropicQuotaExceeded"
+        "501":
+          $ref: "#/components/responses/ProviderCapabilityNotSupported"
+        "502":
```
astaxie, line 394: Same shape mismatch here for Anthropic provider/capability failures: `handleAnthropicMessages` writes all errors through `writeAnthropicError`, so 501/502 responses from this endpoint are `AnthropicErrorResponse` payloads, not the OpenAI-style `ErrorResponse` used by `ProviderCapabilityNotSupported` and `ProviderError`. Please either add Anthropic-specific reusable responses for these cases or inline the Anthropic error schema/examples for `/v1/messages`.

backend/internal/server/gemini_native_http.go
```diff
@@ -8,12 +8,15 @@ import (
...
-	s.registerDynamicGETRoute("/v1/models/{model...}", s.handleModelGet, jsonMethodNotAllowed(http.MethodGet))
+	s.registerPublicSingleMethodRoute(http.MethodGet, "/v1/models", s.handleModels, jsonMethodNotAllowed(http.MethodGet))
+	s.registerPublicDynamicGETRoute("/v1/models/{model...}", s.handleModelGet, jsonMethodNotAllowed(http.MethodGet))
 	s.mux.HandleFunc("/v1/models/", s.handleModel)
-	s.registerSingleMethodRoute(http.MethodGet, "/v1beta/models", s.handleGeminiModels, jsonMethodNotAllowed(http.MethodGet))
-	s.registerDynamicGETRoute("/v1beta/models/{model...}", s.handleGeminiModelGetRoute, geminiModelHeadFallback)
+	s.registerPublicSingleMethodRoute(http.MethodGet, "/v1beta/models", s.handleGeminiModels, jsonMethodNotAllowed(http.MethodGet))
+	s.registerPublicDynamicGETRoute("/v1beta/models/{model...}", s.handleGeminiModelGetRoute, geminiModelHeadFallback)
 	s.registerDynamicMethodRoute(http.MethodPost, "/v1beta/models/{model...}", s.handleGeminiModelPostRoute)
+	for _, action := range []string{"generateContent", "streamGenerateContent", "countTokens"} {
```
astaxie, line 17: This PR now documents and registers `streamGenerateContent` as a public Gemini route, but the Gemini model discovery payload still advertises only `generateContent` and `countTokens` in `geminiModelObject`. Gemini clients that rely on `supportedGenerationMethods` can therefore conclude streaming is unsupported even though `POST /v1beta/models/{model}:streamGenerateContent` works and is in the OpenAPI contract. Please include `streamGenerateContent` in the advertised methods and cover it in the Gemini model discovery test.

astaxie (commented):
Re-reviewed the complete current diff at exact head `613136e71485228355f2188bd27b0bb2a2a865d8`. The Gemini model-discovery fix is correct and covered, all existing review threads are resolved, all GitHub CI checks are green, and the full applicable backend/frontend/repository checks plus Compose rendering pass locally.

The current OpenAPI document still has two contract blockers relative to #260:

1. `AnthropicMessagesRequest` and `AnthropicCountTokensRequest` reuse `ChatMessage`, whose role enum includes `system`, `tool`, and `developer`. Runtime validation accepts only `user` and `assistant`, except `system` when the `anthropic-beta` mid-conversation-system feature is explicitly enabled. Neither Anthropic operation documents `anthropic-version`/`anthropic-beta`. Generated clients therefore permit requests TokenHub rejects and cannot discover the condition for system-role messages. Please add Anthropic-specific message/content schemas and document the relevant headers.

2. The central Chat/Responses contracts remain permissive outlines rather than inspectable/generatable schemas. For example, `ResponsesRequest` omits implemented fields such as `max_output_tokens`, `temperature`, `instructions`, `store`, `reasoning`, and `service_tier`; `input` is unconstrained, while response `output` and Chat `choices` are arbitrary objects. Issue #260 asks for request/response schema inspection and SDK/tool import, and this PR closes that issue as a complete public specification. Please model the supported request/content/tool/output shapes with reusable schemas, or explicitly narrow this PR to a non-closing first phase and leave #260 open.

Local verification: focused OpenAPI/Gemini tests, `go test ./...`, `go vet ./...`, 127 repository tests, all translation/environment/source-line gates, frontend lint/typecheck/tests/build, browser smoke 4/4, Compose render, and `git diff --check` passed.

backend/internal/server/openapi/gateway.openapi.yaml
```diff
@@ -0,0 +1,2326 @@
...
+          type: string
+          enum: [minimal, low, medium, high, xhigh]
+        metadata:
+          type: object
+          additionalProperties: true
+        prompt_cache_key: {}
+        user: {}
+    ChatMessage:
+      type: object
+      required: [role, content]
```
astaxie: `content` should not be required for every chat message. TokenHub accepts and forwards assistant tool-call messages without a `content` field (`ChatMessage.Content` can stay nil and `MarshalJSON` deletes it), and the existing Gemini conversion test builds exactly that shape. Requiring `content` here makes the OpenAPI contract narrower than the runtime/OpenAI-compatible tool-call flow, so generated clients can reject valid requests. Please make `content` optional for chat messages, or model the assistant/tool-call case separately.

backend/internal/server/openapi/gateway.openapi.yaml
```diff
@@ -0,0 +1,2326 @@
...
+            $ref: "#/components/schemas/ChatTool"
+        tool_choice:
+          $ref: "#/components/schemas/ChatToolChoice"
+        parallel_tool_calls:
+          type: boolean
+        response_format:
+          $ref: "#/components/schemas/ChatResponseFormat"
+        reasoning_effort:
+          type: string
+          enum: [minimal, low, medium, high, xhigh]
```
astaxie: `reasoning_effort` is too narrow here: TokenHub normalizes and forwards non-empty effort values, and the Anthropic support matrix/tests explicitly accept `max` for current Claude model families. The same enum is reused under `ResponsesReasoning.effort`, so generated Chat/Responses clients will reject valid requests such as `reasoning_effort: "max"` or `reasoning.effort: "max"`. Please include `max` in both reasoning effort enums, or remove the enum if this gateway surface is intentionally provider-extensible.

backend/internal/server/openapi/gateway.openapi.yaml
```diff
@@ -0,0 +1,2326 @@
...
+        encrypted_content:
+          type: string
+    ResponseTool:
+      type: object
+      required: [type, name]
+      additionalProperties: true
+      properties:
+        type:
+          type: string
+          const: function
```
astaxie: This schema is still narrower than the Responses API surface TokenHub forwards. `ResponsesRequest` preserves unknown raw JSON fields and `OpenAICompatibleAdapter.Responses` sends them upstream, so requests with built-in Responses tools such as `{ "type": "web_search" }` or `file_search` can pass through the gateway. Modeling `ResponseTool.type` as `const: function` makes generated clients reject those valid passthrough requests. Please either add the supported built-in tool shapes here or make the tool item extensible enough for OpenAI Responses tools.

backend/internal/server/openapi/gateway.openapi.yaml
```diff
@@ -0,0 +1,2328 @@
...
+            $ref: "#/components/schemas/ChatTool"
+        tool_choice:
+          $ref: "#/components/schemas/ChatToolChoice"
+        parallel_tool_calls:
+          type: boolean
+        response_format:
+          $ref: "#/components/schemas/ChatResponseFormat"
+        reasoning_effort:
+          type: string
+          enum: [minimal, low, medium, high, xhigh, max]
```
astaxie: The enum is still narrower than the gateway behavior because it omits `none`. TokenHub accepts `none` today: Codex account validation includes it, `codexReasoningEffort` maps it through for Chat-to-Responses conversion, and Gemini 2.5 Flash maps `reasoning_effort: "none"` to `thinkingBudget: 0`. The same closed enum is also used for `ResponsesReasoning.effort`, so generated Chat/Responses clients will still reject valid requests. Please include `none` in both reasoning effort enums, or make the field provider-extensible instead of a closed enum.

backend/internal/server/openapi/gateway.openapi.yaml
```diff
@@ -0,0 +1,2328 @@
...
+          type: string
+          enum: [assistant]
+        content:
+          type: string
+        reasoning_content:
+          type: string
+        tool_calls:
+          type: array
+          items:
+            $ref: "#/components/schemas/ChatToolCall"
```
astaxie: `ChatCompletionDelta.tool_calls` should not reuse the full `ChatToolCall` schema. Streaming tool-call argument chunks emitted by TokenHub contain only `index` plus `function.arguments` after the start chunk; they do not repeat `id`, `type`, or `function.name`. The existing stream encoder and Anthropic stream tests exercise that shape. With this `$ref`, generated SSE clients will reject valid `chat.completion.chunk` frames. Please add a separate delta tool-call schema where `index` is present and the other fields are optional.

backend/internal/server/openapi/gateway.openapi.yaml
```diff
@@ -0,0 +1,2328 @@
...
+
+    Entered project API keys are bearer tokens. Documentation clients must keep them in memory,
+    never in URLs, local storage, logs, examples, or exported files.
+  version: 0.6.0
+  x-tokenhub-contract-version: gateway-2026-08-issue-260
+jsonSchemaDialect: https://spec.openapis.org/oas/3.1/dialect/base
+servers:
+  - url: https://tokenhub.example
+    description: Replaced at runtime with TOKENHUB_PUBLIC_BASE_URL or the current request origin.
+security:
```
astaxie, line 18: The root-level security requirement makes all operations advertise all three auth schemes (`Authorization`, `x-api-key`, and `x-goog-api-key`). That loses the protocol-specific contract the issue asks for: OpenAI-compatible operations should primarily document bearer auth, Anthropic operations should document `x-api-key` as the native alternative, and Gemini operations should document `x-goog-api-key` as the native alternative. Keeping these schemes global will make generated clients and the Swagger UI show Anthropic/Gemini-native headers on unrelated operations. Please move the protocol-native schemes to the relevant operation groups, or add operation-level `security` overrides while keeping bearer auth as the common TokenHub key path.

backend/internal/server/openapi/gateway.openapi.yaml
```diff
@@ -0,0 +1,2328 @@
...
+        index:
+          type: integer
+        message:
+          $ref: "#/components/schemas/ChatCompletionMessage"
+        delta:
+          $ref: "#/components/schemas/ChatCompletionDelta"
+        finish_reason:
+          oneOf:
+            - type: string
+              enum: [stop, length, tool_calls, content_filter]
```
astaxie: This enum omits the legacy `function_call` finish reason even though TokenHub can still pass it through. `openAICompatibleCore.chat` returns the upstream Chat Completions body unchanged, and the finish-reason helper explicitly keeps `function_call` because older upstreams still emit it. With this closed enum, generated clients can reject a valid buffered `/v1/chat/completions` response. Please include `function_call` or make `finish_reason` an extensible string while documenting the common values.

backend/internal/server/openapi_http.go
```diff
@@ -0,0 +1,351 @@
...
+    }
+
+    const authInputObserver = new MutationObserver(hardenSwaggerAuthInputs);
+    authInputObserver.observe(document.body, { childList: true, subtree: true });
+
+    fetch(openAPIURL, { credentials: "same-origin" })
+      .then((response) => response.json())
+      .then((spec) => {
+        tokenhubSpec = spec;
+        SwaggerUIBundle({
```
astaxie, line 354: This still leaks the entered project key after an interactive request. Even with `persistAuthorization: false`, `requestSnippetsEnabled: false`, and `showMutatedRequest: false`, Swagger UI renders the executed Curl block with `Authorization: Bearer <project key>` after `GET /v1/models` runs. I verified this in the PR e2e stack by authorizing, executing `/v1/models`, and checking the DOM; the Curl area contained the generated key. Issue #260 requires entered keys to stay in memory only and not appear in examples or generated outputs, so the docs UI needs to disable or redact the Curl/request display for authorized executions before this is safe.

backend/internal/server/openapi/gateway.openapi.yaml
```diff
@@ -0,0 +1,2367 @@
...
+          type: string
+          nullable: true
+    AnthropicErrorResponse:
+      type: object
+      required: [type, error, request_id]
+      properties:
+        type:
+          type: string
+          const: error
+        error:
```
astaxie, line 1280: [P2] This schema omits `error.code`, but the Anthropic response examples include it and `writeAnthropicError` always emits it from `HTTPError.Code`. Redocly therefore flags the Anthropic error examples as invalid against `AnthropicErrorResponse`, and generated clients will miss a runtime field. Please add `code` (and likely `details` as optional) under the nested Anthropic `error` schema so the contract matches both the examples and runtime payload.

backend/internal/server/openapi/gateway.openapi.yaml
```diff
@@ -0,0 +1,2367 @@
...
+        code:
+          type: string
+          examples: [invalid_api_key, quota_exceeded, provider_error]
+        message:
+          type: string
+        type:
+          type: string
+        param:
+          type: string
+          nullable: true
```
astaxie: [P1] `nullable` is no longer valid for OpenAPI 3.1 schemas. Running `npx -y @redocly/cli@latest lint backend/internal/server/openapi/gateway.openapi.yaml` fails on this field and the same pattern at `ChatCompletionChoice.logprobs`, so the published `/openapi.json` contract does not satisfy issue #260 requirement for a valid OpenAPI 3.1 document. Please model nullable fields with JSON Schema unions such as `type: ["string", "null"]` and give `logprobs` an explicit schema instead of using the OpenAPI 3.0 `nullable` keyword.

Mr0bean:
Follow-up review of the final merged commit:

- [P2] [The multipart image-edit example](https://github.com/astaxie/TokenHub/blob/c533b01c7f50089e57f44606cdbb3ea2b8c59b33/backend/internal/server/openapi/gateway.openapi.yaml#L578) uses `gpt-image-1`, but `normalizeImageGenerationRequest` accepts only `codex-gpt-image-2` and `gpt-image-2`. The published example therefore always returns `400 unsupported_image_model` before reaching image-edit logic. Please use a supported model in the executable example.

- [P2] [`ResponseInputMessage`](https://github.com/astaxie/TokenHub/blob/c533b01c7f50089e57f44606cdbb3ea2b8c59b33/backend/internal/server/openapi/gateway.openapi.yaml#L1702) requires `type`, but OpenAI-compatible callers may send a Responses input message as `{ role, content }` without it. TokenHub preserves and forwards that raw input, so generated clients reject a valid gateway request. Make `type` optional for this branch, or add a shorthand-message branch, and cover it with a contract test.

- [P1] [`ResponseOutputItem`](https://github.com/astaxie/TokenHub/blob/c533b01c7f50089e57f44606cdbb3ea2b8c59b33/backend/internal/server/openapi/gateway.openapi.yaml#L1890) is a closed `oneOf` that contradicts the built-in Responses tools advertised by the same contract. The OpenAI-compatible adapter returns upstream Responses bodies unchanged, so successful web-search, file-search, computer, and code-interpreter calls contain output items that cannot match any branch. Add their schemas or an extensible fallback, and cover one built-in-tool response in the contract test.

- [P2] [The image request schemas](https://github.com/astaxie/TokenHub/blob/c533b01c7f50089e57f44606cdbb3ea2b8c59b33/backend/internal/server/openapi/gateway.openapi.yaml#L2205) omit supported fields: the runtime validates `quality` (`auto`, `low`, `medium`, `high`) for both generations and edits, while edits also accept `n` and `response_format`. The public reference and generated clients cannot discover these controls. Model the shared fields and constraints in both schemas.

- [P2] [The new CORS guidance](https://github.com/astaxie/TokenHub/blob/c533b01c7f50089e57f44606cdbb3ea2b8c59b33/docs/deployment.md#L188) does not match `Server.cors`: unlisted origins still receive `Access-Control-Allow-Origin: *` and reflected requested headers. Since the Swagger interceptor uses `credentials: "omit"`, Bearer/x-api-key requests can complete cross-origin without adding the origin. Either reject unlisted origins in the middleware, or revise the guidance; otherwise operators will believe the allowlist restricts browser gateway access when it does not.

Issue #169, opened 2026-08-08
refactor(backend): split the server package into domain modules


astaxie:
## Background

Most backend code currently lives in `backend/internal/server`: approximately 179 Go files and nearly 80,000 lines, all in the same `server` package. Responsibilities are distinguished mainly by filename prefixes such as `provider_*`, `admin_*`, `store_*`, and `billing_*`.

This organization was straightforward early in the project, but the single package has become a development and maintenance bottleneck as gateway, provider adaptation, routing, project/key management, billing, reconciliation, and analytics capabilities grow. Every change requires searching many files in the same package; any package symbol can be referenced anywhere, so the compiler cannot enforce responsibility boundaries or dependency direction.

## Current problems

- Module ownership depends on filename prefixes and developer knowledge; directories and packages do not express domain boundaries.
- `Server` is responsible for HTTP composition, protocol adaptation, provider invocation, routing, billing, reconciliation, image jobs, observability, and persistence access.
- The `Store` interface has more than 120 methods and mixes persistence capabilities from projects, keys, usage, billing, reconciliation, and other domains.
- Dependencies within the same package can spread freely, encouraging circular conceptual dependencies and implicit coupling.
- Tests easily depend on implementation details, requiring an overly broad code surface to be understood and verified when one domain changes.
- New capabilities usually add more prefixed files, with no structural constraint preventing further package growth.

## Goal

Split the backend into multiple Go packages with clear domain responsibilities so code structure and the compiler enforce module boundaries, public capabilities, and dependency direction.

This refactor should ensure:

- `server` retains only application composition, dependency wiring, and HTTP route assembly.
- HTTP handlers only parse protocols, validate input, call application services, and transform responses; they contain no domain logic.
- Each module defines the narrow interfaces it needs instead of depending on one global, oversized `Store`.
- GORM/SQLite/PostgreSQL implementations act as persistence adapters that depend on persistence ports defined by modules.
- Protocol, provider, routing, and business domains have one-way, explainable dependency relationships.
- Tests target public module interfaces and observable behavior rather than internal implementation.
- Migration is incremental, and every phase can be merged, run, and rolled back independently.

## Proposed modules

The following target responsibilities are suggested. Directory names may be refined during implementation design, but the boundary principles should remain:

- `gateway`: protocol-independent request lifecycle and gateway orchestration.
- `protocol/openai`, `protocol/anthropic`, `protocol/gemini`: external protocol parsing, conversion, and response adaptation.
[...]

PR #296, opened 2026-08-25
refactor(reconciliation): move persistence and admin adapters

bailu-ZZ:
## Summary

This refactor moves reconciliation persistence and administrative HTTP handling out of the server package. It establishes a dedicated reconciliation boundary with module-owned persistence interfaces while preserving existing API behavior and business rules.

## Related Issue

Related #169

## Changes

- Add the reconciliation domain service, audit handling, and narrow persistence ports.
- Add GORM persistence adapters and PostgreSQL integration coverage for reconciliation data.
- Move reconciliation administrative endpoints into the admin package.
- Remove the temporary server reconciliation bridge and obsolete server implementations.
- Preserve existing response formats, error behavior, audit semantics, and billing compatibility.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor or maintenance
- [ ] Documentation

## Verification

- `go test ./...` pass.
- `go vet ./...` pass.
- `node --test tools/*.test.mjs` pass.
- `node tools/check-doc-translations.mjs` pass.
- `node tools/check-ui-translations.mjs` pass.
- `node tools/check-env-contract.mjs` pass.
- `node tools/check-source-lines.mjs` pass.
- `gofmt` and `git diff --check` pass.
- Frontend and SDK suites were not run because this PR only changes backend and repository architecture code.

This is an internal backend refactor with no intentional changes to public API responses, database schema, authentication or authorization behavior, billing rules, credential handling, or operational configuration. Existing reconciliation and billing behavior remains compatible, and the change can be rolled back by reverting this commit.

## Checklist


astaxie (commented):
Reviewed the complete current diff at head a03da8896. I found no code or spec defect, and the full backend test suite passes locally; all current GitHub checks are green. Approval is withheld because GitHub reports this head conflicts with main, so the resolved merge result still needs review and validation.

backend/internal/reconciliation/persistence/billing.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func (b *BillingReader) ListRecordsInRange(connectorID string, from, to time.Time) ([]reconciliation.BillingRecord, error) {
+	records, err := b.reader.ListBillingRecordsInRange(connectorID, from, to)
+	if err != nil {
+		return nil, err
+	}
+	result := make([]reconciliation.BillingRecord, len(records))
+	for i, record := range records {
+		result[i] = reconciliation.BillingRecord{ID: record.ID, ExternalID: record.ExternalID, SourceType: record.SourceType, AccountID: record.AccountID, ProviderID: record.Metadata["provider_id"], ProviderResourceID: record.Metadata["provider_resource_id"], ResourceID: record.Metadata["resource_id"], ProjectID: record.Metadata["project_id"], Model: record.Model, Currency: record.Currency, NetAmount: record.NetAmount, UsageStartAt: record.UsageStartAt, ExternalRequestID: record.ExternalRequestID}
```
astaxie: [P1] Preserve authoritative attribution snapshots

The old production bridge preferred `tokenhub_provider_id` / `tokenhub_resource_id` and treated a present empty snapshot as authoritative. This adapter reads only the mutable legacy fields, so a billing record whose snapshot says `snapshot-provider` but legacy metadata says `legacy-provider` is now reconciled against the latter. I reproduced that result through `BillingReader`. Restore snapshot-first lookup, including the intentionally-empty case, and move the compatibility regression onto the production adapter.

backend/internal/reconciliation/persistence/store.go
```diff
@@ -0,0 +1,337 @@
...
+		from = from.Add(-window)
+		to = to.Add(window)
+	}
+	var rows []usageRow
+	if err := s.db.Where("created_at >= ? AND created_at < ?", from.UTC(), to.UTC()).Order("created_at asc, id asc").Find(&rows).Error; err != nil {
+		return nil, err
+	}
+	result := make([]reconciliation.Usage, len(rows))
+	for i, row := range rows {
+		result[i] = usageFromRecord(row.ID, row.RequestID, row.ProjectID, row.ModelName, row.ProviderID, row.ProviderResourceID, row.CostUSD, row.ProviderCostUSD, row.CreatedAt)
```
astaxie, line 155: [P1] Reconnect explicit zero-cost evidence

Every usage returned here has `ProviderCostKnown == false`; the old bridge set it for nonzero costs and called `applyZeroCostEvidence` for an explicitly free final attempt. A temporary integration test persisted matching `attempt_prepared` plus versioned zero-dollar `shadow_settlement` evidence, then called this production store: it still returned false, so reconciliation rejects the request as `reconciliation_provider_cost_unknown`. Carry the presence/evidence lookup through this boundary and add a production-path regression for zero and nonzero costs.

backend/internal/reconciliation/service.go
```diff
@@ -27,6 +28,67 @@ func NewService(store Store, billingReader BillingReader) *Service {
...
+	items, total := s.store.ListItems(id, status, limit, offset)
+	return run, items, total, nil
+}
+
+func (s *Service) GetRunForAdmin(id string) (Run, error) { return s.store.GetRun(id) }
+
+func (s *Service) Lock(runID, actor string) (before, after Run, err error) {
+	s.runLockMu.Lock()
+	defer s.runLockMu.Unlock()
+	before, err = s.store.GetRun(runID)
```
astaxie, line 51: [P1] Make lock and recalculation serialization shared

`runLockMu` protects only concurrent `Lock` calls; it does not coordinate with `Recalculate`, and `SaveRunLock` later saves the entire stale row. I forced Lock to read first, let Recalculate replace the run/items, then released Lock: the final row regained the old totals and input hash while keeping the recalculated items. The old store operation held the persistence mutex across read/prepare/write. Make this a single conditional update/transaction or share serialization with replacement so a lock cannot overwrite recalculation results.

astaxie (changes requested):
Reviewed the complete current diff at head 6291731f6 against current main and related issue #169. Three compatibility/data-integrity regressions remain inline. Temporary regression tests reproduced all three: billing attribution returned the mutable legacy IDs instead of the authoritative snapshots; explicit versioned zero-cost evidence still produced ProviderCostKnown=false; and an ordered Lock/Recalculate race restored stale totals and the old input hash after recalculation. The existing full Go suite, vet, repository gates, and all current GitHub checks pass, so these cases need production fixes and permanent coverage before approval.

backend/internal/reconciliation/persistence/cost_evidence.go
```diff
@@ -0,0 +1,102 @@
...
+	for _, usage := range usages {
+		if usage.ProviderCostUSD == 0 && strings.TrimSpace(usage.RequestID) != "" {
+			ids = append(ids, usage.RequestID)
+		}
+	}
+	if len(ids) == 0 {
+		return nil
+	}
+	var rows []meteringEvidenceRow
+	if err := db.Where("scope IN ? AND kind IN ?", ids, []string{"attempt_prepared", "shadow_settlement"}).Find(&rows).Error; err != nil {
```
astaxie: [P1] Keep evidence lookups below database parameter limits

The extracted implementation dropped the old 500-usage batching and expands every request ID into one `IN` clause. I exercised this production helper with 40,000 zero-cost usages on the repository's SQLite driver; it fails here with `too many SQL variables`, so a sufficiently large reconciliation window now returns an error instead of reconciling. Restore bounded batches (and preferably deduplicate request IDs) as the pre-refactor implementation did, with a regression above the driver limit.

backend/internal/server/http.go
```diff
@@ -82,26 +87,37 @@ func New(store Store) *Server {
...
-	billingAvailable := billingDependencies.Repository != nil && billingDependencies.ReconciliationReader != nil
-	billingDependencies = normalizeBillingDependencies(billingDependencies)
+// NewWithConfigAndBillingDependencies is retained for source compatibility.
+func NewWithConfigAndBillingDependencies(store Store, config Config, dependencies BillingDependencies) *Server {
+	return NewWithConfigAndDependencies(store, config, dependencies)
+}
+
+func newWithConfig(store Store, config Config, dependencies ApplicationDependencies) *Server {
+	billingAvailable := dependencies.Repository != nil && dependencies.ReconciliationReader != nil
+	reconciliationReadAvailable := dependencies.ReconciliationStore != nil
```
astaxie, line 122: [P2] Preserve reconciliation for the legacy billing constructor

`NewWithConfigAndBillingDependencies` is explicitly retained for compatibility, but callers using its original `{Repository, ReconciliationReader}` shape cannot supply this newly added field. They still compile, yet this flag disables every reconciliation route. A temporary regression using the existing Store decorator returned 503 `reconciliation_store_unavailable`; before this refactor the embedded Store served those routes. Derive the reconciliation adapter from a compatible store or otherwise preserve the old constructor's behavior.

astaxie (changes requested):
Re-reviewed the complete current diff at head 2a3225c63 against current main and issue #169. The latest commit fixes all three prior P1 findings, so those threads can be resolved. Two additional compatibility failures remain inline: large evidence lookups exceed the database parameter limit, and the retained legacy billing constructor disables reconciliation when called with its original dependency shape. Temporary regression tests reproduced both failures. The full Go suite and vet pass after removing those tests; 207 repository tests, all root gates, git diff --check, and all nine GitHub checks pass.

astaxie (commented):
Re-reviewed the complete current diff at head a4fdda576. The bounded, deduplicated evidence lookup and legacy billing-constructor compatibility fixes address both findings from the previous round; the focused persistence/server regressions pass, so those two threads are resolved. No additional code issue was found. This head is still not eligible for approval because GitHub reports merge conflicts with current main; the resolved merge result needs review and applicable validation.

backend/internal/reconciliation/persistence/postgres_integration_test.go
```diff
@@ -0,0 +1,86 @@
+//go:build integration
```
astaxie: [P1] Run this PostgreSQL adapter contract in CI

This file is guarded by the `integration` build tag, so the ordinary `go test ./...` job excludes it. The PostgreSQL job in `.github/workflows/ci.yml:152-157` only invokes tagged tests in `./internal/server` and `./internal/dbschema`, which means this new package and its concurrent `BackfillRuleConnectorSnapshot` contract are never executed by CI despite the PR's stated PostgreSQL integration coverage. Add a tagged `./internal/reconciliation/persistence` invocation to that job and verify it against the PostgreSQL service.

astaxie (changes requested):
Re-reviewed the complete current diff at head `768de8690d`. The merge conflict is resolved cleanly and the prior compatibility fixes remain intact, but two current acceptance gaps block approval.

The PostgreSQL coverage gap is inline. Separately, issue #169 requires each migration phase to document dependency changes and remaining work and requires synchronized architecture documentation. The current `docs/architecture.md:127-136`, `docs/zh-CN/architecture.md:127-137`, and `docs/ja/architecture.md:127-136` still describe only the monolithic server/store ownership and do not document the new `admin`, `reconciliation`, and reconciliation persistence boundaries. Please update all three architecture documents and identify the next migration phase.

Local validation passed the full backend test suite, vet, golangci-lint v2.12.2, 175 repository tests, all root gates, and `git diff --check`. All nine GitHub checks are green. The new tagged PostgreSQL adapter test could not be run locally because Docker/PostgreSQL was unavailable; importantly, the current green PostgreSQL job does not execute it.

PR #333, opened 2026-09-09
fix(api-keys): show setup details and confirm rotation

deepjerry-ai:
## Summary

API Key setup was split from key issuance, and clicking Rotate immediately revoked the old key. Add a reusable setup dialog that opens automatically after creation or rotation, and require explicit confirmation before rotation.
image
image

## Related Issue

N/A — requested UI improvement.

## Changes

- Add a **Use** action with protocol-specific Base URLs, authentication headers, copy controls, and cURL examples for OpenAI Chat Completions / Responses, Anthropic Messages, and Gemini.
- Carry the newly issued key directly into setup. Closing clears it; reopening a saved key shows a placeholder and never retrieves or stores its full value.
- Confirm the target key and immediate revocation before rotation. Cancel sends no mutation; submission is guarded against repeat clicks. Clear pending confirmation on logout, authentication expiry, and view changes.
- Add keyboard focus containment, clipboard failure feedback, inactive-key warnings, focus retention while all confirmation actions are disabled, responsive layouts, and English / Chinese / Japanese copy and user documentation.
- Fix two test-isolation defects exposed during verification: distinguish `/api-keys` pages from `/api/` requests, and keep parent-directory TypeScript imports inside each test's private output directory. A regression test reproduces the latter failure with the base-branch loader.

## Type of Change

- [x] Bug fix
- [x] New feature
- [x] Refactor or maintenance
- [x] Documentation

## Verification

- Node.js 22.23.1: frontend lint and typecheck passed.
- Frontend domain tests: 181 passed. Component tests: 289 passed across 53 files.
- Production build and complete database-free UI acceptance: 32 passed; the 12 API Key/isolation scenarios also passed after the final focus guard. Reviewed desktop/mobile setup, creation, rotation confirmation, success, and failure screenshots with synthetic keys.
- Real-backend browser smoke: 10 passed, including creation-to-setup and logout/relogin clearing pending confirmation.
- Repository gate tests: 175 passed. UI translations, full three-language document co-change check, environment contract, source-line limits, and `git diff --check` passed.
- Independent code/coverage review completed with no remaining blocking findings.
- No backend implementation changed; backend CI remains the contract check.

No backend API, schema, authentication scope, credential persistence, environment, or deployment configuration changes. The full key stays in transient UI state until closing or sign-out; existing keys remain hash-only and cannot be recovered. Rotation still immediately revokes the old key, so callers must update their configuration after confirming. Examples require a model allowed by the project and key. The dialog does not send model requests.

## Checklist


frontend/features/admin/domain/api-key-access.ts
```diff
@@ -0,0 +1,31 @@
...
+
+function shellQuote(value: string) {
+  return `'${value.replaceAll("'", "'\\''")}'`;
+}
+
+export function apiKeyAccessConfig(baseURL: string, protocol: AccessProtocol, secret: string, model: string) {
+  const root = baseURL.trim().replace(/\/+$/, "").replace(/\/(v1|v1beta)$/, "");
+  const modelID = model.trim() || modelPlaceholder;
+  const key = secret || apiKeyPlaceholder;
+  const base = protocol === "anthropic" ? root : `${root}/${protocol === "gemini" ? "v1beta" : "v1"}`;
```
astaxie: **[P2] Use the gateway root for Gemini's client Base URL**

Selecting Gemini currently makes the copyable Base URL `https://gateway/v1beta`. Gemini CLI passes `GOOGLE_GEMINI_BASE_URL` to GoogleGenAI's `httpOptions.baseUrl`, and that SDK appends its API version itself, so copying this value produces `/v1beta/v1beta/models/...` and a 404. This also contradicts `docs/gemini-cli-codex-subscription.md`, which explicitly says not to append `/v1beta`. Please return `root` for Gemini's client Base URL, retaining `/v1beta` in the request endpoint/cURL example, and update the corresponding tests and copy.

frontend/features/admin/shell/admin-console.tsx
```diff
@@ -146,6 +148,7 @@ export function AdminConsole({ defaultBaseURL }: { defaultBaseURL: string }) {
 
   const selectView = useCallback((view: ViewKey, options: { replace?: boolean; routeModelQuery?: string; pluginPageKey?: string } = {}) => {
     if (view !== activeView) {
+      setPendingAction(null);
```
astaxie, line 151: **[P2] Clear pending rotation on browser history navigation too**

This cleanup only runs through `selectView`; browser Back/Forward uses the `routeView` effect, which calls `setActiveView` directly. Reproduced in the real UI with `Cost Centers → Key Management → Rotate → browser Back`: the URL and underlying page return to `/cost-centers`, but the old key's confirmation remains open and actionable. This misses the promised cancellation on view changes and leaves a destructive action attached to a page the user has left. Please clear the pending action on actual route/view changes as well, and cover history navigation in the acceptance test.

astaxie (commented):
Reviewed head 3cd420a7d28e189758a88bfb16354ad0371f8500 and found two P2 issues, detailed inline. No newly introduced repository-standards violations found.

Validation: 181 domain tests, 11 focused component tests, lint, and typecheck passed locally. Under Node.js 22.23.1, all five API key UI acceptance scenarios passed, while an additional browser Back regression reproduced the retained rotation confirmation. The current head's CI checks are all passing.

frontend/features/admin/domain/api-key-access.ts
astaxie: Verified fixed in 89c6675a3f939f709bcb1c396bb01e10ec7cdd40. Gemini now exposes the gateway root as the client Base URL, while its request URL retains /v1beta. The focused component tests and browser setup scenario pass, and all three user guides are synchronized.

frontend/features/admin/shell/admin-console.tsx
astaxie: **[P2] The history fix still misses navigation within Key Management**

Rechecked 89c6675a3f939f709bcb1c396bb01e10ec7cdd40: the Cost Centers case now passes, but the confirmation remains for this real-browser sequence:

1. Open Key A's usage page at `/api-keys/key_review_a/usage`.
2. Click **Key Management** in the sidebar.
3. Open **Rotate** for Key B.
4. Press browser **Back**.

The URL and underlying page return to Key A's usage, but Key B's rotation confirmation remains open and actionable. `viewFromPath()` returns `api-keys` for both the list and usage routes, so the effect at `admin-console.tsx:355-365`, which depends only on `[routeView]`, never executes its new cleanup. Please tie pending-action cleanup to actual pathname changes as well, and cover list-to-usage Back/Forward navigation.

Reproduced with the production frontend and successful, typed synthetic usage responses: `expect(dialog).toHaveCount(0)` receives `1`, with no fixture violations or application errors. The PR's six API key UI scenarios and 11 focused component tests pass, and the current head's CI is green.

astaxie: Verified fixed in 98a59583aa6fda0d82390124c4ad0db3b7780e28. Adding pathname to the effect dependencies now clears the pending rotation when navigating between the key list and a key's usage page.

I reran the exact independent regression that failed in the previous review (Key A usage → Key Management → Rotate Key B → browser Back); it now passes. All seven PR API key UI scenarios also pass, including Back/Forward coverage and the assertion that navigation sends no POST, for eight passing browser checks under Node.js 22.23.1. The production build and targeted lint pass, and CI for this exact head has completed successfully.

No new actionable findings from the incremental standards and behavior review. Both previously reported issues are now verified fixed.

PR #339, opened 2026-09-13
feat(website): add docusaurus documentation site with pages deploy

jackiesre721:
> Write the PR title and every body section in English.
> PR title format: `<type>[optional scope][!]: <short summary>`, max 72 characters.
> Common types include `feat`, `fix`, `docs`, `refactor`, `perf`, `test`, `build`, `ci`, `chore`, `style`, and `revert`. Use a lowercase imperative summary without a trailing period.

## Summary

The repository documentation lives as plain markdown in `docs/` with no browsable entry point. This PR adds a Docusaurus 3 documentation website under `website/` that renders the repository documentation as a searchable trilingual handbook (English, Simplified Chinese, Japanese) and deploys it to GitHub Pages automatically. The site has no copy of the documentation: page trees are generated from `docs/`, `docs/zh-CN/`, and `docs/ja/` on every build, so the website can never drift from the canonical markdown.

## Related Issue

N/A

## Changes

- Add `website/`, a Docusaurus 3 docs-only site with brand theming, mermaid diagram rendering, and a locale dropdown (en / zh-Hans / ja).
- Add `website/scripts/sync-docs.mjs`, which generates the default and i18n page trees from the repository docs on every start and build: it strips the per-file language switcher lines (including the localized `语言：` / `言語：` variants), rewrites cross-document links to the generated locations, points links that leave `docs/` and unpublished locale-only pages at GitHub, injects frontmatter (sidebar position, landing slug, English titles), and writes per-locale sidebar translations. Pages without a translated source fall back to English content so locale builds never break links.
- Declare the site layout (landing page plus eight sidebar sections, 30 pages per locale) in the `SECTIONS` table inside the sync script; generated trees and the edit-URL source map are git-ignored.
- Add `.github/workflows/docs-deploy.yml`, which builds the site on pushes to `main` touching `docs/` or `website/` and publishes it to GitHub Pages with `actions/deploy-pages` (requires the repository Pages source to be set to "GitHub Actions").
- Document site commands, deployment, and how to add pages in `website/README.md`.

- Commit `6ce08d19` (second review round): sanitize the U+0000 markers that react-dom 18's `renderToPipeableStream` emits into the written HTML (they corrupted heading `id` attributes and broke in-page anchors on localized pages), restore Japanese kana in search via a dependency patch that tokenizes kana runs with the lunr Japanese tokenizer, point "Edit this page" of untranslated fallback pages at the English source via per-locale flags in the source map, localize the footer with a swizzled `src/theme/Footer` plus per-locale `code.json` written by the sync script, and add `docs-website-build.yml` so pull requests touching `docs/` or `website/` actually build the site.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor or maintenance
- [ ] Documentation

## Verification

- `npm run typecheck` and `npm run build` in `website/` pass; the build compiles all three locales with `onBrokenLinks` and broken markdown link checks set to throw (30 pages per locale, 29 translated plus 1 English fallback for zh-Hans and ja).
- Served the production build locally and verified in a browser: landing page, sidebar sections, embedded screenshots, mermaid architecture and deployment diagrams, prev/next pagination, dark mode, and the locale dropdown switching to the same page in each language.
- Repository gates pass: `node --test tools/*.test.mjs` (175 tests), `node tools/check-doc-translations.mjs`, `node tools/check-ui-translations.mjs`, `node tools/check-env-contract.mjs`, `node tools/check-source-lines.mjs`, and `git diff --check`.
- Not run: the frontend/backend test suites and SDK smoke tests; this change adds a self-contained `website/` directory and one workflow file and touches no backend, frontend, or SDK code. The Pages deployment itself can only be observed after merge on GitHub.

- No application API, persistence, or runtime behavior changes; the site is a static build served by GitHub Pages.
- The workflow runs on `ubuntu-latest` with pinned actions (`actions/checkout@v4`, `actions/setup-node@v4`, `actions/upload-pages-artifact@v3`, `actions/deploy-pages@v4`), least-privilege permissions (`contents: read`, `pages: write`, `id-token: write`), and a cancel-in-progress `pages` concurrency group.
- Rollout: enable GitHub Pages with the "GitHub Actions" source; the site appears at `https://astaxie.github.io/TokenHub/` after the first run. Rollback by disabling the workflow; no data or configuration is migrated.
- No environment variable, credential, or database changes. Generated trees (`website/docs/`, `website/i18n/`, `website/build/`, `node_modules/`) are git-ignored.

- Second round: `npm ci && npm run build` from a clean install passes for all three locales; the sanitizer step reports the stripped markers and the build fails if any NUL remains. Browser verification against the production build: clicking the table-of-contents entry for a previously corrupted heading on the ja user guide scrolls to the correct anchor; searching the ja site for「リバース」and「データベース」returns results; searching zh-Hans for「反向代理」and the en site for "gateway" still return results; the footer renders translated on zh-Hans and ja and English on the default locale; the untranslated PostgreSQL Setup page links "Edit this page" to the English source.

## Checklist


website/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "tokenhub-docs",
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "sync-docs": "node scripts/sync-docs.mjs",
```
astaxie, line 6: [P1] Commit the documentation sync generator

Both `prestart` and `prebuild` invoke this path, but the PR contains no `website/scripts/` directory or `sync-docs.mjs`. After a clean `npm ci`, `npm run build` fails immediately with `Cannot find module .../website/scripts/sync-docs.mjs`, so none of the promised canonical-doc generation, locale fallbacks, link rewriting, or Pages deployment can run. Add the generator and its source-map output contract, then exercise the build in PR CI.

.github/workflows/docs-deploy.yml
```diff
@@ -0,0 +1,59 @@
...
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'docs/**'
+      - 'website/**'
+      - '.github/workflows/docs-deploy.yml'
+  workflow_dispatch:
+
+permissions:
```
astaxie, line 20: [P1] Keep Pages credentials out of the dependency build job

Workflow-level permissions are inherited by `build`, so `npm ci` and the full Docusaurus dependency graph execute with both `pages: write` and `id-token: write`. That is broader than the claimed least-privilege design and turns a compromised install/build script into a deployment-token risk. Give the build job only `contents: read`; move `pages: write` and `id-token: write` to the deploy job that calls `actions/deploy-pages`.

website/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "prestart": "npm run sync-docs",
+    "start": "docusaurus start",
+    "prebuild": "npm run sync-docs",
+    "build": "docusaurus build",
+    "serve": "docusaurus serve",
+    "clear": "docusaurus clear",
+    "typecheck": "tsc"
+  },
+  "dependencies": {
```
astaxie, line 15: [P2] Implement the promised documentation search

The PR describes the site as searchable and adds search placeholder behavior to the sync contract, but this dependency list has no Algolia or local-search plugin and `docusaurus.config.ts` configures no search provider or search-bar navbar item. A successful build would therefore ship no documentation search UI. Add and configure the intended search implementation, including locale behavior.

astaxie (changes requested):
Reviewed the complete current diff at head a7d752599 against current main and the PR's stated website behavior. The core generation pipeline is absent, the privileged deployment workflow grants deployment credentials to the dependency build, and the promised search capability is not configured. `npm ci` and `npm run typecheck` pass, but `npm run build` deterministically fails with MODULE_NOT_FOUND before Docusaurus starts. All 207 repository tests, root gates, and git diff --check pass. GitHub reports no checks for this head.

jackiesre721:
Thank you for the review. All three findings are addressed in e063ab9c:

**[P1] Commit the documentation sync generator**
The script existed locally but was silently excluded: the repository-root `.gitignore` rule `scripts/` was unanchored, so it also matched `website/scripts/` and the file never made it into the commit. The rule is now anchored to `/scripts/` (its original intent), and `website/scripts/sync-docs.mjs` is tracked. Verified with a clean `npm ci && npm run build` — the build now completes for all three locales instead of failing with `MODULE_NOT_FOUND`.

**[P1] Keep Pages credentials out of the dependency build job**
Workflow-level permissions are reduced to `contents: read`. `pages: write` and `id-token: write` now live only on the `deploy` job that calls `actions/deploy-pages`, so `npm ci` and the Docusaurus build run unprivileged.

**[P2] Implement the promised documentation search**
Added `@easyops-cn/docusaurus-search-local` and configured it in `docusaurus.config.ts` with `language: ['en', 'zh', 'ja']`, matching the site locales so each locale build gets correct tokenization. Verified the built output contains `search-index.json` and the navbar search bar for the en, zh-Hans, and ja builds.

`npm run typecheck`, the repository gates, and `git diff --check` all pass on the new head. Note that this repository reports no CI checks for PR heads (the previous head had none either), so validation was done locally.

website/docusaurus.config.ts
```diff
@@ -0,0 +1,187 @@
...
+  // plugin provides the offline search bar and page for every locale build;
+  // its language list matches the i18n locales above.
+  themes: [
+    '@docusaurus/theme-mermaid',
+    [
+      '@easyops-cn/docusaurus-search-local',
+      {
+        hashed: true,
+        docsRouteBasePath: '/',
+        language: ['en', 'zh', 'ja'],
```
astaxie, line 85: [P2] Select the search language separately for each locale

Including `zh` in every locale's language list makes this plugin override its index tokenizer with the Chinese tokenizer and use `/\w+|\p{Unified_Ideograph}+/gu` for queries. This discards Japanese kana: the installed query tokenizer returns `[]` for `モデル`, `ルーティング`, and `ユーザー`, and the search worker immediately returns no results for empty tokens. The generated Japanese index contains 107 documents mentioning `モデル`, but no indexed kana terms. Configure the tokenizer per locale so the Japanese build uses its Japanese tokenizer, and verify a kana query against the resulting index.

website/docusaurus.config.ts
```diff
@@ -0,0 +1,187 @@
...
+  if (!source) {
+    return undefined;
+  }
+  const localeName = localeMatch ? localeMatch[1] : locale;
+  const sourceDir =
+    localeName && localeName !== 'en'
+      ? (localeSourceDirs[localeName] ?? localeName)
+      : undefined;
+  const docsPath = sourceDir ? `${sourceDir}/${source}` : source;
+  return `${repoUrl}/edit/${repoBranch}/docs/${docsPath}`;
```
astaxie, line 45: [P2] Resolve fallback edit links to the actual English source

The Chinese and Japanese PostgreSQL setup pages intentionally fall back to `docs/postgresql-setup.md`, because neither translated source file exists. This code nevertheless prepends the selected locale: the built pages contain edit links to `docs/zh-CN/postgresql-setup.md` and `docs/ja/postgresql-setup.md`. Thus "Edit this page" cannot edit the content being displayed, contrary to the documented source-map contract. Record the actual source for each locale page, or check for a translation before adding its directory.

website/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "@docusaurus/core": "^3.9.0",
+    "@docusaurus/preset-classic": "^3.9.0",
+    "@docusaurus/theme-mermaid": "^3.10.2",
+    "@easyops-cn/docusaurus-search-local": "^0.55.3",
+    "prism-react-renderer": "^2.4.1",
+    "react": "^18.3.1",
+    "react-dom": "^18.3.1"
```
astaxie, line 22: [P2] Use a renderer that preserves multibyte documentation

The locked `react-dom@18.3.1` streaming renderer flushes zero-filled padding when a multibyte character does not fit its 2,048-byte buffer. Docusaurus uses this renderer, and a fresh production build on Node 22.23.1 contains 276 NUL bytes across 46 HTML files. For example, the Japanese user guide emits the heading ID `codex-サブスクリプシ\0ョン画像生成`, while its TOC and landing-page links target `codex-サブスクリプション画像生成`; browsers replace the NUL in the attribute, so these fragments no longer match. This also reproduces in an isolated React 18 streaming example and disappears with React 19. Upgrade React and React DOM together to a compatible fixed version (the installed Docusaurus supports React 19), then verify the built localized HTML and fragment targets.

website/docusaurus.config.ts
```diff
@@ -0,0 +1,187 @@
...
+      style: 'light',
+      links: [
+        {
+          title: 'Documentation',
+          items: [
+            { label: 'Architecture', to: '/concepts/architecture/' },
+            { label: 'User Guide', to: '/guides/user-guide/' },
+            { label: 'Administrator Guide', to: '/guides/administrator-guide/' },
+            { label: 'Deployment', to: '/operations/deployment/' },
+          ],
```
astaxie, line 157: [P3] Generate translations for the shared site footer

Both localized production sites still render the footer as "Documentation / Architecture / User Guide / Administrator Guide / Deployment" (and the remaining groups in English). The sync script writes sidebar translations only, so there are no Docusaurus theme footer translations to replace these labels. Generate the Chinese and Japanese footer translations alongside the sidebar catalog so the shared navigation follows the selected locale, consistent with the repository's multilingual UI/documentation guidelines.

astaxie (changes requested):
Re-reviewed the complete diff at e063ab9cba8f9ee8a3f324b31865ef26c53cb5c5 against main (35dc26f1b5beff1f18ed972440233c3bd8b492a5). The three previous findings are addressed: the generator is tracked, Pages deployment permissions are scoped to the deploy job, and local search is installed.

## Standards

One P3 finding: the shared footer is not translated on the Chinese and Japanese sites.

## Spec and behavior

Three P2 findings: the global search tokenizer discards Japanese kana; fallback pages link to nonexistent translated source files; and the selected React renderer inserts NUL bytes into localized production HTML, breaking heading fragments.

## Validation

Clean dependency installation, typecheck, and all three locale builds pass. Repeated typecheck/build with Node 22.23.1 and rebuilt after clearing Docusaurus caches; the HTML corruption remains reproducible. Inspected generated HTML, source mappings, and search indexes, and exercised the installed query tokenizer. All 175 repository tests, the full base-to-head translation gate, the UI/environment/source-size gates, and git diff --check pass. All nine GitHub checks on this exact head have passed; none builds the new website. No source changes were made.

jackiesre721:
Thank you for the re-review. All four findings are addressed in 6ce08d19, with a browser-level verification pass this time.

**[P2] NUL bytes in localized production HTML**

Confirmed and root-caused: Docusaurus 3.10 renders pages with `renderToPipeableStream` from react-dom 18 (`@docusaurus/core/lib/client/renderToHtml.js`), which emits U+0000 markers into text split across stream chunks (react/react#31134, closed as not planned). I reproduced it deterministically: 46 built HTML files (24 ja, 20 zh-Hans, 2 en) contained NUL bytes, landing inside heading `id` attributes and hash-link `title`s while the TOC `href`s stayed clean, so anchors no longer matched (`id="指定\u0000\u0000モデル情報"` vs `href="#指定モデル情報"`). Since upstream will not fix it, `npm run build` now ends with `scripts/sanitize-build.mjs`, which strips U+0000 from the emitted HTML and fails the build if a NUL character remains anywhere in the text output. After stripping, every checked heading id equals its TOC href, and clicking the previously broken TOC entry in a real browser scrolls to the heading and updates the hash.

**[P2] Search tokenizer discards Japanese kana**

Confirmed: with `language: ['en', 'zh', 'ja']` the zh setting overrides the single global tokenizer with jieba, whose regex only matches ASCII and CJK ideographs, on both the index side and the client query side. The two are not separable inside the plugin, so a patch-package patch extends both tokenizers with kana runs: index-side runs containing kana are segmented with the lunr Japanese tokenizer (tinyseg), and the client query regex keeps kana. Chinese (jieba) and English tokenization are unchanged. Verified in the browser: the ja index now contains リバースプロキシ/デプロイ/データベース and searching「リバース」and「データベース」returns results;「反向代理」on zh-Hans and "gateway" on en still return results.

**[P2] Fallback pages link to nonexistent translated sources**

Confirmed, scope one page per locale (PostgreSQL Setup). The sync script now records per-locale translation state in `docs-source-map.json`, and `docsEditUrl` points fallback pages at the English source instead.

**[P3] Footer not translated**

Root cause note: the site config is loaded once per build (`initialSiteConfig`), so config-level `translate()` cannot resolve per locale. The fix swizzles `src/theme/Footer` to localize column titles, link labels, and the copyright through `translate()`, with values written by the sync script to `i18n/<locale>/code.json`. Verified rendered footer: zh-Hans shows 架构/用户指南/…© 2026 TokenHub 贡献者…, ja shows アーキテクチャ/ユーザーガイド/…, en stays English.

**Process: none of the CI checks built the website**

Added `docs-website-build.yml`, which runs `npm ci && npm run build` on pull requests touching `docs/` or `website/`. Because the build now fails on remaining NUL characters, this check would have caught the first two findings.

**Validation**

Clean `npm ci && npm run typecheck && npm run build` passes for all three locales (30 pages per locale; the sanitizer strips the emitted markers on every build). Repository gates pass: `node --test tools/*.test.mjs` (175 tests), the translation, UI-translation, environment-contract, and source-size gates, `tsc`, and `git diff --check`.

One observation, out of scope for this PR and present on the base docs as well: the sequence diagram in `docs/architecture.md` throws a mermaid parse error in the browser ("Expecting 'participant' … got 'ACTOR'"), so that one diagram does not render; the flowchart diagrams on the same page render fine. Happy to fix it in a follow-up.

website/docusaurus.config.ts
astaxie: Rechecked at `6ce08d19e7d6b47252c4cfa0594786039e8d6ef0`. Single kana words now work, but the new patch still tokenizes Japanese queries differently from indexed text. The actual heading `よくあるエラー` in `docs/ja/team-leader-guide.md:88` is indexed as `よく`, `ある`, `エラー`; the patched client regex keeps the whole query as `よくあるエラー`. Against the generated Japanese production index, the configured exact/prefix/fuzzy variants return zero matches for that heading, while `エラー` returns 27. Apply the same Japanese segmentation to queries and indexing, or select the language per locale, and cover a multi-token kana phrase. The mismatch is in `website/patches/@easyops-cn+docusaurus-search-local+0.55.3.patch`: client lines 10–13 versus index lines 37–48.

The other reviewed fixes pass: all three locale production builds and typecheck succeed, sanitized HTML contains no NUL characters, translated footers render correctly, and fallback edit links target the English source. This follows up the attempted fix in the existing discussion rather than opening a duplicate thread.

PR #341, opened 2026-09-15
fix: restore migration compatibility and local provider labels

astaxie:
## Summary

Databases that applied the original version 4 metering migration fail startup with `checksum_mismatch` because a later release replaced its SQL checksum with an override. Restore startup compatibility without rewriting migration history. Also present the built-in local provider as “Local Provider” in the console instead of exposing its implementation name.

## Related Issue

N/A

## Changes

- Restore version 4's original SQL-derived checksum and regenerate the migration manifest.
- Recognize the exact intermediate checksum only for the unchanged, clean version 4 metering migration; continue rejecting other checksum mismatches, changed SQL, and dirty ledgers.
- Cover both historical ledger variants, following migrations, restart behavior, and SQLite/PostgreSQL upgrades.
- Localize local-provider labels and the empty-endpoint display in English, Simplified Chinese, and Japanese. Recompute provider options after synchronizing the active locale so switching languages immediately updates the editor.
- Cover language switching from Chinese to English, Japanese, and back without a page reload or provider-data refresh; verify both the visible selected label and unchanged provider type.
- Document recovery commands consistently in all three database-evolution guides.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor or maintenance
- [x] Documentation

## Verification

- Backend: formatting, `go test ./...`, `go vet ./...`, golangci-lint v2.12.2, and migration manifest freshness.
- PostgreSQL: the CI migration/multi-instance selection and dbschema integration suite against an isolated PostgreSQL 16 container.
- Frontend: Node 22.23.1, dependency installation, lint, typecheck, domain/component tests, production build, targeted UI acceptance/capture, and browser smoke tests.
- Repository: gate tests, documentation/UI translations, environment contract, source sizes, and `git diff --check`.
- Review regression: the browser scenario failed before the fix with expected `Local Provider` and received `本地服务`; the fixed scenario and captures pass in all three languages. All 9 GitHub Actions checks passed for `cc256769399e6ebc6a2e441dae22608bc3f5ebac` ([CI run](https://github.com/astaxie/TokenHub/actions/runs/34955308669)).
- Real-provider SDK smoke tests are not applicable: provider credentials and public API behavior are unchanged.

Existing ledger rows are preserved. The compatibility exception is pinned to the original SQL checksum, version, name, phase, unrestricted dialect, and clean state; it provides no generic checksum bypass. Migration 5 continues to supply the audit correlation expansion. No API, environment variable, authentication, or provider execution changes are introduced. Technical provider identifiers remain unchanged. No local seed data, credentials, databases, or unrelated research files are included.

## Checklist


frontend/features/admin/shared/ui.tsx
```diff
@@ -546,6 +546,7 @@ export function providerTypeOptionsFromData(data: Pick<AppData, "plugins" | "pro
 }
 
 function providerTypeOptionLabel(providerType: string, labels: Map<string, string>) {
+  if (providerType === "mock") return providerTypeLabel(providerType);
```
astaxie, line 549: [P3] Refresh cached provider labels when the language changes

This now resolves the `mock` label using the active locale, but `AdminConsole` caches `providerTypeOptionsFromData(data)` with only `[data]` (`shell/admin-console.tsx:106`). Switching the UI language does not invalidate that cache, and the editor renders `option.label` verbatim. I reproduced this on this head with the UI fixture: load `/providers` in Chinese, switch to English, edit the local provider, and open **Advanced**. The list and surrounding controls switch to English, but the provider type option remains `本地服务` instead of `Local Provider`; the added language-switch assertion fails with exactly those received/expected values. Recompute the options when the language changes after updating the active runtime locale, or resolve this built-in label at render time. Please cover switching languages without reloading in the UI test.

Issue #343, opened 2026-09-16
feat: supported SQLite-to-PostgreSQL upgrade path with a read-only preflight command


zhyese:
### Affected area

Persistence and migrations

### Problem or use case

A deployment graduates from single-node SQLite to multi-replica PostgreSQL as adoption grows. The only documented path (`docs/postgresql-setup.md`) is a manual `sqlite3 .dump`, hand-editing the SQL into PostgreSQL syntax, and a `psql` import. For this codebase that path is unsafe:

- The dialects have diverged deliberately: per-dialect baseline schemas, different `request_logs.commit_sequence` triggers, and `real` vs `decimal` billing columns.
- Sensitive columns persist `enc:v1:` AES-256-GCM ciphertext. A wrong `TOKENHUB_SECRET_KEY` on the target silently decrypts to empty strings (the store layer returns `""` on failure), losing provider credentials without any visible error.
- `analytics_sequences` watermarks and the migration ledger are runtime state a blind SQL copy cannot rebuild.

`docs/postgresql-setup.md` already notes: "A data migration tool is planned for a future release."

### Proposed solution

Add `tokenhub db` upgrade commands backed by a design doc:

1. `tokenhub db upgrade-plan` — read-only preflight (ready to review as a PR):
   - Source inventory: tables, row counts, and buckets (protected / history / system / other).
   - Secret key resolution: `--secret-key`, then `TOKENHUB_SECRET_KEY`, then the `.secret-key` sidecar file beside the source database (read-only; provisioning stays with startup).
   - Ciphertext canary: every registered protected column is scanned and each distinct `enc:v1:` value must decrypt with the resolved key. Serving-configuration columns (provider credentials, connector credentials, bootstrap password) produce blockers; historical artifacts (image/response job payloads, OAuth session records) produce warnings. This catches a wrong key before any data moves, which the store layer cannot do on its own.
   - Target classification: empty (a not-yet-existing SQLite target file is classified without being created), tokenhub (occupied targets block), or unrecognized (blocks).
   - Text and JSON output; findings are reported rather than failing the command.
2. `tokenhub db upgrade-run` (follow-up): consistent source snapshot via the existing online backup API, target schema adopted by the application's own migration flow, batched copy through the GORM models, optional decrypt-and-re-encrypt under a new key, `analytics_sequences` watermark rebuild.
3. `tokenhub db upgrade-verify` (follow-up): per-table row counts plus sampled row hashes, JSON report.

### Alternatives considered

[...]

Issue #302, opened 2026-08-27
Backend fails to compile on Windows: syscall.Flock is undefined


piaosky:
**Affected area**: Persistence and migrations

**TokenHub version**: v0.7.0 (commit 6e6bc71)

**Deployment method**: Not deployment-related (local development)

**Current behavior**

On Windows, the backend does not compile. `go build ./...` fails with:

```
# tokenhub/backend/internal/dbschema
internal\dbschema\lock.go:83:18: undefined: syscall.Flock
internal\dbschema\lock.go:83:50: undefined: syscall.LOCK_EX
internal\dbschema\lock.go:83:66: undefined: syscall.LOCK_NB
```

Because many packages import `internal/dbschema`, this breaks `go build` / `go test` for the entire backend (`cmd/tokenhub`, `internal/server`, ...). CGO is enabled with a working MinGW-w64 toolchain; the failure is purely the Unix-only `flock(2)` API used by `acquireSQLiteMigrationLock`.

**Expected behavior**

`go build ./...` succeeds on Windows, so contributors can develop and run backend tests locally.

**Steps to reproduce**

1. Clone TokenHub on Windows with Go 1.26 and a C toolchain (`CGO_ENABLED=1`).
2. `cd backend && go build ./...`
3. Observe the compile errors above.

**Relevant logs or configuration**

Included above.

**Additional context**

The standard fix is a platform split: keep `flock(2)` behind `//go:build !windows` and use `LockFileEx` (from `golang.org/x/sys/windows`, already an indirect dependency) on Windows, preserving the exclusive non-blocking semantics and the bounded-wait retry loop. I have this implemented and validated on both sides (Windows 11: build, vet, and the dbschema lock-contention tests pass; Linux container with Go 1.26: build passes) and will open a PR that fixes this issue.

Note for completeness: two further Windows blockers exist in the browser e2e stack (`sqlite://C:\...` drive-letter URLs rejected by `url.Parse`, and Node >= 18.20 refusing to spawn `.cmd` shims without a shell). Happy to send those separately if Windows support is welcome.

**Submission checklist**

- [x] I searched open and closed issues for duplicates.
- [x] The issue title and all explanatory text are written in English.
- [x] I removed or redacted all credentials and other secrets.

piaosky:
The fix is ready and verified. Since pull requests do not appear to be enabled on this repository (the pulls API returns 404 for cross-fork creation), I am sharing the branch for direct consumption instead:

[...]

PR #345, opened 2026-09-17
feat(db): add SQLite-to-PostgreSQL upgrade preflight command

zhyese:
## Summary

Deployments that outgrow single-node SQLite currently have only a manual `sqlite3 .dump` → hand-edit → `psql` import path (documented in `docs/postgresql-setup.md`), which is unsafe for this codebase: per-dialect baseline schemas, `enc:v1:` AES-256-GCM ciphertext that silently decrypts to empty strings under a wrong `TOKENHUB_SECRET_KEY`, and `analytics_sequences` watermarks that a blind SQL copy cannot rebuild. This PR adds the first, read-only phase of the planned upgrade tooling: a `tokenhub db upgrade-plan` preflight that inspects a source database and a candidate target and reports findings without moving or modifying any data.

## Related Issue

Refs #343 — delivers phase 1 (`upgrade-plan`). The `upgrade-run` and `upgrade-verify` commands remain follow-ups as laid out in the issue.

## Changes

- New `backend/internal/dbupgrade` package: table inventory and bucket classification (protected / history / system / other), secret-key resolution (`--secret-key` → `TOKENHUB_SECRET_KEY` → `.secret-key` sidecar), ciphertext canary that decrypts every distinct `enc:v1:` value in protected columns (blockers for serving configuration, warnings for historical artifacts), target classification (empty / tokenhub-occupied / unrecognized), and text + JSON rendering that reports findings instead of failing the command.
- New `tokenhub db upgrade-plan` CLI wiring in `backend/internal/dbcli` (requires `--target <database-url>`) with tests.
- Read-only `.secret-key` sidecar reader (`ReadSQLiteSecretKeySidecar`) and ciphertext verification helper (`VerifySecretCiphertext`) in `backend/internal/server`; provisioning and startup behavior are unchanged.
- New design document `docs/development/database-upgrade-design.md` covering all three planned commands; `docs/postgresql-setup.md` now points to the preflight instead of promising a future tool.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor or maintenance
- [ ] Documentation

## Verification

- `gofmt` clean on all changed Go files; `go vet ./...` passes in a Linux Go container (the backend does not compile natively on Windows, #302).
- `go test ./...` passes for every package except pre-existing plugin-fixture failures local to this Windows checkout: `internal/plugin` and `internal/server` fixture tests exec `testdata/*.sh` scripts whose shebangs carry a CRLF `\r` from this machine's `autocrlf=true` checkout (`od -c` shows `#!/bin/sh\r\n`), so Linux `fork/exec` reports "no such file or directory". All 19 failing tests are that single fixture-exec pattern; the fixtures are git-unmodified and unrelated to this diff. CI's LF checkout is the authoritative run for them.
- New behavior is covered by `backend/internal/dbupgrade/plan_test.go` and `backend/internal/dbcli/cli_test.go`.
- Repo gates: `git diff --check` on the PR range, doc-translation, UI-translation, env-contract, and source-line checks all pass (the local ci-aggregate node test shows the same known CRLF-checkout artifact and is covered by CI).
- Frontend checks not run: no frontend files changed.

Read-only tooling: it opens the source database and inspects files without writing. Secret-key material is resolved only from existing channels (the `--secret-key` flag, `TOKENHUB_SECRET_KEY`, or the `.secret-key` sidecar) and is never printed; the preflight reports only whether each protected value decrypts. No changes to runtime behavior, the public API surface, environment variables, or deployment.

## Checklist

- [x] Environment variable changes are synchronized across examples, Compose, `start.sh`, and deployment documentation where applicable. (N/A — no environment variable changes.)
- [x] Shared user-facing behavior is documented consistently in English, Simplified Chinese, and Japanese where applicable. (Developer-facing documentation only; no user-facing UI copy changed.)
- [x] `data/model-catalog.yaml` remains tracked and catalog changes were reviewed where applicable. (N/A — no catalog changes.)


backend/internal/dbupgrade/plan.go
```diff
@@ -0,0 +1,193 @@
...
+func BuildPlan(ctx context.Context, opts Options) (*Plan, error) {
+	plan := &Plan{Blockers: []string{}, Warnings: []string{}}
+
+	secretKey, keySource, keyNotice := resolveSecretKey(opts.SourceURL, opts.SecretKey)
+	plan.Secret.KeySource = string(keySource)
+	if keyNotice != "" {
+		plan.Warnings = append(plan.Warnings, keyNotice)
+	}
+
+	driver, db, err := server.OpenRawDatabase(opts.SourceURL)
```
astaxie, line 26: [P2] Open the source without creating a missing database

A typo in `--from sqlite:///.../typo-source.db` currently creates a zero-byte source file and returns a successful plan with an empty inventory, `secret.verified: true`, and no blockers. I reproduced this with the built CLI; the missing target correctly stayed absent. This violates the read-only preflight contract and reports readiness without inspecting the intended deployment. Open the source with existing-file/read-only semantics and reject an absent or unrecognized source before reporting READY.

backend/internal/dbupgrade/inspect.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	version, err := ledgerVersion(ctx, db, driver)
+	if err != nil {
+		return report, err
+	}
+	report.State = TargetStateTokenhub
+	report.LedgerVersion = version
+	// Probing a handful of core tables keeps classification cheap on large
+	// targets; COUNT(*) would seq-scan every table.
+	for _, table := range []string{"providers", "admin_users", "api_keys", "usage_records", "request_logs"} {
```
astaxie, line 108: [P2] Check all application tables before declaring the target empty

The five-table probe misses existing data in `projects`, model/routing configuration, billing connectors, and other application tables. With a fully adopted target, inserting only `projects(id, name)` makes `BuildPlan` report `has_data: false`, no blockers, and the warning that the schema has "no data". This contradicts the preflight contract that occupied targets block and makes its migration-readiness report unsafe to rely on. Probe each non-system application table with `SELECT 1 ... LIMIT 1`, excluding only explicitly permitted bootstrap/system rows.

backend/internal/dbupgrade/canary.go
```diff
@@ -0,0 +1,90 @@
...
+	"fmt"
+	"regexp"
+
+	"tokenhub/backend/internal/server"
+)
+
+// ciphertextTokenPattern matches enc:v1 payloads inside raw column text. It
+// finds both plain ciphertext columns and ciphertext embedded in JSON
+// documents such as providers.sensitive_headers.
+var ciphertextTokenPattern = regexp.MustCompile(regexp.QuoteMeta(server.SecretCiphertextPrefix) + `[A-Za-z0-9_-]+`)
```
astaxie, line 15: [P2] Verify complete stored ciphertext values instead of regex prefixes

The regex can turn an undecryptable stored value into a valid token before verification. I appended `!` (also reproduced with `=`) to a real encrypted `providers.api_key`: `VerifySecretCiphertext` rejects the complete value, but `BuildPlan` verifies only the original base64 prefix and reports `verified: true`, zero failed ciphertexts, and no blockers. A bare `enc:v1:` is also skipped entirely. The canary therefore misses the credential corruption it is intended to detect. Validate complete scalar values, and parse JSON columns to validate each complete protected string; malformed prefixed values must fail.

astaxie (commented):
Reviewed head `04513e2984729d277badf03f91328465ab83571e` against the PR contract and issue #343. Three reproducible findings are inline.

Standards: the source open violates the documented read-only contract.

Spec: target occupancy and complete-ciphertext verification have false-negative cases. The existing dbupgrade/dbcli tests pass; temporary regression cases reproduce both false negatives, and a CLI reproduction confirms the missing-source side effect.

PR #346, opened 2026-09-19
feat(typesafe): add native system one provider integration

legendtkl:
## Summary

TokenHub cannot currently route TypeSafe Jev decision requests. Add a dedicated TypeSafe provider and native `POST /v1/systemone` endpoint so applications can submit shared state with `choice`, `noul`, and `score` questions through the gateway's existing governance and billing lifecycle.

## Related Issue

N/A

## Changes

- Add the native adapter, request and response validation, model discovery and probes with full custom-header and sensitive-name validation before resource merging, read-only health probes, capability filtering, and compatible-route failover.
- Integrate project authentication, model permissions, quotas, privacy and guardrail checks, protocol-scoped hooks, actual token metering, request logs, and traces. Preserve independently valid native usage even when a response envelope is malformed, across both the adapter and provider-call hooks.
- Add Jev catalog entries and the built-in provider package, expose the `decision` modality in model configuration, and extend OpenAPI plus English, Simplified Chinese, and Japanese documentation.
- Add backend contract and governance regressions, a UI configuration scenario, and an official `@typesafe-ai/sdk@0.6.0` smoke test. Pin the workspace root during isolated UI runs so unrelated parent lockfiles cannot disrupt the fixture server.
- Capture native TypeSafe request IDs and return the winning upstream ID through the SDK's `withResponse().requestId`, with browser header exposure. Preserve SDK 0.6.0 optional/null request fields. Compare score legends semantically against the winning route's effective criteria, retaining masking and exact JSON numeric precision; localize the decision modality.
- Make the plugin catalog type-mapping test read its local fixture directly, removing an unrelated external refresh and snapshot-retention dependency exposed during validation.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor or maintenance
- [x] Documentation

## Verification

- Backend: `go test ./...`, `go vet ./...`, and `golangci-lint run ./...` with v2.12.2 passed. Formatting and migration manifest checks passed.
- Frontend on Node.js 22.23.1: dependency installation, lint, typecheck, 181 domain tests, 297 component tests, and production build passed.
- Browser coverage: 38 UI scenarios and 10 E2E smoke tests passed. The TypeSafe UI scenario and capture, plus all 10 E2E smoke tests, also passed after the latest compatibility fixes.
- Official SDK smoke passed against the actual local Go gateway and a synthetic HTTP upstream, covering provider import, model publication, project authentication, all three primitives, native usage, and omitted/null SDK fields. The SDK `withResponse()` check also passed after the response-header fix. This does not establish live model quality or full SDK compatibility.
- Repository gates: 175 tool tests, documentation and UI translations, environment contract, source sizes, and `git diff --check` passed. The documentation translation check also passed with explicit base/head commits.
- Independent read-only reviews covered gateway governance and provider/API contracts. Final follow-up changes were re-reviewed with no remaining findings.
- Response regressions cover inclusive probability-sum boundaries, permitted legend types, nesting limits, semantic legend equality (including exact large numbers), privacy and route transforms, winning-route failover, post-hook tampering, score/choice consistency with probabilities, and sanitized HTTP 502 errors that retain per-attempt usage. Focused tests (19.221s), SDK smoke, full Go tests (server package 243.828s), vet, and lint passed for the final malformed-usage and sensitive-header fixes. A new independent reviewer audited the full PR and re-reviewed both resulting fixes with no remaining findings.
- All nine GitHub Actions checks passed for latest pushed head `2ee57398db08c35ab3cee64101fa338e17b53e47`: Backend, Backend PostgreSQL, Backend database N-1 contract, Plugin DevKit, Frontend, Frontend browser smoke, Deployment, Repository gates, and CI. [CI run](https://github.com/astaxie/TokenHub/actions/runs/35449355456).
- Copilot's review quota was exhausted during follow-up, so its latest review did not cover this head. A new independent read-only reviewer completed a full-diff audit and confirmed both subsequent fixes with no remaining findings. All actionable review threads are resolved; the quota error is not counted as a successful review.
- PostgreSQL, N-1 database compatibility, Plugin DevKit, and deployment jobs run in GitHub Actions rather than separately on the local machine. No database or deployment changes are included.

### Live configuration and inference screenshots

Captured from the running local TokenHub console on 2026-09-20 (Asia/Shanghai), using the live TypeSafe upstream and a synthetic customer-service request. The configured `jev-1.13.0` route returned HTTP 200 in 470 ms, with 383 input tokens and 61 output tokens. Request logs, both request IDs, and usage accounting were verified. This single request verifies connectivity and integration, not model quality.

| Provider configuration | Enabled Jev route | Successful request details |
| --- | --- | --- |
| TypeSafe provider configuration with the stored API key not displayed | Enabled jev-1.13.0 decision route to TypeSafe Jev | Live Jev request with HTTP 200, 470 ms latency, and 444 total tokens |

The provider credential is already saved; the empty API Key field means “keep the existing key.” The request log shows only the gateway key's masked identifier. No full credentials are visible in these screenshots. Applications call `POST /v1/systemone` with a TokenHub project key restricted to the published Jev model.

- The endpoint is additive. Existing Chat, Responses, Embeddings, streaming, and multi-protocol `/v1/models` contracts are preserved. This phase supports the SDK's `systemOne()` method, without a dedicated playground or internal semantic routing.
- Reuse encrypted provider credentials, SSRF-protected upstream transport, bounded JSON parsing, and sanitized upstream errors. Guardrails inspect state, instructions, question IDs, and criteria; required redaction of structural keys fails closed. New tests verify protocol-scoped authentication cannot be bypassed.
- Consistency checks allow one percentage point for probability sums and Choice winner differences, and 1% of the score range for Score expectation differences, plus floating-point epsilon. Probabilities are not renormalized. These are explicit gateway tolerances; independently rounded upstream distributions can exceed them because TypeSafe publishes no wire precision guarantee.
- Bill upstream-reported input/output tokens using configured prices. Low-confidence answers remain valid results. Catalog pricing is dated and should be rechecked before rollout.
- No database migration or deployment environment variable is required. Roll out through an explicit Jev model and dedicated project key; disable the model or route to stop traffic.
- All committed fixtures are synthetic. The SDK lockfile uses the public npm registry and preserves existing dependency versions; four pre-existing low-severity `ai` / `@ai-sdk` advisories remain outside this change.

## Checklist


astaxie (commented):
Reviewed head `2ee57398db08c35ab3cee64101fa338e17b53e47`. The existing TypeSafe/System One focused test suite passes, and all nine current CI checks are green. Two additional local reproductions with synthetic upstreams exposed the boundary cases below.

backend/internal/server/gateway_systemone_hooks.go
```diff
@@ -0,0 +1,101 @@
...
+				array[index] = next
+				encoded, _ := json.Marshal(array)
+				set(encoded)
+			}, keyRedacted)
+		}
+		return
+	}
+	var object map[string]json.RawMessage
+	if json.Unmarshal(raw, &object) != nil || object == nil {
+		return
```
astaxie, line 98: [P1] Inspect numeric JSON leaves in outbound sensitive-data checks

Structured inputs can contain sensitive values as JSON numbers, but this branch silently drops every numeric leaf from the guardrail fragments. With an all-project sensitive-data policy configured to block `phone`, I reproduced `state: {"contact":"13812345678"}` returning HTTP 403 with no upstream call, while `state: {"contact":13812345678}` returns HTTP 200 and forwards the phone upstream. The same walker handles nested instructions and criteria, so those locations have the same bypass. Include numeric leaves in policy inspection while preserving unmatched numeric precision, and fail closed if a required mask cannot preserve the field's JSON type. Please add regressions for structured numeric PII; the current all-inputs test only exercises string values and keys.

backend/internal/server/provider_typesafe.go
```diff
@@ -0,0 +1,161 @@
...
+		return SystemOneResponse{}, Usage{}, err
+	}
+	response, err := a.request(ctx, provider, http.MethodPost, "/systemone", body)
+	if err != nil {
+		return SystemOneResponse{}, Usage{}, err
+	}
+	defer response.Body.Close()
+	data, err := io.ReadAll(io.LimitReader(response.Body, (8<<20)+1))
+	if err != nil || len(data) > 8<<20 {
+		return SystemOneResponse{}, Usage{MeteringInvalid: true}, invalidSystemOneResponse()
```
astaxie, line 43: [P2] Preserve cancellation errors while reading the response body

If the caller cancels after the upstream has flushed HTTP 200 headers but before its JSON body completes, `io.ReadAll` returns `context.Canceled`. This branch replaces it with `invalidSystemOneResponse()` and `ProviderErrorTransientSame`, so the router no longer recognizes that the client is gone and considers the abandoned call eligible for failover. A synchronized local reproduction produces `errors.Is(err, context.Canceled) == false` and `shouldFailoverRoutedError(err, false) == true` for TypeSafe; the existing OpenAI JSON path preserves cancellation and declines failover for the same exchange. Preserve or classify cancellation before translating malformed/oversized bodies to the sanitized provider error, and add a cancellation-during-body-read test (the current cancellation test only cancels before sending).

backend/internal/server/gateway_systemone_hooks.go
legendtkl: Fixed in 2a8bd8d1bbd401df78cec5cd60f99b6c20e6a2af.

Numeric JSON leaves now participate in the same outbound policy checks as strings, including nested state, instructions, and all three question criteria. Unmatched values keep their exact original JSON representation. If a policy requires masking a numeric value, the gateway returns `403 guardrail_blocked` before contacting the upstream, since a text mask cannot preserve its JSON type.

Regressions cover block and mask policies across all five input locations, plain and scientific-notation phone numbers, exact large-integer/fraction/exponent preservation while masking a sibling string, and bounded scientific-notation parsing (including million-digit exponents). The exponent is bounded before conversion; an independent follow-up review found no remaining issues.

backend/internal/server/provider_typesafe.go
legendtkl: Fixed in 2a8bd8d1bbd401df78cec5cd60f99b6c20e6a2af.

The TypeSafe inference and model-discovery adapters now preserve cancellation/deadline errors from response-body reads before applying sanitized invalid-response handling. The router therefore recognizes `context.Canceled` and does not allow failover for an abandoned call.

A synchronized HTTP regression flushes status 200, waits until the adapter starts reading the body, then cancels the caller and checks both `errors.Is(err, context.Canceled)` and the no-failover decision. It covers inference and discovery. A separate truncated-body regression confirms unrelated read failures still produce sanitized, retryable provider errors. Independent review found no remaining issues.

backend/internal/server/gateway_systemone_hooks.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+func (s *Server) runGatewaySystemOnePrivacyPreHooks(ctx context.Context, call CallContext, headers http.Header, req *SystemOneRequest) error {
+	return s.runGatewayPrivacyPreHooks(ctx, call, headers, *req, func(data json.RawMessage) error {
+		return applySystemOneRequestPatch(req, data)
+	})
+}
+
+func (s *Server) runGatewaySystemOneGuardrailPreHooks(ctx context.Context, call CallContext, req *SystemOneRequest) error {
+	return s.runGatewayGuardrailPreHooks(ctx, call, *req, systemOneGuardrailTargets(req), func(data json.RawMessage) error {
```
astaxie, line 26: [P1] Bound the guardrail target traversal

`systemOneGuardrailTargets` has no bound on the number of targets or leaves. The request validator only caps nesting depth; a legal sub-8 MiB JSON array can contain roughly two million scalar leaves, and this traversal allocates hundreds of MiB. The callbacks also re-marshal the whole parent array/object for each replacement. This cost is incurred even when no target will be replaced, so an authenticated caller can trigger a memory/CPU spike with a normal request. Please enforce a shared node/leaf/target budget during validation/traversal and avoid rebuilding whole containers per leaf (or skip target construction when no relevant hook is configured).

backend/internal/server/systemone_json.go
```diff
@@ -0,0 +1,68 @@
...
+
+func normalizeSystemOneNumber(value json.Number) json.Number {
+	mantissa := string(value)
+	sign := ""
+	if strings.HasPrefix(mantissa, "-") {
+		sign, mantissa = "-", mantissa[1:]
+	}
+	exponent := new(big.Int)
+	if index := strings.IndexAny(mantissa, "eE"); index >= 0 {
+		exponent.SetString(mantissa[index+1:], 10)
```
astaxie, line 53: [P1] Bound the exponent token before `SetString`

`normalizeSystemOneNumber` passes the complete caller-controlled exponent to `big.Int.SetString`. Although the canonical form keeps exponents symbolic, parsing a multi-million-digit exponent is still linear in input size and causes substantial CPU/GC work; the same response can be normalized repeatedly by adapter and gateway validation. An 8-million-digit exponent took about 26.8 seconds locally while fitting under the request-size limit. Bound the exponent token length before `SetString` (or compare it lexically against a safe limit) and return a bounded validation result.

astaxie (commented):
Re-reviewed head `1b4fd23c5d1b5eb17e99f3ef426093b0c918ea04`. The focused System One tests pass and the current CI checks are green. Four additional correctness/security findings are attached inline.

backend/internal/server/gateway_systemone_hooks.go
```diff
@@ -0,0 +1,165 @@
...
+			appendSystemOneJSONTargets(targets, child, fmt.Sprintf("%s.%d", id, index), func(next json.RawMessage) {
+				array[index] = next
+				encoded, _ := json.Marshal(array)
+				set(encoded)
+			}, unsafeRedaction)
+		}
+		return
+	}
+	var object map[string]json.RawMessage
+	if json.Unmarshal(raw, &object) != nil || object == nil {
```
astaxie, line 97: [P1] Reject duplicate JSON object members

Decoding into `map[string]json.RawMessage` silently keeps only the last value for a duplicate key. If no replacement occurs, the original raw JSON is forwarded, so a blocked marker in an earlier duplicate can bypass the guardrail while the inspected map contains only the later value. I reproduced a 200 response with the marker reaching the upstream under a blocking policy. Parse objects with a duplicate-key-rejecting decoder (and apply the same rule to response-consistency decoding).

docs/typesafe.md
```diff
@@ -0,0 +1,83 @@
...
+
+The catalog snapshot verified on **2026-09-19** uses **USD 0.042 per million input tokens** and **USD 0 output price**. Output tokens are still counted toward usage and token quotas. Final billing uses upstream-reported tokens and configured prices, never estimated tokens or a charge per question. Check alias pricing and Provider cost rates before rollout. Source: [TypeSafe models](https://docs.typesafe.ai/models).
+
+## Rollout and rollback
+
+Start with a dedicated project key and an explicit Jev route. Validate labeled examples for the intended language and task before application decisions depend on the scores. Observe latency, upstream errors, actual usage, and costs. This integration adds no database migration or deployment environment variable. Disable the published model or its routes to stop traffic. Disabling the catalog plugin only removes it from Provider setup; it does not disable existing routes.
```
astaxie, line 81: [P2] Align plugin-disable behavior with the rollback guarantee

Disabling the catalog plugin currently also removes the executable TypeSafe adapter/capability from the provider registry. Existing TypeSafe routes are then filtered out and `/v1/systemone` returns 501, so traffic is disabled. That contradicts this documented guarantee. Either separate catalog metadata from adapter lifecycle so disabling the catalog leaves existing routes usable, or update the docs/UI to state that disabling the plugin disables those routes.

backend/internal/server/gateway_systemone_hooks.go
legendtkl: Fixed in `b8dd2bf6`, preserved in the latest merged head `191b3a8b`. Request/response decoding now enforces a shared 32,768-node budget (containers, scalars, and object keys) before building full maps or targets, with a separate 32,768-target guardrail limit and an 8 MiB payload cap. Replacements mutate one parsed tree and serialize each changed entry once; an absent guardrail-pre hook skips its target construction.

Regressions cover wide arrays, aggregate fields, target overflow, batched masking of 4,096 strings with exact numeric text, and 64-level criterion compatibility.

Validation: local backend/frontend/gates and SDK checks passed; all nine checks passed in [CI run 35846350865](https://github.com/astaxie/TokenHub/actions/runs/35846350865).

backend/internal/server/systemone_json.go
legendtkl: Fixed in `b8dd2bf6`, preserved in the latest merged head `191b3a8b`. Exponent text is limited to 128 digits, including leading zeros, before any `big.Int.SetString` call. The same bound is enforced during native JSON decoding, so oversized values cannot reach repeated legend normalization. In-range values retain symbolic, exact-decimal comparison.

Boundary tests cover 128/129 digits, positive and negative exponents, zero/leading-zero exponents, and a million-digit exponent.

Validation: local backend/frontend/gates and SDK checks passed; all nine checks passed in [CI run 35846350865](https://github.com/astaxie/TokenHub/actions/runs/35846350865).

backend/internal/server/gateway_systemone_hooks.go
legendtkl: Fixed in `b8dd2bf6`, preserved in the latest merged head `191b3a8b`. A bounded token decoder now rejects duplicate object members before map/struct decoding, including escaped spellings of the same key. Alternate casing of native protocol fields is also rejected. This applies to request bodies, request patches, native responses, and response patches. Raw provider-call responses retain exact numbers until native validation and independent usage extraction; post-response patches are checked before a subsequent stage can normalize away a duplicate.

Regressions cover state/instructions/criteria, question IDs, answers/legends/usage, every applicable request/response hook boundary, the reproduced cross-stage normalization bypass, and preservation of only unambiguous valid usage.

Validation: local backend/frontend/gates and SDK checks passed; all nine checks passed in [CI run 35846350865](https://github.com/astaxie/TokenHub/actions/runs/35846350865).

docs/typesafe.md
legendtkl: Fixed in `b8dd2bf6`, preserved in the latest merged head `191b3a8b`. Corrected the English, Simplified Chinese, and Japanese rollout guidance: disabling `tokenhub.provider.typesafe` removes the executable adapter as well as the catalog entry, making existing TypeSafe routes unavailable. A model with only these routes returns `501 provider_capability_not_supported`; re-enabling restores the adapter for the existing configuration. Selective shutdown should disable individual models/routes.

Added an admin-API regression that enables the plugin, verifies native HTTP 200, disables it and verifies 501 without an upstream call, then re-enables it and verifies HTTP 200 again.

Validation: local backend/frontend/gates and SDK checks passed; all nine checks passed in [CI run 35846350865](https://github.com/astaxie/TokenHub/actions/runs/35846350865).

astaxie (approved):
Reviewed the complete current diff at head `191b3a8b0`. The latest bounded JSON decoder, shared node/target budgets, exponent bound, duplicate-member rejection, and corrected plugin lifecycle documentation address the four prior findings; their regression coverage passes and all threads are resolved. No additional blocking issue was found.

Local validation passed the full backend test suite, vet, golangci-lint v2.12.2, frontend lint/typecheck/tests/build, 10 browser smoke tests, 42 UI fixture scenarios, 175 repository tests, all root gates, and `git diff --check`. All nine GitHub checks are green.

PR #347, opened 2026-09-20
feat(routing): add opt-in Jev semantic model selection

legendtkl:
## Summary

Existing routing strategies cannot use the task's meaning to choose between configured generation models. Add optional Jev semantic selection that promotes an eligible Provider/model pair within the current priority tier while retaining normal execution and failover. Jev uses an independent TypeSafe client and does not require a generation Provider adapter.

## Related Issue

N/A

## Changes

- Require the actual Chat `max_tokens` parameter in explicit candidate declarations; other output-budget fields do not imply an adapter translation.
- Add bounded TypeSafe Choice evaluation, strict response validation, concurrency and timeout limits, and safe decision audit events. Errors, low confidence, and unsupported requests retain the base order.
- Gate evaluation behind a default-off deployment switch, exact project allowlist, and per-model `off`/`shadow`/`enforce` policy. Preserve existing candidate permissions, priority tiers, resource ordering, and affinity.
- Store policy updates atomically with routes. Preserve the managed policy during catalog refreshes and ordinary model edits, with consistent PostgreSQL row locking to prevent stale edits from restoring a disabled policy.
- Add the model-routing editor, English/Chinese/Japanese UI copy and documentation, synchronized environment configuration, and synthetic backend/UI/browser tests.
- Resolve the isolated UI server path from Next.js's tracing manifest so acceptance tests also work with parent-directory lockfiles.

## Type of Change

- [x] Bug fix
- [x] New feature
- [x] Refactor or maintenance
- [x] Documentation
- [x] Deployment or configuration

## Verification

- Output-budget review regression: nine parameter-support cases and eight full HTTP scenarios cover alternative budget declarations, streaming/non-streaming, and enabled/disabled controls. Before the fix, enabled scenarios reproduced HTTP 400; after it, all return 200 and incompatible upstreams receive no calls. Full backend tests, targeted race tests, vet, lint, and a fresh independent review passed after the fix.
- PR CI for `c0c2004e662d401f6e1d5d3a4fd019baebc2ddfd`: [all 9 main CI checks passed](https://github.com/astaxie/TokenHub/actions/runs/35503249189), including browser smoke and previous-release database compatibility. [Migration Unit Tests passed](https://github.com/astaxie/TokenHub/actions/runs/35503249227); the label-gated LiteLLM Migration E2E job was skipped as configured.
- Backend: formatting, `go test ./...`, `go vet ./...`, and pinned `golangci-lint v2.12.2` passed; semantic-routing and startup regression tests also passed with `-race`.
[...]

PR #350, opened 2026-09-20
feat(routing): make Jev a configurable model routing strategy

legendtkl:
## Summary

Make Jev an explicit model routing strategy. A client requests an ordinary public model alias such as `auto-chat`; TokenHub asks Jev to choose among configured candidate models, then executes the original request through the selected Provider. This replaces the configuration experience introduced in #347 with the agreed classify-then-execute flow for both Chat Completions and Responses.

## Related Issue

N/A. Follow-up to merged PR #347. Design reference: [Building a harness with Jev](https://www.langchain.com/blog/building-a-harness-with-jev).

## Changes

- Add Jev alongside existing routing strategies, with explicit candidate Provider/model pairs, task criteria, selection instructions, default model, and confidence threshold.
- Preserve admission controls, parameter compatibility, tools, streaming, resource ordering and session affinity. Evaluation failure uses the default; failover stays within configured candidates without evaluating again. Preserve the saved candidate order across editor reloads and confidence-only edits.
- Support synchronous, streaming, and background Responses with durable ownership and exact route/resource bindings for `previous_response_id`, including IDs rewritten by output plugins. Publish background bindings atomically with successful, ownership-fenced job completion after output approval; revalidate the bound model on every continuation.
- Normalize endpoint budget declarations in capability-only catalogs and explicitly declare the shipped GPT-6 Astra Chat Completions/Responses budget fields. Preserve strict wire-parameter compatibility.
- Add schema migration 6, PostgreSQL integration coverage, regression tests, UI acceptance scenarios, and synchronized English, Simplified Chinese, and Japanese documentation. Preserve legacy overlays until reconfigured.

## Type of Change

- [x] Bug fix
- [x] New feature
- [x] Refactor or maintenance
- [x] Documentation

### 1. Enable the evaluator for a test project

Configure the existing server environment variables, substituting the real project ID and a server-side TypeSafe key:

```dotenv
TOKENHUB_SEMANTIC_ROUTING_ENABLED=true
TOKENHUB_SEMANTIC_ROUTING_PROJECTS=prj_example
TOKENHUB_TYPESAFE_API_KEY=<server-side-typesafe-key>
TOKENHUB_TYPESAFE_MODEL=jev-1.13.0
TOKENHUB_SEMANTIC_ROUTING_TIMEOUT_MS=1000
```

The project allowlist contains exact IDs separated by commas. External evaluation is off by default. All instances must share the same stable TokenHub server secret for Responses bindings.

### 2. Configure an ordinary public model

1. Create a public model named `auto-chat` (or another name) and add routes to the intended upstream models.
2. Open **Routing → Model routing policy**, select the public model, and choose **Jev Smart Routing** (`Jev 智能路由`).
3. Select candidate models, enter each candidate's task criteria, and provide selection instructions. For example, use one model for simple extraction/translation and another for complex analysis/code changes.
4. Select a default model, set minimum confidence, and click **Apply strategy** (`应用策略`). The default is used for evaluator failures or low-confidence decisions.

`auto-chat` is an example alias, not a reserved model or a separate API. Other strategies use this same public-model configuration flow. Jev is the classifier and does not need to be registered as a generation Provider.

Reimport previously saved Provider models from the updated catalog to refresh their supported parameter declarations. The shipped GPT-6 Astra entries declare `max_completion_tokens` for Chat Completions and `max_output_tokens` for Responses; existing saved Provider models are not automatically rewritten.

### 3. Call Chat Completions or Responses

Set `TOKENHUB_BASE_URL` to the gateway origin and `TOKENHUB_API_KEY` to a project API key from the allowlisted project.

```bash
curl "$TOKENHUB_BASE_URL/v1/chat/completions" \
  -H "Authorization: Bearer $TOKENHUB_API_KEY" \
  -H 'Content-Type: application/json' \
  -d '{"model":"auto-chat","messages":[{"role":"user","content":"Explain this code change."}]}'
```

```bash
curl -N "$TOKENHUB_BASE_URL/v1/responses" \
  -H "Authorization: Bearer $TOKENHUB_API_KEY" \
  -H 'Content-Type: application/json' \
  -d '{"model":"auto-chat","input":"Explain this code change.","stream":true}'
```

Both APIs support `stream: true`. Tools and generation parameters remain in the request sent to the selected target. A Responses continuation must use the same API key and public alias and set `previous_response_id` to the returned response `id`; it reuses the original route and resource without another Jev decision. An expired, unknown, foreign, removed, unavailable, disabled, or request-incompatible binding returns 409 instead of silently switching models. Background response IDs become continuable only after output approval and successful job completion. Start a new request with full input when that route is no longer usable.

Full configuration, administration PATCH examples, fallback rules, and rollout details: [English](https://github.com/legendtkl/TokenHub/blob/codex/jev-routing-strategy/docs/semantic-routing.md), [Simplified Chinese](https://github.com/legendtkl/TokenHub/blob/codex/jev-routing-strategy/docs/zh-CN/semantic-routing.md), [Japanese](https://github.com/legendtkl/TokenHub/blob/codex/jev-routing-strategy/docs/ja/semantic-routing.md).

## Screenshots

Captured from the real console with synthetic Provider/model data after passing the isolated UI acceptance suite. These show the Simplified Chinese interface; English and Japanese translations are included. No production data or credentials are present.

**Select Jev alongside the existing strategies.**

Jev strategy alongside existing routing strategies

**Configure selection instructions, default model, minimum confidence, and candidate task criteria.** The screenshot uses a test threshold of `0.8`, not a recommended accuracy target.

Jev candidate and fallback configuration

<details>
<summary>Mobile configuration at 390 × 844</summary>

The first segment shows the Apply strategy action and selection settings; the second shows both candidate criteria and the routing notes.

Mobile strategy and selection settings
Mobile candidate criteria and routing notes

</details>

## Verification

- Latest pushed head `3c32dd43da35e9923744712d9a131a79d5a5bc0d`: [all 9 CI checks passed](https://github.com/astaxie/TokenHub/actions/runs/35616222258), including the aggregate required CI check.
- Backend: formatting, uncached `go test ./...`, `go vet ./...`, golangci-lint v2.12.2, and focused Jev/semantic-routing/background-job race tests passed. Migration manifest regeneration was unchanged.
- PostgreSQL 16: the full CI integration selection and database schema suite passed, including new binding/job-completion atomicity, cancellation, ownership fencing, conflict rollback, and retry coverage on SQLite and PostgreSQL.
- Review regressions: reproduced the reported failures before fixing them. Added real bundled-catalog import and budgeted HTTP requests for both APIs with evaluation enabled/disabled; blocked/denied/cancelled output-hook continuation tests; disabled/deleted/incompatible bound-model tests; and saved candidate-order tests with a confidence-only edit and UI save/reload round trip.
- Node 22.23.1 frontend: `npm ci`, lint, typecheck, 181 domain tests, 309 component tests, production build, all 10 full-stack browser E2E tests, and all 41 isolated UI scenarios passed. Three Jev capture scenarios passed; all seven refreshed images are byte-identical to the previously inspected captures shown above.
- Repository: all 175 gate tests, full documentation co-change check against `origin/main`, UI translations, environment contract, source-size gate, and `git diff --check origin/main HEAD` passed.
- SDK: all three shipped smoke scripts (`test:deepseek`, `test:anthropic-messages`, `test:security-policy`) passed against an isolated TokenHub binary with a temporary SQLite database and the built-in mock Provider. These validate SDK and gateway protocol compatibility; they do not constitute real DeepSeek/Anthropic provider tests.
- `feature-dev`: the implementation completed three independent read-only review rounds. The subsequent four actionable owner review comments were fixed with regression coverage, replied to individually, and resolved after all latest-head CI checks passed.
- Plugin DevKit, deployment/native-install, and binary/database compatibility contract checks passed during the original feature delivery; their local suites were not rerun for these routing/UI review fixes. Latest-head CI runs their applicable checks again.
- Real TypeSafe/generation-provider calls were not run. Synthetic upstream tests validate wire behavior, routing, fallback and persistence, not workload classification accuracy or production latency. The unrelated label-gated LiteLLM migration E2E suite was not requested or enabled.

- Existing `/v1` request shapes remain unchanged. An additive schema migration (version 6) creates durable Responses bindings. Existing strategies and legacy overlays keep their behavior until edited; selecting an ordinary strategy disables new Jev classification.
- Only eligible latest-user text (at most 8192 bytes), selection instructions, and bounded candidate descriptions reach the fixed TypeSafe endpoint. System/developer content, tool definitions/results, media, credentials and headers are excluded. Evaluation remains default-off and project-allowlisted.
- Detailed bindings expire after 30 days. Non-expiring HMAC markers contain no raw response ID, prompt or account and prevent protected IDs from bypassing ownership checks through another alias after expiry. Marker storage grows with protected IDs. Bindings remain enforced after strategy changes; protected Responses and all continuations bypass the result cache.
- Evaluator usage stays separate from generation usage/customer billing; the public alias's pricing still applies. Confidence is not a calibrated task-success probability.
- Roll back routing behavior by selecting an ordinary strategy or disabling the global gate. Preserve bindings and migration history, and respect the binary/database compatibility manifest for binary rollback. No new environment variables or deployment files are introduced.

## Checklist

- [x] Environment variable changes are synchronized across examples, Compose, `start.sh`, and deployment documentation where applicable. No new variables.
- [x] `data/model-catalog.yaml` remains tracked and catalog changes were reviewed where applicable. The shipped Astra budget declarations were synchronized across the model catalog, provider catalog, and bundled OpenAI catalog.


backend/internal/server/jev_routing.go
```diff
@@ -0,0 +1,232 @@
...
+	if len(model.OutputModalities) > 0 && !slices.Contains(model.OutputModalities, "text") {
+		return false
+	}
+	if model.ContextWindow > 0 && (int64(size) > model.ContextWindow || int64(budget) > model.ContextWindow-int64(size)) {
+		return false
+	}
+	for parameter, used := range parameters {
+		if used && len(model.SupportedParameters) > 0 && !slices.Contains(model.SupportedParameters, parameter) {
+			return false
+		}
```
astaxie, line 94: **[P1] Reconcile strict parameter admission with the shipped catalog**

`SupportedParameters` becomes a hard rejection list here, but normal catalog imports do not declare basic output-budget fields. For example, the bundled GPT-6 Astra entry imports `[image_input reasoning reasoning_effort response_format tools]`; `providerModelFromCatalog()` copies that list unchanged, and `catalogModelParameters()` does not add token-budget fields. Using this actual imported metadata for the configured candidates, I reproduced HTTP 503 `jev_no_eligible_candidates` for both Chat with `max_completion_tokens: 128` and Responses with `max_output_tokens: 128`, before either evaluation or generation. Thus enabling Jev breaks otherwise valid budgeted requests, including when evaluation is disabled and the default should serve them. Please synchronize catalog/ingestion declarations with this stricter contract, preserving protocol-specific wire names, and cover a normally imported catalog model rather than only fixtures with empty parameter lists.

backend/internal/server/response_jobs.go
```diff
@@ -585,6 +590,9 @@ func (s *Server) processResponseJob(job ResponseJob, owner string, leaseTTL time
 		return
 	}
 	response, route, usage, attempts, invokeErr := s.executeRoutedResponsesContext(ctx, envelope.Headers, routed, request)
+	if invokeErr == nil {
+		invokeErr = s.bindJevResponse(ctx, routed.Call, route, response, job.ID)
```
astaxie: **[P2] Publish background bindings only after output approval**

This makes the public `job.ID` a valid continuation before `response_post` and `guardrail_post` accept the result. The caller already knows that ID from submission. With the first `guardrail_post` deliberately blocked, I reproduced a job still marked `running` while a request using `previous_response_id=job.ID` returned 200 and reached the upstream with its real response ID. The binding also remains usable after the original job fails with `gateway_hook_denied`. This exposes pending or rejected upstream state for continuation; the continuation still has its own hooks, but it can already consume a result the original output policy has not approved. Publish the binding only after successful output processing/completion, or require successful job state during lookup. The synchronous path already binds after its output hooks.

backend/internal/server/jev_response_binding.go
```diff
@@ -0,0 +1,225 @@
...
+		for _, candidate := range policy.Candidates {
+			permitted = permitted || (candidate.ProviderID == binding.ProviderID && candidate.ProviderModel == binding.ProviderModel)
+		}
+		if permitted {
+			for _, route := range routed.Routes {
+				if route.Route.ID == binding.RouteID && route.Provider.ID == binding.ProviderID && route.ProviderModel == binding.ProviderModel && routeResourceID(route) == binding.ResourceID {
+					routed.Routes = []RouteSelection{route}
+					routed.Call.JevResponseBound = true
+					setRawJSONField(req.raw, "previous_response_id", binding.UpstreamID, true)
+					return nil
```
astaxie, line 176: **[P2] Revalidate the bound candidate's catalog status**

This early return skips the `SemanticProviderModels` / `jevModelFits` eligibility checks used for new Jev requests. After creating a response on `model_1` and setting that `ProviderModel` to `disabled`, I verified that a fresh request correctly selected `model_0`, but a continuation using the original response ID still invoked disabled `model_1` and returned 200. Matching an admitted route does not enforce catalog status. Please recheck the bound candidate's current eligibility and reject an unavailable candidate with 409, while retaining the existing no-reclassification/no-failover behavior. Otherwise disabling a candidate does not stop its existing response chains, contrary to the documented active-catalog requirement.

frontend/features/admin/views/semantic-routing-policy.tsx
```diff
@@ -1,45 +1,104 @@
...
+    const catalog = data.providerModels.find(model => model.provider_id === route.provider_id && model.upstream_model === route.provider_model);
+    unique.set(key, previous ?? { id: route.id, provider_id: route.provider_id, provider_model: route.provider_model, criteria: catalog?.metadata?.routing_description ?? "" });
+  }
+  return [...unique.values()];
+}
+
+export function initialJevPolicy(model: Model, routes: ModelRoute[], data: AppData): SemanticRoutingPolicy {
+  const saved = readSemanticRoutingPolicy(model);
+  const options = jevModelOptions(routes, data, saved);
+  const candidates = saved.candidates ? options.filter(option => saved.candidates?.some(candidate => candidate.id === option.id)) : options;
```
astaxie: **[P2] Preserve saved candidate order when reopening the editor**

Filtering `options` reconstructs candidates in route order instead of the saved configuration order. Calling the actual `initialJevPolicy` with routes `[A, B, C]`, saved candidates `[A, C, B]`, and default `A` produces `[A, B, C]`; changing only confidence then saves that reordered list. The UI can itself create `[A, C, B]` by deselecting and reselecting `B`. Since the backend uses candidate configuration order for fallback, this unrelated edit silently changes which model is tried when `A` is unavailable or fails. Preserve `saved.candidates` order while filtering out candidates whose routes no longer exist.

astaxie (commented):
Reviewed the routing, Responses continuation/binding, and administration changes at 2a22b0020edb99a2205c991e6d95a4a4dc0e2aac. Four actionable issues are documented inline with reproduction details.

Validation: the existing backend suite (`go test ./... -count=1`), all 175 repository gate tests, documentation/UI translation checks, environment contract, source-size check, and diff whitespace check passed locally. The head's nine published CI checks also passed. Additional isolated reproductions exposed the cases below; no product code was changed.

backend/internal/server/jev_routing.go
legendtkl: Fixed in 3c32dd43.

The shipped GPT-6 Astra entries now explicitly declare `max_completion_tokens` and `max_output_tokens` in all three catalog sources. Catalog normalization also supplies an advertised endpoint's baseline budget field when a capability-only entry declares no budget fields; explicit budget declarations remain authoritative, so this does not treat cross-protocol fields as interchangeable.

`TestJevImportedCatalogBudgetRequests` imports the actual bundled catalog through the admin API and exercises budgeted Chat Completions and Responses requests with evaluation both enabled and disabled. It verifies the original budget field reaches the selected/default Provider. `TestCatalogBudgetDeclarationsKeepWireNames` covers the restrictive cases. The usage guide now notes that previously saved Provider models must be reimported to refresh their declarations.

backend/internal/server/response_jobs.go
legendtkl: Fixed in 3c32dd43.

Background execution now captures the original upstream response ID without publishing a binding. `FinalizeResponseJob` persists that binding in the same database transaction as successful, ownership-fenced job completion, after all output hooks approve. Pending, denied, cancelled, and stale-worker completions cannot publish a binding; a binding conflict rolls back job completion as well.

`TestJevBackgroundBindingWaitsForOutputApproval` covers blocking/denying/cancelling both response-post and guardrail-post hooks through HTTP continuations. `TestJevBackgroundBindingAtomicCompletion` and its PostgreSQL integration invocation cover cancellation, lost ownership, conflict rollback, and a successful retry while preserving the original upstream ID.

backend/internal/server/jev_response_binding.go
legendtkl: Fixed in 3c32dd43.

A bound continuation now reloads the exact bound Provider/model metadata and applies the same active-status, wire-parameter, modality, and context eligibility checks used for a new Responses request. Missing, disabled, or incompatible models return `409 jev_previous_route_unavailable` without another evaluation or upstream call; metadata lookup failures remain 503. The exact route/resource binding is preserved and never fails over.

`TestJevContinuationRevalidatesCandidateMetadata` covers disabled and deleted models, unsupported parameters, insufficient context, and incompatible output modalities. It also verifies that a fresh request can use the remaining eligible candidate.

frontend/features/admin/views/semantic-routing-policy.tsx
legendtkl: Fixed in 3c32dd43.

`initialJevPolicy` now filters the saved candidate array in its existing order instead of rebuilding it from route order. A confidence-only edit therefore preserves `[A, C, B]`, and removing a missing route preserves the relative order of the remaining candidates.

Added component regressions for both cases. The UI acceptance scenario also creates a non-route-order candidate list, saves and reloads it, changes only confidence, and verifies the second saved payload keeps that order.

astaxie (commented):
Re-review of the latest head `3c32dd43`: the four previously reported issues are covered by the new transaction, catalog, continuation, and UI tests. I found two remaining compatibility/configuration issues below.

The published CI checks for this head are green. I independently reproduced both cases with the bundled catalog and the current frontend helper.

backend/internal/server/provider_catalog_parameters.go
```diff
@@ -0,0 +1,25 @@
...
+	"slices"
+	"strings"
+)
+
+// Capability-only catalogs omit routine token budgets. Expand only an advertised
+// endpoint's baseline wire field; explicit budget declarations remain authoritative.
+func catalogBudgetParameters(parameters []string, endpoints string) []string {
+	for _, budget := range []string{"max_tokens", "max_completion_tokens", "max_output_tokens"} {
+		if slices.Contains(parameters, budget) {
+			return parameters
```
astaxie: **[P1] Keep the Chat Completions budget field when another budget field is declared**

This early return treats any one budget name as satisfying every advertised endpoint. The shipped GPT-6 Astra metadata contains `max_completion_tokens` and `max_output_tokens` but no `max_tokens`, while it advertises Chat Completions. `jevModelFits` checks the actual Chat wire field `max_tokens` as an exhaustive capability, so an ordinary `/v1/chat/completions` request with `max_tokens: 128` now has no eligible candidate and returns 503 `jev_no_eligible_candidates` before evaluation or the Provider. I reproduced this on this head after importing the actual bundled catalog; the imported parameters were `[image_input max_completion_tokens max_output_tokens reasoning reasoning_effort response_format tools]`. Add `max_tokens` whenever `chat/completions` is advertised (even if another endpoint uses a different budget field), or otherwise reconcile the protocol-specific declarations. The previous fix covers `max_completion_tokens`/Responses but leaves the existing Chat API shape broken.

frontend/features/admin/views/semantic-routing-policy.tsx
```diff
@@ -1,45 +1,104 @@
...
+    const catalog = data.providerModels.find(model => model.provider_id === route.provider_id && model.upstream_model === route.provider_model);
+    unique.set(key, previous ?? { id: route.id, provider_id: route.provider_id, provider_model: route.provider_model, criteria: catalog?.metadata?.routing_description ?? "" });
+  }
+  return [...unique.values()];
+}
+
+export function initialJevPolicy(model: Model, routes: ModelRoute[], data: AppData): SemanticRoutingPolicy {
+  const saved = readSemanticRoutingPolicy(model);
+  const options = jevModelOptions(routes, data, saved);
+  const candidates = saved.candidates ? saved.candidates.filter(candidate => options.some(option => option.id === candidate.id)) : options;
```
astaxie: **[P2] Reconcile saved candidates after a route's upstream model changes**

This preserves the saved candidate object solely by `id`. If route `r1` was saved as `{provider_id: p, provider_model: old}`, an admin edits that same route to `new`, and then changes only Jev confidence, `initialJevPolicy` returns the stale `{provider_model: old}` candidate because the current option still has id `r1`. Applying the policy then sends the old pair; the backend rejects it with `Jev candidates must reference this model's configured routes`, so the policy cannot be saved until the user toggles the candidate off and on. The editor signature also does not include provider/model fields, so this can survive the route edit without a remount. Preserve the saved order, but replace each retained candidate with the current option data when its route id remains present.

backend/internal/server/provider_catalog_parameters.go
legendtkl: Fixed in 94a3da59.

Removed the global early return: each advertised endpoint now independently adds its own baseline budget field, while retaining explicitly declared fields and leaving the outgoing request's parameter names unchanged. The three shipped Astra catalog sources also explicitly include `max_tokens`, alongside `max_completion_tokens` and `max_output_tokens`, so catalogs without raw endpoint metadata work as well.

The regression imports the actual bundled catalog through the admin API and covers all three budget fields with evaluation enabled and disabled (six HTTP cases), asserting both the chosen/default Provider and unchanged wire fields. The new `max_tokens` cases reproduced the reported 503 before the fix. Normalization tests cover pre-existing budget declarations and endpoint-specific behavior. Usage examples and all three language guides were updated; previously saved Provider models still need reimporting to refresh their declarations.

frontend/features/admin/views/semantic-routing-policy.tsx
legendtkl: Fixed in 94a3da59.

`initialJevPolicy` now walks saved candidates in saved order and uses the matching current option data. The editor key also includes `provider_id` and `provider_model`, so changing either field refreshes a mounted editor instead of retaining its stale draft. Reconciled candidates are marked as unsaved so Apply is available immediately when their criteria are valid. The replacement option uses its current catalog criteria; when none are available, criteria can be entered directly without toggling the candidate.

Five new component regressions reproduced the stale-data failures and now pass, covering both Provider and upstream-model edits, retained order/default, mounted-editor refresh, confidence-only save, and missing criteria. The UI acceptance scenario edits the real route form, changes only Jev confidence, verifies the outgoing current Provider/model pair and saved order, and reloads successfully.

Issue #349, opened 2026-09-20
tokenhub 0.8.0配置本地reranker模型无法使用


dlzero996-stack:
### Affected area

Provider and model routing

### TokenHub version

0.8.0

### Deployment method

Docker Compose with PostgreSQL

### Current behavior

操作1：配置bge-reranker-v2-m3重排模型后，识别的类型为chat，无法使用重排模型
操作2：接入阿里云api后，无法识别出重排模型名称qwen-rerank

<img width="1378" height="344" alt="Image" src="https://github.com/user-attachments/assets/e040da22-1bee-43c9-b5fd-2a1ca6fb9cf5" />

<img width="805" height="647" alt="Image" src="https://github.com/user-attachments/assets/a6b2d801-2cfb-4234-895b-24350838b70d" />

### Expected behavior

provider模型上游模型识别参数应为rerank

### Steps to reproduce

Provider 渠道：配置重排模型

### Relevant logs or configuration

```shell

```

### Submission checklist

- [x] I searched open and closed issues for duplicates.
- [x] The issue title and all explanatory text are written in English.
- [x] I removed or redacted all credentials and other secrets.

PR #351, opened 2026-09-21
feat(retrieval): support text embeddings and reranking

deepjerry-ai:
## Summary

Add governed text embedding and reranking support for retrieval applications. Embedding requests retain their vector contract across providers, hooks, cache hits, and failover; `/v1/rerank` returns a consistent text-ranking response with independent tenant and supplier metering.

## Related Issue

Related to #349.

## Changes

- Implement dense float/base64 embedding contracts, explicit provider protocols, stable vector-space routing, model discovery, and cache/request-hook invariants.
- Add text reranking adapters, gateway authentication and quota controls, provider plugin support, administrator testing, and provider configuration controls.
- Preserve missing versus explicitly zero usage, validate contradictory usage, price native search units independently of tokens, and retain provider cost for free tenant requests.
- Validate publication capabilities and prices, check all initial embedding routes together, allow old routes to be disabled, and expose retrieval pricing in the model creation wizard.
- Keep embedding, reranking, and shared metering changes in separate commits; add synchronized English, Simplified Chinese, and Japanese guides and OpenAPI documentation.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Refactor or maintenance
- [x] Documentation

## Verification

- Backend: full `go test ./...`, `go vet ./...`, and `golangci-lint run ./...`.
- Frontend: lint, typecheck, unit tests, production build, and 10 browser E2E tests.
- Named UI acceptance: embedding/rerank provider settings and model creation pricing; synthetic fixtures, request assertions, and screenshot inspection.
- Plugin Go SDK tests, vet, and lint; repository test, translation, environment, source-line, and diff gates.
- All eight CI jobs passed on `61774c89b848505be1748422f915f29193d05617` ([run](https://github.com/astaxie/TokenHub/actions/runs/35631784701)).
- Grok findings were verified against code and worthwhile issues fixed. Exactly two Codex review rounds completed. Neither reported P0/P1; all five P2 findings from round one were fixed. The final review reports four P2 follow-ups listed below, within the requested P1 release gate.
- Real supplier credentials and version-specific OpenAI SDK, Dify, LangChain, and LlamaIndex acceptance were not exercised. Local integration tests use SQLite; PR CI passed PostgreSQL integration and N-1 database compatibility checks.

No SQL tables, columns, indexes, migrations, environment variables, or deployment changes. Configuration and metering evidence use existing JSON fields. This scope is text-only; sparse, quantized, multimodal, streaming, and asynchronous batch retrieval are excluded.

New routes require supported protocols and confirmed prices. Existing published routes retain compatibility for unrelated edits. Multiple embedding sources require matching explicit vector-space identities. Embedding cache plugins must store and return the host-provided `cache_key`; legacy or mismatched entries become misses. Request hooks cannot change the client vector contract, and final responses are validated after hooks. Missing usage remains unknown instead of becoming free. Native search-unit prices are configured in model metadata; token-based versioned rate cards are not extended in this PR.

### Known P2 follow-ups from final review

- Publication through resource groups or automatic resource selection validates parent-provider pricing/protocols without fully matching resource overrides; differing protocols can publish a route that is later unavailable.
- Provider inventory capability flags consider the parent configuration and can hide otherwise callable models whose protocol is configured only on a resource.
- Capability annotation performs repeated full provider-list reads per model; load providers once to avoid query amplification on large inventories.
- Rerank TPM reservation estimates query/documents but excludes optional `instruction`; long instructions can underreserve tokens before settlement.

These findings are retained transparently rather than starting a third review round. No merge or deployment is requested.

## Checklist

- [x] Environment variable changes are synchronized across examples, Compose, `start.sh`, and deployment documentation where applicable. (No environment changes.)
- [x] `data/model-catalog.yaml` remains tracked and catalog changes were reviewed where applicable. (No catalog data changes.)


backend/internal/server/provider_catalog.go
```diff
@@ -840,7 +846,9 @@ func normalizeProviderBaseURL(id string, raw string) string {
 func normalizeModelModality(value string) string {
 	value = strings.ToLower(strings.TrimSpace(value))
 	switch {
-	case strings.Contains(value, "embed"):
+	case strings.Contains(value, "rerank"):
```
astaxie, line 850: [P1] Migrate existing reranker inventory rows. This inference only affects newly discovered or backfilled rows. On a 0.8.0 upgrade, an existing `bge-reranker-v2-m3` ProviderModel is already stored as `chat`, so `backfillProviderModelsFromRoutes` skips it; when inventory is missing, `providerModelFromRoute` later overwrites this inference with the legacy public-model modality. The exact #349 deployment therefore remains unusable after upgrade. Add an idempotent reconciliation for existing rerank-shaped rows, and do not replace inferred `rerank` with legacy `chat` metadata.

data/builtin-plugins/providers/alibaba/catalog.json
```diff
@@ -3190,6 +3190,26 @@
         }
       },
       "type": "chat"
+    },
+    {
+      "id": "qwen3-rerank",
```
astaxie, line 3195: [P1] Update the canonical provider catalog, not only this generated package copy. Runtime startup and imports load `data/provider-catalog.json`, and `tools/generate-builtin-provider-plugins.mjs` generates these package files from it. As committed, neither this model nor the Alibaba CN `gte-rerank-v2` entry reaches runtime discovery, while the package manifests remain stale (`models_count` 80 vs. 81 and 111 vs. 112). Add both models to the canonical source and regenerate the packages.

backend/internal/server/gateway_embeddings_http.go
```diff
@@ -20,6 +20,11 @@ func (s *Server) handleEmbeddings(w http.ResponseWriter, r *http.Request) {
 		writeError(w, r, NewHTTPError(400, "missing_model", "model is required"))
 		return
 	}
+	if err := validateEmbeddingRequest(req); err != nil {
```
astaxie, line 23: [P1] Enforce embedding modality and tenant pricing after admission, matching the rerank handler. `validateEmbeddingRequest` checks payload shape only, so a published `chat` model can still be called through `/v1/embeddings` (this returns 200 with a capable route), bypassing the publication-time embedding price requirement and settling with chat pricing. Reject unless `call.Model.Modality == "embedding"` and an embedding price or explicit free-price confirmation is configured before hooks and routing.

backend/internal/server/retrieval_usage.go
```diff
@@ -0,0 +1,120 @@
...
+	var count float64
+	if err != nil || json.Unmarshal(encoded, &count) != nil || count < 0 || math.Trunc(count) != count || count >= float64(math.MaxInt64) {
+		return nil, false
+	}
+	n := int64(count)
+	return &n, true
+}
+func retrievalUsage(body map[string]any, searchUnits bool) Usage {
+	usage := usageFromMap(body)
+	evidence := &RetrievalUsageEvidence{Unit: "token", Source: "unreported"}
```
astaxie, line 29: [P1] Keep the admission reservation when retrieval token usage is unavailable. Cohere-style rerank reports `search_unit` evidence and TEI may leave usage unreported, so `meteredTokens` is zero; `executeRoutedWithStore` then leaves `RateLimitTokens` at zero and `FinishCall` refunds the entire estimate. Two sequential requests can both pass a TPM limit sized for one request. Preserve the request estimate as the quota actual when retrieval evidence has no token quantity (or define a native-unit conversion), and add sequential TPM coverage.

backend/internal/server/retrieval_support.go
```diff
@@ -0,0 +1,218 @@
...
+				found = true
+				break
+			}
+		}
+	}
+	if !found {
+		return nil
+	}
+	// Publication must use the same resource overrides as execution.
+	if route.ProviderResourceID != "" {
```
astaxie: [P2] Validate every effective resource configuration used by this route. `ProviderResourceID` covers only an explicit resource; routes with `ResourceGroup` or automatic selection remain validated against the parent Provider even though `LoadRouteCandidates` later expands them and merges each resource `Options`. A resource-only `rerank_protocol` or `embedding_protocol` can make a valid route fail publication, while an incompatible override can pass here and fail at runtime.

backend/internal/server/provider_models.go
```diff
@@ -96,6 +96,12 @@ func (s *Server) handleAdminProviderModels(w http.ResponseWriter, r *http.Reques
 		}
 		models = filtered
 	}
+	for i := range models {
+		if provider, ok := s.providerByID(models[i].ProviderID); ok {
+			supported := s.providerRetrievalSupport(provider, models[i].Modality) && retrievalTextInputSupported(models[i])
```
astaxie: [P2] Compute support from effective resource configurations too. A Provider with `rerank_protocol` or `embedding_protocol` configured only on a resource is marked `call_supported=false` here, and the model-creation selectors filter it out even though runtime routing through that resource is valid. Treat the model as callable when at least one eligible effective resource supports it, while handling parent fallback consistently.

backend/internal/server/provider_models.go
```diff
@@ -96,6 +96,12 @@ func (s *Server) handleAdminProviderModels(w http.ResponseWriter, r *http.Reques
 		}
 		models = filtered
 	}
+	for i := range models {
+		if provider, ok := s.providerByID(models[i].ProviderID); ok {
```
astaxie: [P2] Load Providers once before this loop and index them by ID. `providerByID` calls `ListProviders()` for every model, so listing N Provider models performs N full Provider-list reads. This turns a common admin refresh into query amplification as inventories grow.

astaxie (commented):
Inline review findings.

backend/internal/server/retrieval_support.go
```diff
@@ -0,0 +1,218 @@
...
+	if provider.Type == ProviderMock {
+		return nil
+	}
+	search := model.Modality == "rerank" && providerRerankProtocol(provider) == "cohere"
+	if !retrievalPriceConfigured(model, search, false) {
+		return NewHTTPError(400, "retrieval_price_required", "Configure the tenant price or explicitly confirm a free token price before publishing")
+	}
+	for _, upstream := range s.store.ListProviderModels() {
+		if upstream.ProviderID == provider.ID && upstream.UpstreamModel == route.ProviderModel {
+			if !retrievalTextInputSupported(upstream) || (upstream.Modality != "" && upstream.Modality != model.Modality) {
```
astaxie: [P2] Normalize Provider-model modality aliases before this exact comparison. The tracked SiliconFlow model `netease-youdao/bce-reranker-base_v1` has `type: "reranker"`; `providerModelFromCatalog` persists that verbatim, while public models use `rerank`, so publishing this valid catalog model always returns `model_operation_mismatch`. Canonicalize aliases during import, including existing rows, or compare normalized values here.

backend/internal/server/api_key_rate_limit.go
```diff
@@ -20,6 +20,12 @@ func requestTokenReservation(payload any) int64 {
 		input = saturatingAddNonNegative(input, estimateRawJSONTokens(request.raw["tools"]))
 		input = saturatingAddNonNegative(input, estimateRawJSONTokens(request.raw["text"]))
 		return saturatingAddNonNegative(input, outputTokenReservation(int64(request.MaxTokens)))
+	case RerankRequest:
+		var tokens int64
+		for _, document := range request.Documents {
+			tokens = saturatingAddNonNegative(tokens, EstimateTextTokens(request.Query+" "+document))
```
astaxie, line 26: [P2] Include `request.Instruction` in the rerank TPM reservation. Qwen, Jina, and DashScope forward this field upstream, so a long instruction consumes input tokens without being reserved and can exceed the configured TPM limit before settlement.

backend/internal/server/retrieval_support.go
```diff
@@ -0,0 +1,218 @@
...
+			switch providerEmbeddingProtocol(provider) {
+			case "openai", "cohere", "jina", "voyage", "dashscope", "tei":
+				return true
+			default:
+				return false
+			}
+		}
+		return true
+	case "rerank":
+		return adapterSupports(descriptor, AdapterCapabilityRerank) && (provider.Type != ProviderOpenAICompatible || validRerankProtocol(providerRerankProtocol(provider)))
```
astaxie: [P1] Do not block the built-in qwen/local adapters. Both types register the same OpenAICompatibleAdapter and the admin UI exposes rerank_protocol, but their descriptors omit AdapterCapabilityRerank (only openai_compatible received it). This gate therefore rejects publication and runtime routing for a built-in local BGE or Qwen Provider even when a valid protocol is configured, leaving the linked local-reranker use case broken. Advertise rerank for those aliases and apply protocol validation to every OpenAICompatibleAdapter-backed type.

backend/internal/server/retrieval_support.go
```diff
@@ -0,0 +1,218 @@
...
+			return nil
+		}
+	}
+	return NewHTTPError(400, "provider_model_required", "Import and configure the upstream model before publishing")
+}
+func (s *Server) pricedRerankRoutes(model Model, routes []RouteSelection) []RouteSelection {
+	result := make([]RouteSelection, 0, len(routes))
+	models := s.store.ListProviderModels()
+	for _, route := range routes {
+		if !s.providerRetrievalSupport(route.Provider, "rerank") {
```
astaxie: [P2] Preserve routes implemented by a scoped `provider_call` hook. The caller first keeps routes supported either by the native adapter or by a rerank provider-call hook, but this unconditional native `providerRetrievalSupport` check drops every hook-only route before `executeRoutedRerank` can invoke the hook. Embeddings has no contradictory second capability filter. Carry the hook-provided capability through this pricing filter and add a plugin-only rerank route test.

backend/internal/server/embedding_invariants.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+	before, _, err := embeddingInputCount(original.Input)
+	if err != nil {
+		return err
+	}
+	after, _, err := embeddingInputCount(next.Input)
+	if err != nil {
+		return err
+	}
+	if original.Model != next.Model || before != after || !reflect.DeepEqual(original.Dimensions, next.Dimensions) ||
```
astaxie: [P2] Preserve tokenized-input semantics across hook patches. `embeddingInputCount` returns whether each request uses token IDs, but both flags are discarded here, so a hook can replace `[1,2,3]` with arbitrary text (or rewrite the token IDs) and pass as long as both represent one input. That violates the "only text may be rewritten" invariant stated above and silently changes the client embedding semantics. Reject text/token mode changes and require tokenized input to remain identical, with a regression test.

backend/internal/server/embedding_contract.go
```diff
@@ -0,0 +1,272 @@
...
+			return 0, false, embeddingRequestError("input must not be empty")
+		}
+		if _, ok := values[0].(float64); ok {
+			for _, v := range values {
+				n, ok := v.(float64)
+				if !ok || n < 0 || math.Trunc(n) != n || n > math.MaxInt32 {
+					return 0, true, embeddingRequestError("token IDs must be nonnegative integers")
+				}
+			}
+			return 1, true, nil
```
astaxie, line 84: [P1] Reserve tokenized embedding inputs by token count. This branch correctly recognizes a flat token-ID array, but `requestTokenReservation` later serializes it through `EmbeddingInputText` and estimates text length. For inputs such as `[0, 0, ...]`, that reserves roughly half the actual tokens, so one request can exceed the configured TPM limit before provider usage is settled. Sum token-array lengths directly and add admission coverage for flat and batched token IDs.

backend/internal/server/gateway_embeddings_http.go
```diff
@@ -65,6 +76,18 @@ func (s *Server) handleEmbeddings(w http.ResponseWriter, r *http.Request) {
 		writeError(w, r, err)
 		return
 	}
+	space, err := s.embeddingSpaceContract(r.Context(), req.Model, nil)
+	if err != nil {
+		s.finishFailedRoutedCall(r, RoutedCall{Call: call}, nil, Usage{}, err, auditPayload)
+		writeError(w, r, err)
+		return
+	}
+	call.EmbeddingCacheKey, err = embeddingCacheKey(call, space, req)
```
astaxie, line 85: [P1] Bind the cache key to the actual transformed embedding input. Route-scoped `request_transform` hooks run later inside `executeRoutedEmbeddings` and are allowed to rewrite text, so two failover routes can embed different text under this same pre-transform key. A response produced by route B can then be returned as a cache hit when route A would have transformed the request differently. Either disallow text rewrites at the route-scoped stage or include the deterministic transformed request and hook revision in the cache contract.

backend/internal/server/gateway_embeddings_http.go
```diff
@@ -107,6 +142,7 @@ func (s *Server) handleEmbeddings(w http.ResponseWriter, r *http.Request) {
 		writeError(w, r, err)
 		return
 	}
+	routed.Routes = compatibleEmbeddingRoutes(routed.Routes, space)
```
astaxie, line 145: [P1] Revalidate embedding routes against the current provider inventory before execution. An existing route can become invalid without being edited: the model PATCH path can change a routed chat model to `embedding`, and ProviderModel modality or pricing can also change after publication. This handler filters adapter capability and vector space only, so it can invoke `/embeddings` for a chat or unpriced upstream and record an unknown supplier cost as zero. Add the embedding equivalent of `pricedRerankRoutes`, or validate all affected routes when these settings change.

backend/internal/server/provider_models.go
```diff
@@ -274,7 +285,7 @@ func providerModelFromCatalog(providerID string, model ProviderCatalogModel) Pro
 		CanonicalName:             firstNonEmpty(model.CanonicalName, canonicalModelName(model.ID, model.DisplayName)),
 		Category:                  standardModelCategory(firstNonEmpty(model.Category, inferModelCategory(model.ID, model.DisplayName))),
 		Family:                    firstNonEmpty(model.Family, inferModelFamily(model.ID)),
-		Modality:                  firstNonEmpty(model.Type, normalizeModelModality(model.ID)),
+		Modality:                  canonicalProviderModality(firstNonEmpty(model.Type, normalizeModelModality(model.ID))),
```
astaxie: [P1] Let rerank-shaped IDs override stale catalog `chat` types. `firstNonEmpty` always chooses a nonempty `model.Type`, and tracked catalogs still label obvious rerank models as `chat` (for example `voyage/rerank-2.5` and `nvidia/rerank-qa-mistral-4b`). Startup reconciliation only repairs rows already in the database, so importing either model afterward persists it as `chat` and keeps it unavailable for rerank publication until a restart. Canonicalize the source catalogs or apply rerank-name inference when the catalog type is the legacy `chat` fallback.

backend/internal/server/rerank_contract.go
```diff
@@ -0,0 +1,187 @@
...
+		return nil, err
+	}
+	var raw any
+	if err = json.Unmarshal(encoded, &raw); err != nil {
+		return nil, err
+	}
+	// Validation after safety hooks must not reconstruct documents from input.
+	check := request
+	check.ReturnDocuments = false
+	if _, err := normalizeRerankResponse(raw, "jina", check); err != nil {
```
astaxie, line 160: [P2] Enforce the normalized ranking order here. `normalizeRerankResponse` sorts a temporary result, but that value is discarded and `validateRerankResult` returns the original response. A provider-call, cache, or post-processing hook can therefore return scores out of descending order and still pass final validation, exposing an inconsistent ranking to clients. Validate monotonic order while preserving safety redactions, or return a normalized copy that retains those edits.

astaxie (commented):
Re-reviewed the current head. The confirmed issues are inline below.

frontend/features/admin/views/model-directory.tsx
```diff
@@ -235,15 +236,15 @@ function ExternalModelsTable({ api, data, models, readOnly, busy, onOpenRoutes,
...
                   </td>
                   <td><StatusPill status={publication === "published" ? "active" : "disabled"} label={tx(publicationLabel(publication))} /></td>
                   <td><RuntimeStatus state={runtime} active={activeRoutes.length} total={routes.length} /></td>
                 </> : <td><StatusPill status="active" label={tx("当前账号可用")} /></td>}
-                <td><strong>{priceMetric(model.input_price_usd_per_1m)}</strong><span>{tx("输入")} · {priceMetric(model.output_price_usd_per_1m)} {tx("输出")}</span></td>
+                <td>{model.modality === "rerank" && model.metadata?.search_unit_price_usd?.trim() ? <><strong>${formatMoney(Number(model.metadata.search_unit_price_usd))}</strong><span>{tx("搜索单元价格 USD/次")}</span>{model.input_price_usd_per_1m != null ? <span>{tx("输入")} · {modelTokenPriceMetric(model)}</span> : null}</> : <><strong>{modelTokenPriceMetric(model)}</strong><span>{model.modality === "embedding" ? "Embedding" : tx("输入")}{model.modality !== "embedding" && model.modality !== "rerank" ? <> · {priceMetric(model.output_price_usd_per_1m)} {tx("输出")}</> : null}</span></>}</td>
```
astaxie: [P2] Format this user-visible USD value with `Intl.NumberFormat(languageLocale(), ...)`. `formatMoney` uses fixed decimal formatting, and manually prefixing `$` violates the repository rule that currencies follow the selected application language.

backend/internal/server/gateway_rerank_http.go
```diff
@@ -0,0 +1,176 @@
...
+		return
+	}
+	decision, err := s.evaluateOutboundGuardrails(r.Context(), call.Project.ID, rerankGuardrailTargets(&req))
+	auditPayload := guardrailRequestAuditPayload(req.Model, decision, req)
+	if err != nil {
+		s.finishFailedRoutedCall(r, RoutedCall{Call: call}, nil, Usage{}, err, auditPayload)
+		writeError(w, r, err)
+		return
+	}
+	resp, usage, hit, err := s.runGatewayCacheLookupHooks(r.Context(), call, req)
```
astaxie, line 78: [P1] Bind rerank caching to the request that is actually sent upstream. The lookup occurs here before route selection, but executeRoutedRerank later permits a route-scoped request_transform hook to rewrite query, documents, instruction, and truncation. Cache writes then use the original req at line 171. A response generated from one route-specific transformed request can therefore be returned for another route before its transform runs. Freeze rerank content after cache binding, as embeddings now does, or make the cache contract include the deterministic route-transformed request and hook revision.

backend/internal/server/gateway_rerank_http.go
```diff
@@ -0,0 +1,176 @@
...
+		writeError(w, r, err)
+		return
+	}
+	usage, err = s.runGatewayUsageAttributionHooks(r.Context(), routed.Call, route, resp, usage, providerRouteProtocolRerank)
+	if err != nil {
+		s.finishFailedRoutedCall(r, routed, attempts, usage, err, auditPayload)
+		writeError(w, r, err)
+		return
+	}
+	usage, err = pluginRetrievalUsage(resp, usage, retrievalResultUsesSearchUnits(resp, usage))
```
astaxie, line 158: [P1] Determine the expected billing unit from the selected effective route, not from plugin-reported usage. On a token-priced Jina route, a provider_call hook can return retrieval_evidence.unit = "search_unit" with no token counters; retrievalResultUsesSearchUnits treats that evidence as authorization, so validation accepts it even though route publication required only token prices. nativeRetrievalCost then cannot price the evidence because no search-unit price was required, and token pricing sees zero tokens, recording zero tenant and provider cost. Use providerRerankProtocol(route.Provider) == "cohere" or the equivalent effective-route contract, and reject mismatched units.

backend/internal/server/rerank_contract.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+	previousScore := math.Inf(1)
+	for _, value := range items {
+		item := value.(map[string]any)
+		score := item["relevance_score"].(float64)
+		if score > previousScore {
+			return nil, NewHTTPError(502, "invalid_rerank_response", "Final rerank results must be ordered by descending relevance_score")
+		}
+		previousScore = score
+		if document, exists := item["document"]; exists && document != nil {
```
astaxie, line 180: [P2] Reject returned documents when return_documents is false. Provider-call, cache, and post-hook output reaches this final validator, but any document object with string-valued text is accepted regardless of request.ReturnDocuments and returned unchanged. This violates the response flag and can reintroduce document content into client responses and logs that requested indices and scores only. Continue preserving redacted or omitted documents when the flag is true, but require document to be absent when it is false.

astaxie (commented):
Re-reviewed the current head. Three remaining issues are inline below.

backend/internal/server/gateway_rerank_http.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+	routed.Routes = s.pricedRerankRoutes(routed.Call, s.routesWithAdapterCapabilityOrProviderCall(routed.Call, routed.Routes, AdapterCapabilityRerank, providerRouteProtocolRerank))
+	if len(routed.Routes) == 0 {
+		err := NewHTTPError(http.StatusNotImplemented, "provider_capability_not_supported", "No rerank route has a supported protocol and configured provider/tenant prices")
+		s.finishFailedRoutedCall(r, routed, nil, Usage{}, err, auditPayload)
+		writeError(w, r, err)
+		return
+	}
+	call = routed.Call
+	cacheRoute := routed.Routes[0]
```
astaxie: [P2] Make failover cache entries discoverable

The lookup key is built only from `routed.Routes[0]`, but after failover lines 175-176 write the successful response under the actual winning route's route-specific key. While the first candidate remains eligible (and again after it recovers), an identical request only probes the first candidate's key, so the cached second-route result can never hit and the gateway repeats the upstream failure and failover. Probe the eligible route keys in plan order or use a plan-stable cache contract that still binds protocol/upstream configuration, and add a two-route regression where route 1 fails, route 2 succeeds, then the same request uses the cached result.

astaxie (changes requested):
Re-reviewed the complete current diff at head `177e7344e`. The latest commits fix the prior rerank cache-input binding, route-native billing units, and returned-document validation findings; the 21 verified prior threads were resolved. One new failover cache-contract issue remains inline.

Local validation passed the full backend test suite, vet, golangci-lint v2.12.2, the full plugin-devkit suite, frontend lint/typecheck/tests/build, 10 browser smoke tests, 52 UI fixture scenarios, 177 repository tests, all root gates, and `git diff --check`. All nine GitHub checks are green.

